PR #343, opened 2022-11-21
feat: Handle large annotation limit

Arvindthiru:

Fixes #

I have:


pkg/controllers/work/apply_controller.go
```diff
@@ -50,6 +52,8 @@ import (
 
 const (
 	workFieldManagerName = "work-api-agent"
+	fleetSystemNamespace = "fleet-system"
```
Ealianis, line 55: I think it's time for us to create a constants package. We avoided it during the reserved namespace work as we used a prefix. But my current PR is using a string literal for an ObjectKey. 

https://github.com/Azure/fleet/pull/346/files#diff-df64e37f2222b761c62bd8bf32c0347aae0fe1f46ff6acf0e66c2b61bd30d761R398

pkg/controllers/work/apply_controller.go
```diff
@@ -50,6 +52,8 @@ import (
...
+	configMapNamePrefix  = "configmap-"
```
Ealianis, line 56: Do we need pkg level constants / vars if they are only used once within the source? I believe I've received that feedback before.

pkg/controllers/work/apply_controller.go
```diff
@@ -629,6 +669,140 @@ func buildManifestAppliedCondition(err error, action applyAction, observedGenera
...
+		}
+	}
+	return nil
+}
+
+func (r *ApplyWorkReconciler) getRandomConfigMapName(ctx context.Context) string {
+	generator := names.SimpleNameGenerator
+	var configMapName string
+	var configMap v1.ConfigMap
+	for {
```
Ealianis, line 786: What happens if we get stuck in a prolonged or endless loop due to the API error not being IsNotFound.
I wonder if we should use a bound retry ?
https://pkg.go.dev/k8s.io/client-go/util/retry

pkg/controllers/work/patch_util.go
```diff
@@ -109,6 +107,15 @@ func setModifiedConfigurationAnnotation(obj runtime.Object) error {
 	return metadataAccessor.SetAnnotations(obj, annotations)
 }
 
+// setModifiedConfigurationAnnotation serializes the object into byte stream and returns it.
```
Ealianis, line 110:
```suggestion
// computeModifiedConfiguration serializes the object into a byte stream, and returns it.
```

pkg/controllers/work/apply_controller.go
```diff
@@ -629,6 +672,144 @@ func buildManifestAppliedCondition(err error, action applyAction, observedGenera
...
+	configMap := &v1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      configMapName,
+			Namespace: fleetSystemNamespace,
+		},
+		Data: map[string]string{},
+	}
+	configMap.Data[manifestHashKey] = manifestHash
+	configMap.Data[lastAppliedConfigKey] = lastModifiedConfig
+	// add applied work owner reference. let's check if it works
```
Ealianis, line 685: Comment doesn't add value.

pkg/controllers/work/apply_controller.go
```diff
@@ -629,6 +672,144 @@ func buildManifestAppliedCondition(err error, action applyAction, observedGenera
...
+		}
+	}
+	return nil
+}
+
+func (r *ApplyWorkReconciler) getRandomConfigMapName(ctx context.Context) (string, error) {
+	generator := names.SimpleNameGenerator
+	var configMapName string
+	var configMap v1.ConfigMap
+	for {
```
michaelawyu, line 792: Hi Arvind! Yeah, I agree with Ealianis' comment; this is not a very good pattern, the for-loop here essentially bypasses the exponential back-off mechanism that `controll-runtime` provides, and in adverse situations the controller might get trapped in an infinite loop.

Plus, the logic here may not work as you would expect; `controller-runtime` clients use cached reads; if the cache is not fresh enough, name collisions can still occur.

pkg/controllers/work/apply_controller.go
michaelawyu, line 56: A nit here: this prefix might be a little generic.

pkg/controllers/work/apply_controller.go
```diff
@@ -335,26 +339,38 @@ func (r *ApplyWorkReconciler) decodeManifest(manifest workv1alpha1.Manifest) (sc
...
+		lastModifiedConfig, err := computeModifiedConfiguration(manifestObj)
+		if err != nil {
 			return nil, ManifestNoChangeAction, err
 		}
 		actual, err := r.spokeDynamicClient.Resource(gvr).Namespace(manifestObj.GetNamespace()).Create(
 			ctx, manifestObj, metav1.CreateOptions{FieldManager: workFieldManagerName})
 		if err == nil {
 			klog.V(2).InfoS("successfully created the manifest", "gvr", gvr, "manifest", manifestRef)
+			// create configmap
+			if err := r.createConfigMap(ctx, configMapName, owner, manifestHash, lastModifiedConfig); err != nil {
```
michaelawyu, line 371: Hi Arvind! The `createConfigMap` function is repeatedly called throughout the code; a little bit refactoring could help make the code more clear.

michaelawyu: Or more precisely, this place might not be the best one to do this create call.

pkg/controllers/work/apply_controller.go
```diff
@@ -629,6 +672,144 @@ func buildManifestAppliedCondition(err error, action applyAction, observedGenera
...
+			Namespace: fleetSystemNamespace,
+		},
+		Data: map[string]string{},
+	}
+	configMap.Data[manifestHashKey] = manifestHash
+	configMap.Data[lastAppliedConfigKey] = lastModifiedConfig
+	// add applied work owner reference. let's check if it works
+	addOwnerRef(owner, configMap)
+	if err := r.spokeClient.Create(ctx, configMap); err != nil {
+		klog.ErrorS(err, "config map create failed", "configMap")
```
Ealianis, line 688: It looks as though are going to get multiple error logs that contain the same message. 
See  line 716. https://github.com/Azure/fleet/pull/343/files#diff-9c51763423a3a9a026f47bc21555619edb62a1cf676c669b643e57b53c39c7caR716

pkg/controllers/work/apply_controller.go
```diff
@@ -629,6 +672,144 @@ func buildManifestAppliedCondition(err error, action applyAction, observedGenera
...
+		klog.ErrorS(err, "failed to update config map", "configMap", configMap.Name)
+		return err
+	}
+	return nil
+}
+
+// We can calculate the manifestHash & lastAppliedConfig again wiping out previous state or make sure the annotations in the last applied config doesn't cause issue in patch.
+func (r *ApplyWorkReconciler) migrateToConfigMap(ctx context.Context, gvr schema.GroupVersionResource, obj *unstructured.Unstructured, owner metav1.OwnerReference) error {
+	// we can also use obj.getAnnotations().
+	annots, err := metadataAccessor.Annotations(obj)
```
michaelawyu, line 758: Hi Arvind! We might need to double-confirm this; for unstructured objects, this function seems to return a copy of the annotations map, but in this piece of code all the updates are in-place.

See also:
https://github.com/kubernetes/apimachinery/blob/4fbe8e4db39b17a53c0756caf58466baad871871/pkg/apis/meta/v1/unstructured/unstructured.go#L407
https://github.com/kubernetes/apimachinery/blob/4fbe8e4db39b17a53c0756caf58466baad871871/pkg/apis/meta/v1/unstructured/helpers.go#L166

pkg/controllers/work/apply_controller.go
```diff
@@ -629,6 +672,144 @@ func buildManifestAppliedCondition(err error, action applyAction, observedGenera
...
+	if annots == nil {
+		return errors.New("manifest does not have any annotations")
+	}
+	// Assuming that manifest has both annotations on it
+	manifestHash, ok1 := annots[manifestHashAnnotation]
+	lastModifiedConfig, ok2 := annots[lastAppliedConfigAnnotation]
+	if ok1 || ok2 {
+		delete(annots, manifestHashAnnotation)
+		delete(annots, lastAppliedConfigAnnotation)
+		configMapName, err := r.getRandomConfigMapName(ctx)
```
michaelawyu, line 771: (See the comment above about refactoring the createConfigMap call; a little bit refactoring can help here)

pkg/controllers/work/apply_controller.go
```diff
@@ -629,6 +672,144 @@ func buildManifestAppliedCondition(err error, action applyAction, observedGenera
...
+}
+
+func (r *ApplyWorkReconciler) getConfigMap(ctx context.Context, obj *unstructured.Unstructured, owner metav1.OwnerReference) (*v1.ConfigMap, error) {
+	var configMap v1.ConfigMap
+	annotations := obj.GetAnnotations()
+	configMapName, ok := annotations[configMapNameAnnotation]
+	if !ok {
+		return nil, errors.New("manifest doesn't have a config map")
+	}
+	if err := r.spokeClient.Get(ctx, types.NamespacedName{Name: configMapName, Namespace: fleetSystemNamespace}, &configMap); err != nil {
```
michaelawyu, line 702: Hi Arvind! I understand the purpose of this code snippet here, but it could really use some refactoring.

* The createConfigMap call happens for the third time
* It really is not a very good idea to use recursion here.

pkg/controllers/work/apply_controller.go
```diff
@@ -375,41 +391,64 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+		return nil, ManifestNoChangeAction, err
+	}
+
+	klog.InfoS("generation for configmap", "generation", curObjConfigMap.GetGeneration())
+
+	// set config map name annotation on manifest object to prevent deletion during patch.
+	setConfigMapNameAnnotation(curObjConfigMap.GetName(), manifestObj)
+
+	// Use manifest generation to check if configmap is outdated
+	if curObj.GetGeneration() != curObjConfigMap.GetGeneration() {
```
michaelawyu, line 422: Hi Arvind! A few issues here:

* first of all, IIRC Kubernetes API server does not track generations for `configMaps`; the field will always be 0
* This checks compares the generations of two very different objects, which will only work if the two values are in-sync (there are no guarantees)

Arvindthiru: Yes, was checking the logs and realized the generation for config map is always zero and the manifest's generation is also zero after patch

pkg/controllers/work/apply_controller.go
```diff
@@ -426,6 +465,11 @@ func (r *ApplyWorkReconciler) patchCurrentResource(ctx context.Context, gvr sche
 		klog.ErrorS(patchErr, "failed to patch the manifest", "gvr", gvr, "manifest", manifestRef)
 		return nil, ManifestNoChangeAction, patchErr
 	}
+
+	// calculate manifest hash & last modified config again to set the new manifest hash & last applied config after patch gets applied
+	if err := r.updateConfigMap(ctx, curObjConfigMap, manifestObj); err != nil {
```
michaelawyu, line 470: Hi Arvind! Just to double-check, is it really necessary to update the configmap for a second time here?

pkg/controllers/work/apply_controller.go
```diff
@@ -375,41 +391,64 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+	if curObj.GetGeneration() != curObjConfigMap.GetGeneration() {
+		if err := r.updateConfigMap(ctx, curObjConfigMap, curObj); err != nil {
+			return nil, ManifestNoChangeAction, err
+		}
+	}
+
 	// We only try to update the object if its spec hash value has changed.
-	if manifestObj.GetAnnotations()[manifestHashAnnotation] != curObj.GetAnnotations()[manifestHashAnnotation] {
-		return r.patchCurrentResource(ctx, gvr, manifestObj, curObj)
+	if manifestHash != curObjConfigMap.Data[manifestHashKey] {
```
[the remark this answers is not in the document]
Arvindthiru: Ideally the configmap would get updated only if it's outdated meaning a we patched the object for some reason we were not able to update the configmap

pkg/controllers/work/apply_controller.go
Arvindthiru: This is a draft PR, was using comments to keep track

pkg/controllers/work/patch_util.go
```diff
@@ -25,17 +26,14 @@ func init() {
 
 // threeWayMergePatch creates a patch by computing a three-way diff based on
 // an object's current state, modified state, and last-applied-state recorded in its annotation.
-func threeWayMergePatch(currentObj, manifestObj client.Object) (client.Patch, error) {
+func threeWayMergePatch(currentObj, manifestObj client.Object, currentObjConfigMap *v1.ConfigMap) (client.Patch, error) {
```
michaelawyu, line 29: Hi Arvind! It might be helpful to take another look at the implementation of the CreateThreeWayMergePatch function; the semantics do not seem to correlate with the args passed in at this moment.

pkg/controllers/work/apply_controller.go
michaelawyu: As we have discussed over the meeting, the name collision check currently in the PR might not work in adverse conditions (out-of-sync caches), and if the check fails to guard against name collisions, two manifests may share the same configmap and the behavior is undetermined.

Plus, w/o doubly links (or a guaranteed 1:1 name mapping), partial failures may also lead to undetermined behaviors.

pkg/controllers/work/patch_util.go
Arvindthiru: Discussed in the meeting, the args match

pkg/controllers/work/apply_controller.go
Arvindthiru: This need to be updated with the object from the hub cluster

PR #353, opened 2023-01-12
fix: relax the case for last applied annotation is removed

ryanzhang-oss:
Do not return an error if someone removes the last applied annotation in the applied resources. The three way merge lib can handle this case.

Fixes # 349

I have:

IT

### DIFF
--- a/pkg/controllers/work/apply_controller_integration_test.go
+++ b/pkg/controllers/work/apply_controller_integration_test.go
@@ -475,6 +475,7 @@ var _ = Describe("Work Controller", func() {
 		})
 
 		It("Check that the apply still works if the last applied annotation does not exist", func() {
+			ctx = context.Background()
 			cmName := "test-merge-without-lastapply"
 			cmNamespace := defaultNS
 			cm = &corev1.ConfigMap{
@@ -498,8 +499,8 @@ var _ = Describe("Work Controller", func() {
 
 			By("create the work")
 			work = createWorkWithManifest(workNamespace, cm)
-			err := k8sClient.Create(context.Background(), work)
-			Expect(err).ToNot(HaveOccurred())
+			err := k8sClient.Create(ctx, work)
+			Expect(err).Should(Succeed())
 
 			By("wait for the work to be applied")
 			waitForWorkToApply(work.GetName(), work.GetNamespace())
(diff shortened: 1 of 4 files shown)


michaelawyu (dismissed):
LGTM ;)

pkg/controllers/work/patch_util.go
```diff
@@ -114,14 +114,17 @@ func setModifiedConfigurationAnnotation(obj runtime.Object) error {
 func getOriginalConfiguration(obj runtime.Object) ([]byte, error) {
```
michaelawyu, line 114: Hi Ryan! Just a nit here: this function seems to be no longer returning an error in any branch.

ryanzhang-oss: good point, fixed

pkg/controllers/work/apply_controller_integration_test.go
```diff
@@ -474,6 +474,71 @@ var _ = Describe("Work Controller", func() {
...
+					},
+				},
+				Data: map[string]string{
+					"data1": "test1",
+				},
+			}
+
+			By("create the work")
+			work = createWorkWithManifest(workNamespace, cm)
+			err := k8sClient.Create(context.Background(), work)
```
zhiying-lin, line 501: suggest to reuse one context

pkg/controllers/work/apply_controller_integration_test.go
```diff
@@ -474,6 +474,71 @@ var _ = Describe("Work Controller", func() {
...
+			Expect(err).ToNot(HaveOccurred())
```
zhiying-lin, line 502: nit
```suggestion
			Expect(err).Should(Succeed())
```

pkg/controllers/work/patch_util.go
```diff
@@ -114,14 +114,17 @@ func setModifiedConfigurationAnnotation(obj runtime.Object) error {
...
 	annots, err := metadataAccessor.Annotations(obj)
 	if err != nil {
-		return nil, fmt.Errorf("cannot access metadata.annotations: %w", err)
+		klog.ErrorS(err, "cannot access metadata.annotations", "gvk", obj.GetObjectKind().GroupVersionKind())
+		return nil, err
 	}
 	if annots == nil {
-		return nil, errors.New("object does not have lastAppliedConfigAnnotation")
+		klog.Warning("object does not have annotation", "obj", obj)
```
zhiying-lin, line 121: better to leave a comment in the code, 

like // ignoring empty annotation,  three way merge lib (?)  can handle this case

Could you please also make it clear what's the three way merge lib (add code path?)? without full context, it's really hard to understand the comment.

pkg/controllers/work/patch_util.go
```diff
@@ -114,14 +114,17 @@ func setModifiedConfigurationAnnotation(obj runtime.Object) error {
...
+		return nil, nil
 	}
 	original, ok := annots[lastAppliedConfigAnnotation]
 	if !ok {
-		return nil, errors.New("object does not have lastAppliedConfigAnnotation")
+		klog.Warning("object does not have lastAppliedConfigAnnotation", "obj", obj)
```
zhiying-lin, line 126: ditto, add a comment, why we ignore the annotation

michaelawyu (approved):
LGTM ;)

pkg/controllers/work/patch_util.go
ryanzhang-oss: the threeWayMergePatch function is in the same file. We assume that the reviewer is familiar with the existing code.

pkg/controllers/work/patch_util.go
```diff
@@ -117,6 +117,7 @@ func getOriginalConfiguration(obj runtime.Object) ([]byte, error) {
 		klog.ErrorS(err, "cannot access metadata.annotations", "gvk", obj.GetObjectKind().GroupVersionKind())
 		return nil, err
 	}
+	// the threeWayMergePatch lib can handle the case that the original is empty
```
zhiying-lin, line 120:
```suggestion
	// The func threeWayMergePatch can handle the case that the original is empty.
```

zhiying-lin (approved):

zhiying-lin (approved):

michaelawyu (approved):
LGTM ;)

PR #355, opened 2023-02-02
feat: Annotation size limit

### LINKED ISSUE #295: [BUG] Can't apply a resource if it's too big
### **Describe the bug**
The three way merge annotation hits the annotation length limit

```
'Failed to apply manifest: Secret "sh.helm.release.v1.ryan-redis.v1"
          is invalid: metadata.annotations: Too long: must have at most 262144 bytes'
```

### **Environment**
Please provide the following: 
- Hub cluster details
- Member cluster details

### **To Reproduce**
[Follow the steps of the redis install guide](https://techcommunity.microsoft.com/t5/apps-on-azure-blog/run-scalable-and-resilient-redis-with-kubernetes-and-azure/ba-p/3247956)

### **Expected behavior**
The secret should be omitted as it's created by the helm. This bug actually makes it work.

### **Screenshots**
```shell
 status:
    conditions:
    - lastTransitionTime: "2022-09-16T18:57:06Z"
      message: Successfully scheduled resources for placement
      observedGeneration: 1
      reason: ScheduleSucceeded
      status: "True"
      type: Scheduled
    - lastTransitionTime: "2022-09-16T21:14:48Z"
      message: there are failed to apply manifests, please check the `failedResourcePlacements`
        status
      observedGeneration: 1
      reason: ApplyFailed
      status: "False"
      type: Applied
    failedPlacements:
    - clusterName: member1
      condition:
        lastTransitionTime: "2022-09-16T21:14:47Z"
        message: 'Failed to apply manifest: Secret "sh.helm.release.v1.ryan-redis.v1"
          is invalid: metadata.annotations: Too long: must have at most 262144 bytes'
        reason: appliedManifestFailed
        status: "False"
        type: Applied
      kind: Secret
      name: sh.helm.release.v1.ryan-redis.v1
      namespace: redis-test
      version: v1
    - clusterName: member2
      condition:
        lastTransitionTime: "2022-09-16T21:14:46Z"
        message: 'Failed to apply manifest: Secret "sh.helm.release.v1.ryan-redis.v1"
          is invalid: metadata.annotations: Too long: must have at most 262144 bytes'
        reason: appliedManifestFailed
        status: "False"
        type: Applied
      kind: Secret
      name: sh.helm.release.v1.ryan-redis.v1
      namespace: redis-test
      version: v1

```

Arvindthiru:

The changes in the PR validates the annotation size of the last modified config annotation on the manifest and if it crosses the annotation size threshold it removes the annotation and creates/updates the object on the cluster using server side apply instead of using three way merge.

This ensures that when large objects are propagated we don't fail to apply the object

Fixes #295 

I have:

Unit tests

Redis guide: https://techcommunity.microsoft.com/t5/apps-on-azure-blog/run-scalable-and-resilient-redis-with-kubernetes-and-azure/ba-p/3247956

Propagated redis helm secret to member cluster

Screenshot 2023-02-05 at 5 09 36 PM
Screenshot 2023-02-05 at 5 09 49 PM

https://blog.atomist.com/kubernetes-apply-replace-patch/

https://stackoverflow.com/questions/28459418/use-of-put-vs-patch-methods-in-rest-api-real-life-scenarios/39338329#39338329


pkg/controllers/work/apply_controller.go
```diff
@@ -340,6 +346,16 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
 		Name:      manifestObj.GetName(),
 		Namespace: manifestObj.GetNamespace(),
 	}
+	// Marshaling adds about 100-200 bytes to the object's actual size
+	marshaledObj, marshalErr := json.Marshal(manifestObj)
+	if marshalErr != nil {
+		return nil, ManifestNoChangeAction, marshalErr
+	}
+	if len(marshaledObj) > (TotalAnnotationSizeLimitB - len(manifestHashAnnotation) - len(lastAppliedConfigAnnotation) - manifestHashLength) {
+		klog.V(2).InfoS("Size of manifest object is greater than 262,013 bytes", "Name", manifestObj.GetName(), "size of marshaled object", len(marshaledObj), "kind", manifestObj.GetKind())
```
Ealianis, line 355: I believe this log is to indicate a flow change. That being the object will be applied, but we won't be storing the manifest within the annotation. Should we indicate that within the log?

pkg/controllers/work/apply_controller_test.go
```diff
@@ -346,6 +348,17 @@ func TestApplyUnstructured(t *testing.T) {
...
 	specHashFailObj.Object["test"] = math.Inf(1)
 
+	utils.GetObjectFromManifest("../../../test/integration/manifests/resources/test-large-secret.yaml", &largeSecret)
+	rawSecret, _ := json.Marshal(largeSecret)
+	var largeObj unstructured.Unstructured
+	largeObj.UnmarshalJSON(rawSecret)
+
+	// add check to see if we cannot retrieve object
+	applyDynamicClient := fake.NewSimpleDynamicClient(runtime.NewScheme())
+	applyDynamicClient.PrependReactor("apply", "*", func(action testingclient.Action) (handled bool, ret runtime.Object, err error) {
```
Ealianis, line 358: "Apply"? It's probably an HTTP verb.

pkg/controllers/work/apply_controller.go
```diff
@@ -340,6 +346,16 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
 		Name:      manifestObj.GetName(),
 		Namespace: manifestObj.GetNamespace(),
 	}
+	// Marshaling adds about 100-200 bytes to the object's actual size
+	marshaledObj, marshalErr := json.Marshal(manifestObj)
+	if marshalErr != nil {
+		return nil, ManifestNoChangeAction, marshalErr
+	}
+	if len(marshaledObj) > (TotalAnnotationSizeLimitB - len(manifestHashAnnotation) - len(lastAppliedConfigAnnotation) - manifestHashLength) {
```
ryanzhang-oss, line 354: Why are we checking the size of the manifestObj instead of the size of the last applied Annotation?

ryanzhang-oss: does the object have the lastAppliedConfigAnnotation label now?

pkg/utils/test_util.go
```diff
@@ -37,6 +47,29 @@ func GetEventString(object runtime.Object, eventtype, reason, messageFmt string,
 			object.GetObjectKind().GroupVersionKind().Kind, object.GetObjectKind().GroupVersionKind().GroupVersion())
 }
 
+// GetObjectFromRawExtension returns an object decoded from the raw byte array
```
Arvindthiru, line 50: Had to copy these methods from the test package cause I was not able to import them form a *_test.go file https://github.com/Azure/fleet/blob/main/test/integration/utils_test.go#L62-L78

pkg/controllers/work/apply_controller.go
Arvindthiru: Computing and checking the last applied config, current implementation doesn't have the label hence I'm subtracting it from the 256KB allowed for annotation limit

pkg/controllers/work/apply_controller.go
```diff
@@ -340,6 +346,16 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+		klog.V(2).InfoS("Size of manifest object is greater than 262,013 bytes", "Name", manifestObj.GetName(), "size of marshaled object", len(marshaledObj), "kind", manifestObj.GetKind())
+		return r.applyObject(ctx, gvr, manifestObj)
```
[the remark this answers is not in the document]
Arvindthiru: Fixed by checking the manifest hash for updates on the object

pkg/utils/test_util.go
Arvindthiru: https://github.com/golang/go/issues/39565

pkg/controllers/work/apply_controller.go
Arvindthiru: Changed log

pkg/utils/test_util.go
ryanzhang-oss: why not remove the dup from the utils_test?

Arvindthiru: Moved methods from util_test.go

pkg/controllers/work/apply_controller.go
```diff
@@ -384,12 +394,55 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+			return r.applyObject(ctx, gvr, manifestObj)
+		}
 		return r.patchCurrentResource(ctx, gvr, manifestObj, curObj)
 	}
 
 	return curObj, ManifestNoChangeAction, nil
 }
 
+// isAnnotationSizeValid returns true if the total size of the annotations for a manifest is less than 256KB.
+func isAnnotationsSizeValid(gvr schema.GroupVersionResource, manifestRef klog.ObjectRef, annotations map[string]string) bool {
```
michaelawyu, line 417: Hi Arvind! For this method, maybe it is better idea to reuse the [ValidateAnnotationsSize](ValidateAnnotationsSize) function from the Kubernetes API validation package directly.

pkg/controllers/work/apply_controller.go
```diff
@@ -351,6 +356,16 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
 		if err := setModifiedConfigurationAnnotation(manifestObj); err != nil {
 			return nil, ManifestNoChangeAction, err
 		}
+		annotations := manifestObj.GetAnnotations()
+		klog.V(2).InfoS("validating annotation size for manifest",
+			"gvr", gvr, "manifest", manifestRef)
+		err := validation.ValidateAnnotationsSize(annotations)
+		if err != nil {
+			klog.ErrorS(err, "not using three way merge for manifest removing last applied config annotation",
```
zhiying-lin, line 364: should it be debug info only?

pkg/controllers/work/apply_controller_test.go
```diff
@@ -346,6 +347,46 @@ func TestApplyUnstructured(t *testing.T) {
...
 
+	utils.GetObjectFromManifest("../../../test/integration/manifests/resources/test-large-secret.yaml", &largeSecret)
+	largeSecret.ObjectMeta = metav1.ObjectMeta{
+		OwnerReferences: []metav1.OwnerReference{
+			ownerRef,
+		},
+	}
+	rawSecret, _ := json.Marshal(largeSecret)
+	var largeObj unstructured.Unstructured
+	_ = largeObj.UnmarshalJSON(rawSecret)
```
zhiying-lin, line 358: please do not ignore the error. you should fail the test if the setup has errors.

https://github.com/golang/go/wiki/CodeReviewComments#handle-errors

pkg/utils/test_util.go
```diff
@@ -7,8 +7,13 @@ package utils
```
zhiying-lin, line 7: we need to create a separate package for the test utils, not just putting into a separate file.
Otherwise, callers may accidently use the test util by calling util package.

The package was not introduced by this PR, and could address it separately.

pkg/utils/test_util.go
```diff
@@ -37,6 +47,29 @@ func GetEventString(object runtime.Object, eventtype, reason, messageFmt string,
 			object.GetObjectKind().GroupVersionKind().Kind, object.GetObjectKind().GroupVersionKind().GroupVersion())
 }
 
+// GetObjectFromRawExtension returns an object decoded from the raw byte array
+func GetObjectFromRawExtension(rawByte []byte, obj runtime.Object) {
+	json, err := yaml.ToJSON(rawByte)
+	if err != nil {
```
zhiying-lin, line 53: we should not ignore any error.

pkg/controllers/work/apply_controller_test.go
```diff
@@ -80,6 +80,7 @@ var (
 	testManifest         = workv1alpha1.Manifest{RawExtension: runtime.RawExtension{
 		Raw: rawTestDeployment,
 	}}
+	largeSecret v1.Secret
```
zhiying-lin, line 83: please declare the variable close to where you use it to improve the readability.

pkg/utils/test_util.go
zhiying-lin, line 50: missing period

zhiying-lin (changes requested):
- please add e2e tests to cover create/update the large secret.
- please update your PR description to indicate how/why your PR fixes the bug.

pkg/utils/test_util.go
Arvindthiru: Will address this and add E2E in a separate PR

Arvindthiru:
Putting the PR as draft to work on E2E

Arvindthiru:
Instead of removing the annotation I'm currently setting the annotation to "". This is done because of the scenario described below.

- Create an object within the annotation limit, we get an with spec-hash and last modified config annotations on it
- Then when we update the object to be large, we use server side apply to update the object but since it's server side apply even if we remove the annotation in the controller it sees the pre-existing last applied config and since the controller is not managing the field now it leaves it alone and updates other fields
- This is not the correct behavior as we have a large object with the small object's last applied config
- To avoid this we can update the field to be empty, and the next time around when we update the object to be small three way merge works fine as it accepts empty string as last applied config it uses an empty map

pkg/controllers/work/apply_controller.go
```diff
@@ -351,6 +356,16 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+				"gvr", gvr, "obj", manifestRef)
+			delete(annotations, lastAppliedConfigAnnotation)
+			manifestObj.SetAnnotations(annotations)
+		}
```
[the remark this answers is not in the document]
Arvindthiru: Can I handle this change in a separate PR

pkg/controllers/work/apply_controller_test.go
```diff
@@ -444,7 +514,7 @@ func TestApplyUnstructured(t *testing.T) {
 			} else {
 				assert.Truef(t, err == nil, "err is not nil for Testcase %s", testName)
 				assert.Truef(t, applyResult != nil, "applyResult is not nil for Testcase %s", testName)
-				assert.Equalf(t, testCase.resultSpecHash, applyResult.GetAnnotations()[manifestHashAnnotation],
+				assert.Equalf(t, testCase.resultSpecHash, applyResult.GetAnnotations()[ManifestHashAnnotation],
```
zhiying-lin, line 517: should we add the validation of the lastAppliedConfig?

test/e2e/work_load_test.go
```diff
@@ -19,16 +19,19 @@ import (
 	workapi "sigs.k8s.io/work-api/pkg/apis/v1alpha1"
 
 	"go.goms.io/fleet/apis/v1alpha1"
+	workcontroller "go.goms.io/fleet/pkg/controllers/work"
```
zhiying-lin, line 22: We need to rename the package only when there is a conflict.

test/e2e/work_load_test.go
```diff
@@ -320,5 +323,129 @@ var _ = Describe("workload orchestration testing", func() {
...
+			By("check if resources in namespace are propagated to member cluster")
+			ownerReferences := []metav1.OwnerReference{
+				{
+					APIVersion:         workapi.GroupVersion.String(),
+					BlockOwnerDeletion: pointer.Bool(false),
+					Kind:               "AppliedWork",
+					Name:               crp.Name,
+				},
+			}
+			expectedNamespace := namespace
```
zhiying-lin, line 401: prefer to use "want" to "expected", to match with "got".

test/e2e/work_load_test.go
```diff
@@ -320,5 +323,129 @@ var _ = Describe("workload orchestration testing", func() {
...
+			Expect(err).Should(Succeed())
+			// testLargeSecret has the same name as testSmallSecret
+			Expect(HubCluster.KubeClient.Update(ctx, &testLargeSecret)).Should(Succeed(), "Failed to update secret %s to be large in %s cluster", testLargeSecret.Name, HubCluster.ClusterName)
+			expectedSecret = &testLargeSecret
+			expectedSecret.OwnerReferences = ownerReferences
+
+			// Ignoring Annotations here because fleet.azure.com/last-applied-configuration has live fields, checking to see if it's not empty instead.
+			gotSecret = utils.CmpSecret(ctx, *MemberCluster, &types.NamespacedName{Name: testLargeSecret.Name, Namespace: testLargeSecret.Namespace}, expectedSecret, resourceIgnoreOptions)
+			Expect(gotSecret.Annotations[workcontroller.LastAppliedConfigAnnotation]).To(BeEmpty())
+			Expect(gotSecret.Annotations[workcontroller.ManifestHashAnnotation]).ToNot(Equal(testSmallSecretSpecHash))
```
zhiying-lin, line 424: did not see much value to compare with the smallSecretSpecHash. If you want to validate this annotation, you need to compare the exact value.

pkg/controllers/work/apply_controller.go
```diff
@@ -383,13 +399,50 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+			return nil, ManifestNoChangeAction, err
+		}
+		annotations := manifestObj.GetAnnotations()
+		klog.V(2).InfoS("validating annotation size for manifest",
+			"gvr", gvr, "manifest", manifestRef)
+		if err = validation.ValidateAnnotationsSize(annotations); err != nil {
+			klog.ErrorS(err, "not using three way merge for manifest removing last applied config annotation",
+				"gvr", gvr, "obj", manifestRef)
+			annotations[LastAppliedConfigAnnotation] = ""
+			manifestObj.SetAnnotations(annotations)
```
zhiying-lin, line 417: As i mentioned before, this part could be part of the setModifiedConfigurationAnnotation func. then you don't need to duplicate the code in two places.

For the update, you only need to check if the LastAppliedConfigAnnotation is empty, server side apply, otherwise, three way merge.

pkg/controllers/work/apply_controller.go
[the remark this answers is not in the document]
zhiying-lin: Could you please address it in the same PR? if you need the help, i'm willing to help you via the call.

pkg/controllers/work/apply_controller_test.go
Arvindthiru: Checking lastAppliedConfig is difficult cause it contains live fields

pkg/controllers/work/apply_controller.go
```diff
@@ -531,6 +562,7 @@ func computeManifestHash(obj *unstructured.Unstructured) (string, error) {
 	manifest.SetSelfLink("")
 	manifest.SetDeletionTimestamp(nil)
 	manifest.SetManagedFields(nil)
+	manifest.SetOwnerReferences(nil)
```
Arvindthiru, line 565: Had to make this change to get E2E to pass for spec hash check, because spec hash for object kept changing on different due live UID field in OwnerRef which changes only on different E2E runs.

test/e2e/work_load_test.go
Arvindthiru: weird issue Goland doesn't allow me to directly import this package it only allow me to import it when used as an alias

pkg/controllers/work/apply_controller.go
Arvindthiru: This causes some UTs to fail to fix it we need to remove live fields UID, BlockOwnerDeletion and Controller field if populated in OwnerRef

Arvindthiru: reverted change and added comment explaining why

test/e2e/work_load_test.go
zhiying-lin: https://go.dev/doc/effective_go#names
Another convention is that the package name is the base name of its source directory;
it may be caused by that we name the package is "controllers" instead of "work".

You could rename it for now and we could fix it in a separate PR for the whole repo.

pkg/controllers/work/apply_controller.go
```diff
@@ -383,13 +389,46 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
 	}
 
 	// We only try to update the object if its spec hash value has changed.
-	if manifestObj.GetAnnotations()[manifestHashAnnotation] != curObj.GetAnnotations()[manifestHashAnnotation] {
+	if manifestObj.GetAnnotations()[ManifestHashAnnotation] != curObj.GetAnnotations()[ManifestHashAnnotation] {
+		// we need to merge the owner reference between the current and the manifest since we support one manifest
+		// belong to multiple work so it contains the union of all the appliedWork.
+		manifestObj.SetOwnerReferences(mergeOwnerReference(curObj.GetOwnerReferences(), manifestObj.GetOwnerReferences()))
+		// record the raw manifest with the hash annotation in the manifest.
+		if err = setModifiedConfigurationAnnotation(manifestObj); err != nil {
```
zhiying-lin, line 397:
```suggestion
		if err := setModifiedConfigurationAnnotation(manifestObj); err != nil {
```

pkg/controllers/work/apply_controller_test.go
```diff
@@ -346,6 +346,54 @@ func TestApplyUnstructured(t *testing.T) {
...
+		OwnerReferences: []metav1.OwnerReference{
+			ownerRef,
+		},
+	}
+	rawSecret, err := json.Marshal(largeSecret)
+	if err != nil {
+		t.Errorf("failed to marshal secret: %s", err)
+	}
+	var largeObj unstructured.Unstructured
+	if err = largeObj.UnmarshalJSON(rawSecret); err != nil {
```
zhiying-lin, line 363:
```suggestion
	if err := largeObj.UnmarshalJSON(rawSecret); err != nil {
```

pkg/controllers/work/apply_controller_test.go
```diff
@@ -346,6 +346,54 @@ func TestApplyUnstructured(t *testing.T) {
...
+		t.Errorf("failed to unmarshal JSON: %s", err)
+	}
+	updatedLargeObj := largeObj.DeepCopy()
+
+	largeObjSpecHash, _ := computeManifestHash(&largeObj)
```
zhiying-lin, line 368: please fail the test if it has error

pkg/controllers/work/patch_util.go
```diff
@@ -105,8 +106,22 @@ func setModifiedConfigurationAnnotation(obj runtime.Object) error {
...
+	annotations[LastAppliedConfigAnnotation] = string(modified)
+	if err = metadataAccessor.SetAnnotations(obj, annotations); err != nil {
+		return err
+	}
+	annotations, err = metadataAccessor.Annotations(obj)
+	if err != nil {
+		return err
+	}
+	if err := validation.ValidateAnnotationsSize(annotations); err != nil {
+		klog.InfoS(fmt.Sprintf("setting last applied config annotation to empty, %s", err))
```
zhiying-lin, line 118:
```suggestion
		klog.klog.V(2).InfoS(fmt.Sprintf("setting last applied config annotation to empty, %s", err))
```
need to add the verbosity level

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use

pkg/controllers/work/patch_util.go
```diff
@@ -105,8 +106,22 @@ func setModifiedConfigurationAnnotation(obj runtime.Object) error {
...
+		annotations[LastAppliedConfigAnnotation] = ""
+		if err = metadataAccessor.SetAnnotations(obj, annotations); err != nil {
+			return err
+		}
+	}
+	return nil
```
zhiying-lin, line 124:
```suggestion
	annotations[LastAppliedConfigAnnotation] = string(modified)
	if err := validation.ValidateAnnotationsSize(annotations); err != nil {
		klog.V(2).InfoS(fmt.Sprintf("setting last applied config annotation to empty, %s", err))
		annotations[LastAppliedConfigAnnotation] = ""
	}
	return metadataAccessor.SetAnnotations(obj, annotations)
```
you don't need to set & get the annotation multiple times.

Please update the func comment to indicate the "set the annotation to empty" logic.

pkg/controllers/work/apply_controller_test.go
```diff
@@ -433,6 +481,28 @@ func TestApplyUnstructured(t *testing.T) {
 			resultAction:   ManifestUpdatedAction,
 			resultErr:      nil,
 		},
+		"test apply succeeds for large manifest when object does not exist": {
```
zhiying-lin, line 484: please add cases 
1. when create large manifest fails
2.  when update large manifest fails

pkg/controllers/work/patch_util.go
```diff
@@ -107,19 +108,12 @@ func setModifiedConfigurationAnnotation(obj runtime.Object) error {
...
+		klog.V(2).InfoS(fmt.Sprintf("setting last applied config annotation to empty, %s", err))
 		annotations[LastAppliedConfigAnnotation] = ""
-		if err = metadataAccessor.SetAnnotations(obj, annotations); err != nil {
-			return err
-		}
+	}
+	if err = metadataAccessor.SetAnnotations(obj, annotations); err != nil {
+		return err
 	}
 	return nil
```
zhiying-lin:
```suggestion
	return metadataAccessor.SetAnnotations(obj, annotations)
```

to simplify the code

zhiying-lin (approved):
one minor comment, others LGTM

pkg/controllers/work/apply_controller.go
```diff
@@ -383,13 +389,46 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+		manifestObj.SetOwnerReferences(mergeOwnerReference(curObj.GetOwnerReferences(), manifestObj.GetOwnerReferences()))
+		// record the raw manifest with the hash annotation in the manifest.
+		if err := setModifiedConfigurationAnnotation(manifestObj); err != nil {
+			return nil, ManifestNoChangeAction, err
+		}
+		annotations := manifestObj.GetAnnotations()
+		if annotations[LastAppliedConfigAnnotation] == "" {
+			klog.V(2).InfoS("using server side apply for manifest", "gvr", gvr, "manifest", manifestRef)
+			return r.applyObject(ctx, gvr, manifestObj)
+		}
```
ryanzhang-oss, line 404: Why is this needed if the setModifiedConfigurationAnnotation returns if the annotation is set or not. It seems that the information is ready but we somehow checked it again.

pkg/controllers/work/apply_controller.go
```diff
@@ -383,13 +389,46 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+	manifestObj *unstructured.Unstructured) (*unstructured.Unstructured, applyAction, error) {
+	manifestRef := klog.ObjectRef{
+		Name:      manifestObj.GetName(),
+		Namespace: manifestObj.GetNamespace(),
+	}
+	options := metav1.ApplyOptions{
+		FieldManager: workFieldManagerName,
+		Force:        true,
+	}
+	manifestObj, err := r.spokeDynamicClient.Resource(gvr).Namespace(manifestObj.GetNamespace()).Apply(ctx, manifestObj.GetName(), manifestObj, options)
```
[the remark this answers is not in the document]
Arvindthiru: I'm using server side apply because,

- It forces conflicts on fields which is managed by the controller similar to three way merge patch
- It doesn't care about fields changed by users on the member cluster side

pkg/controllers/work/apply_controller.go
Arvindthiru: In the current implementation, setModifiedConfigurationAnnotation sets the lastAppliedConfig Annotation if the annotation size is valid if not it sets the  lastAppliedConfig annotation to "". The check is to ensure if we require a server side apply or three way merge patch

ryanzhang-oss (approved):

pkg/controllers/work/patch_util_test.go
```diff
@@ -0,0 +1,44 @@
...
+			obj:      largeObj,
+			wantBool: false,
+			wantErr:  nil,
+		},
+	}
+
+	for testName, testCase := range tests {
+		t.Run(testName, func(t *testing.T) {
+			gotBool, gotErr := setModifiedConfigurationAnnotation(testCase.obj)
+			assert.Equalf(t, testCase.wantBool, gotBool, "got bool not matching for Testcase %s", testName)
```
zhiying-lin, line 40: Can we try to avoid the Assert libraries in the UT?

"Avoid the use of 'assert' libraries to help your tests. "
https://github.com/golang/go/wiki/TestComments#assert-libraries

PR #362, opened 2023-05-12
feat: a new "scope" argument for refresh-token

mingqishao:

Today the Azure provider are using a static scope in token. In future the different scope need to be used for the flexible scenario. 

This changes are:
1. add a new `--scope` arguments in refresh-token. The new scope argument will be pass to Azure token provider. 
3. For the backward compatibility, the scope argument is optional. The old static scope will be used by default. 

I have:

I added unit test for changes codes

### DIFF
--- a/cmd/authtoken/main.go
+++ b/cmd/authtoken/main.go
@@ -60,7 +60,9 @@ func parseArgs() (interfaces.AuthTokenProvider, error) {
 	}
 
 	azureCmd.Flags().StringVar(&clientID, "clientid", "", "Azure AAD client ID (required)")
-	azureCmd.Flags().StringVar(&scope, "scope", "", "Azure AAD token scope (optinal)")
+	// TODO: this scope argument is specific for Azure provider. We should allow registering and parsing provider specific argument
+	// in provider level, instead of global level.
+	azureCmd.Flags().StringVar(&scope, "scope", "", "Azure AAD token scope (optional)")
 	_ = azureCmd.MarkFlagRequired("clientid")
 
 	rootCmd.AddCommand(secretCmd, azureCmd)
--- a/cmd/authtoken/main.go
+++ b/cmd/authtoken/main.go
@@ -23,10 +23,6 @@ var (
 	configPath string
 )
 
-const (
-	aksAADApplicationID = "6dae42f8-4368-4678-94ff-3960e28e3630"
-)
-
 func parseArgs() (interfaces.AuthTokenProvider, error) {
 	var tokenProvider interfaces.AuthTokenProvider
 	rootCmd := &cobra.Command{Use: "refreshtoken", Args: cobra.NoArgs}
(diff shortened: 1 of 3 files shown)


cmd/authtoken/main.go
```diff
@@ -23,6 +23,10 @@ var (
 	configPath string
 )
 
+const (
+	aksAADApplicationID = "6dae42f8-4368-4678-94ff-3960e28e3630"
+)
+
```
ryanzhang-oss, line 29: why move the azure specific const to the main?

cmd/authtoken/main.go
```diff
@@ -50,15 +54,17 @@ func parseArgs() (interfaces.AuthTokenProvider, error) {
...
 		Use:  "azure",
 		Args: cobra.NoArgs,
 		Run: func(_ *cobra.Command, args []string) {
-			tokenProvider = azure.New(clientID)
+			tokenProvider = azure.New(clientID, scope)
 		},
 	}
 
 	azureCmd.Flags().StringVar(&clientID, "clientid", "", "Azure AAD client ID (required)")
+	azureCmd.Flags().StringVar(&scope, "scope", aksAADApplicationID, "Azure AAD token scope (optinal)")
```
ryanzhang-oss, line 67:
```suggestion
	azureCmd.Flags().StringVar(&scope, "scope", "", "Azure AAD token scope (optional)")
```

cmd/authtoken/main_test.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func TestParseArgs(t *testing.T) {
+	t.Run("all arguments", func(t *testing.T) {
+		os.Args = []string{"refreshtoken", "azure", "--clientid=test-client-id", "--scope=test-scope"}
+		t.Cleanup(func() {
+			os.Args = nil
+		})
+		tokenProvider, err := parseArgs()
+		assert.NotNil(t, tokenProvider)
```
ryanzhang-oss, line 17: this didn't really verify that the scope is parsed.  just checking no error is not enough.

pkg/authtoken/providers/azure/azure_msi.go
```diff
@@ -17,15 +17,12 @@ import (
...
 type azureAuthTokenProvider struct {
 	clientID string
+	scope    string
 }
 
-func New(clientID string) interfaces.AuthTokenProvider {
+func New(clientID, scope string) interfaces.AuthTokenProvider {
 	return &azureAuthTokenProvider{
 		clientID: clientID,
 	}
```
ryanzhang-oss, line 28:
```suggestion
        if scope == "" {
            scope = aksScope
        }
	return &azureAuthTokenProvider{
		clientID: clientID,
		scope: scope,
	}
```

ryanzhang-oss (changes requested):
The new scope field is not even assigned.

cmd/authtoken/main.go
mingqishao: I have moved back the const to `pkg/authtoken/providers/azure/azure_msi.go`, and keep the default scope value logic there too.

cmd/authtoken/main.go
mingqishao: made changes as suggested

pkg/authtoken/providers/azure/azure_msi.go
mingqishao: made changes as suggested

mingqishao:
> The new scope field is not even assigned.

I found that too. That is shame. I have fixed it.

ryanzhang-oss (approved):

PR #364, opened 2023-05-19
feat: allow member agent use OS's root certificate authority

mingqishao:

When member agent use secure TLS talk to hub, it  use either a specified CA bundle or CA data in configuration. This changes to allow member agent use OS's [root certificate](https://go.dev/src/crypto/x509/root_linux.go). 

The changes are:
1. refactor the codes of building hub's kubeconfig
2. It is not mandatory to specify CA data or CA bundle in arguments

I have:

### DIFF
--- a/cmd/memberagent/main.go
+++ b/cmd/memberagent/main.go
@@ -39,7 +39,7 @@ import (
 
 var (
 	scheme               = runtime.NewScheme()
-	useCAAuth            = flag.Bool("use-ca-auth", false, "Use identity and CA bundle to authenticate the member agent.")
+	useCertificateAuth   = flag.Bool("use-ca-auth", false, "Use key and certificate to authenticate the member agent.")
 	tlsClientInsecure    = flag.Bool("tls-insecure", false, "Enable TLSClientConfig.Insecure property. Enabling this will make the connection inSecure (should be 'true' for testing purpose only.)")
 	hubProbeAddr         = flag.String("hub-health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	hubMetricsAddr       = flag.String("hub-metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
@@ -71,7 +71,7 @@ func main() {
 		klog.ErrorS(errors.New("hub server api cannot be empty"), "error has occurred retrieving HUB_SERVER_URL")
 		os.Exit(1)
 	}
-	hubConfig, err := buildHubConfig(hubURL, *useCAAuth, *tlsClientInsecure)
+	hubConfig, err := buildHubConfig(hubURL, *useCertificateAuth, *tlsClientInsecure)
 	if err != nil {
 		klog.ErrorS(err, "error has occurred building kubernetes client configuration for hub")
 		os.Exit(1)
--- a/cmd/memberagent/main.go
+++ b/cmd/memberagent/main.go
@@ -71,10 +71,9 @@ func main() {
 		klog.ErrorS(errors.New("hub server api cannot be empty"), "error has occurred retrieving HUB_SERVER_URL")
 		os.Exit(1)
 	}
-	tokenFilePath := os.Getenv("CONFIG_PATH")
-
-	if !*useCAAuth && tokenFilePath == "" {
-		klog.ErrorS(errors.New("hub token file path cannot be empty if CA auth not used"), "error has occurred retrieving CONFIG_PATH")
+	hubConfig, err := buildHubConfig(hubURL, *useCAAuth, *tlsClientInsecure)
+	if err != nil {
+		klog.ErrorS(err, "error has occurred building kubernetes client configuration for hub")
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/memberagent/main.go
```diff
@@ -71,10 +71,9 @@ func main() {
 		klog.ErrorS(errors.New("hub server api cannot be empty"), "error has occurred retrieving HUB_SERVER_URL")
 		os.Exit(1)
 	}
-	tokenFilePath := os.Getenv("CONFIG_PATH")
-
-	if !*useCAAuth && tokenFilePath == "" {
-		klog.ErrorS(errors.New("hub token file path cannot be empty if CA auth not used"), "error has occurred retrieving CONFIG_PATH")
+	hubConfig, err := buildHubConfig(hubURL, *useCAAuth, *tlsClientInsecure)
```
jim-minter, line 74: nit: even though we should not rename the "use-ca-auth" argument, it would be nice to rename the variable `useCAAuth` to `useCertificateAuth`, and rename the description `"Use identity and CA bundle to authenticate the member agent."` to `"Use key and certificate to authenticate the member agent."`

cmd/memberagent/main.go
```diff
@@ -184,12 +110,74 @@ func main() {
...
+		}
+		hubConfig.BearerTokenFile = tokenFilePath
+	}
+
+	hubConfig.TLSClientConfig.Insecure = tlsClientInsecure
+	if !tlsClientInsecure {
+		caBundleFile := os.Getenv("CA_BUNDLE")
+		if caBundleFile != "" {
+			hubConfig.TLSClientConfig.CAFile = caBundleFile
+		}
```
jim-minter, line 166: nit

```suggestion
		hubConfig.TLSClientConfig.CAFile = os.Getenv("CA_BUNDLE")
```

jim-minter (dismissed):
A couple of nits, but LGTM.  I have checked through the refactor carefully.  Thanks for adding the unit test.

I have also checked that our base Docker image *does* indeed include a CA cert bundle.

@ryanzhang-oss please merge?

cmd/memberagent/main.go
```diff
@@ -184,12 +110,74 @@ func main() {
...
 	}
 }
 
+func buildHubConfig(hubURL string, useCAAuth bool, tlsClientInsecure bool) (*rest.Config, error) {
+	var hubConfig = &rest.Config{
+		Host: hubURL,
+	}
+	if useCAAuth {
+		identityKeyFile := os.Getenv("IDENTITY_KEY")
+		identityCertFile := os.Getenv("IDENTITY_CERT")
```
Ealianis, line 125: nit: identityKeyFile has a different naming convention than tokenFile*Path*. Perhaps we should rename to be consistent.

Ealianis: Or update the error?
```
if identityCertFile == "" {
			err := errors.New("identity cert file ***path*** cannot be empty")
```

mingqishao: fixed

cmd/memberagent/main.go
mingqishao: fixed this comments with new variable name `useCertificateAuth` and description.

jim-minter (approved):
lgtm!

cmd/memberagent/main.go
```diff
@@ -184,12 +110,71 @@ func main() {
...
+	if !tlsClientInsecure {
+		hubConfig.TLSClientConfig.CAFile = os.Getenv("CA_BUNDLE")
+		hubCA := os.Getenv("HUB_CERTIFICATE_AUTHORITY")
+		if hubCA != "" {
+			caData, err := base64.StdEncoding.DecodeString(hubCA)
+			if err != nil {
+				klog.ErrorS(err, "cannot decode hub cluster certificate authority data")
+				return nil, err
+			}
+			hubConfig.TLSClientConfig.CAData = caData
```
ryanzhang-oss, line 171: This will override the CAFile so it's another place that we want to make sure that users know what they are doing. It's either or.

cmd/memberagent/main.go
```diff
@@ -184,12 +110,71 @@ func main() {
...
+		if err != nil {
+			klog.ErrorS(err, "cannot retrieve token file from the path %s", tokenFilePath)
+			return nil, err
+		}
+		hubConfig.BearerTokenFile = tokenFilePath
+	}
+
+	hubConfig.TLSClientConfig.Insecure = tlsClientInsecure
+	if !tlsClientInsecure {
+		hubConfig.TLSClientConfig.CAFile = os.Getenv("CA_BUNDLE")
```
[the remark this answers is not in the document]
mingqishao: This PR has been merged, so I fixed it in separate PR: #365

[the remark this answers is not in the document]
mingqishao: IMO, the original logic is specific for Linkedin, that is vendor specified and  should not be in upstream's codes. 

If CA_BUNDLE is empty, that will fallback to OS's default CA bundle. Other open sources are using same logic,  for example curl. The OS's CA bundle is always be used by default unless user specific in argument or in environment variable `CURL_CA_BUNDLE`

cmd/memberagent/main.go
mingqishao: yes, the HUB_CERTIFICATE_AUTHORITY will override the CA_BUNDLE. I do agree that will make confuse for user.

Do you want me to add a validation for that? the validation will check only one of them will be present, otherwise will return a error

PR #365, opened 2023-05-22
feat: allow member agent use OS's root certificate authority (follow up)

mingqishao:

This is following up some comments on another PR: #364 after PR merged. 

I have:

### DIFF
--- a/cmd/memberagent/main.go
+++ b/cmd/memberagent/main.go
@@ -178,8 +178,9 @@ func buildHubConfig(hubURL string, useCertificateAuth bool, tlsClientInsecure bo
 			return nil, err
 		}
 
-		hubConfig.TLSClientConfig.CAFile = caBundle
-		if hubCA != "" {
+		if caBundle != "" {
+			hubConfig.TLSClientConfig.CAFile = caBundle
+		} else if hubCA != "" {
 			caData, err := base64.StdEncoding.DecodeString(hubCA)
 			if err != nil {
 				klog.ErrorS(err, "cannot decode hub cluster certificate authority data")
@@ -188,7 +189,6 @@ func buildHubConfig(hubURL string, useCertificateAuth bool, tlsClientInsecure bo
 			hubConfig.TLSClientConfig.CAData = caData
 		}
 	}
-
 	return hubConfig, nil
 }
 
--- a/cmd/memberagent/main.go
+++ b/cmd/memberagent/main.go
@@ -164,11 +164,13 @@ func buildHubConfig(hubURL string, useCertificateAuth bool, tlsClientInsecure bo
 		if ok && caBundle == "" {
 			err := errors.New("environment variable CA_BUNDLE should not be empty")
 			klog.ErrorS(err, "failed to validate system variables")
+			return nil, err
 		}
 		hubCA, ok := os.LookupEnv("HUB_CERTIFICATE_AUTHORITY")
 		if ok && hubCA == "" {
 			err := errors.New("environment variable HUB_CERTIFICATE_AUTHORITY should not be empty")
 			klog.ErrorS(err, "failed to validate system variables")
+			return nil, err
 		}
 		if caBundle != "" && hubCA != "" {
 			err := errors.New("environment variables CA_BUNDLE and HUB_CERTIFICATE_AUTHORITY should not be set at same time")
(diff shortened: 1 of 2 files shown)


cmd/memberagent/main.go
```diff
@@ -160,8 +160,15 @@ func buildHubConfig(hubURL string, useCAAuth bool, tlsClientInsecure bool) (*res
 
 	hubConfig.TLSClientConfig.Insecure = tlsClientInsecure
 	if !tlsClientInsecure {
-		hubConfig.TLSClientConfig.CAFile = os.Getenv("CA_BUNDLE")
+		caBundle := os.Getenv("CA_BUNDLE")
```
ryanzhang-oss, line 163:
```suggestion
		caBundle, err := os.LookupEnv("CA_BUNDLE")
```

ryanzhang-oss: in this way, we can differentiate when user didn't set the env vs set it as empty. Empty is not allowed.

cmd/memberagent/main.go
```diff
@@ -160,8 +160,15 @@ func buildHubConfig(hubURL string, useCAAuth bool, tlsClientInsecure bool) (*res
...
 		hubCA := os.Getenv("HUB_CERTIFICATE_AUTHORITY")
```
ryanzhang-oss, line 164: Same here

cmd/memberagent/main.go
mingqishao: I have fixed that in [this commit](https://github.com/Azure/fleet/pull/365/commits/52b9aaa882e5156dd0dd6ef7326492be4f8677c2).

I also checked all other places are using `os.Getenv()`. Other places don't need change since there are empty string  checking for all of them.

cmd/memberagent/main.go
```diff
@@ -160,8 +160,23 @@ func buildHubConfig(hubURL string, useCAAuth bool, tlsClientInsecure bool) (*res
 
 	hubConfig.TLSClientConfig.Insecure = tlsClientInsecure
 	if !tlsClientInsecure {
-		hubConfig.TLSClientConfig.CAFile = os.Getenv("CA_BUNDLE")
-		hubCA := os.Getenv("HUB_CERTIFICATE_AUTHORITY")
+		caBundle, ok := os.LookupEnv("CA_BUNDLE")
+		if ok && caBundle == "" {
+			err := errors.New("environment variable CA_BUNDLE should not be empty")
+			klog.ErrorS(err, "failed to validate system variables")
```
ryanzhang-oss, line 166: Why not return the err here? I am not sure why we just log this as user can't see the error log.  I don't think an empty env is valid.  I assume that not setting it means using os default but setting it as "" ambiguous.

cmd/memberagent/main.go
```diff
@@ -160,8 +160,23 @@ func buildHubConfig(hubURL string, useCAAuth bool, tlsClientInsecure bool) (*res
...
+			err := errors.New("environment variable HUB_CERTIFICATE_AUTHORITY should not be empty")
+			klog.ErrorS(err, "failed to validate system variables")
+		}
+		if caBundle != "" && hubCA != "" {
+			err := errors.New("environment variables CA_BUNDLE and HUB_CERTIFICATE_AUTHORITY should not be set at same time")
+			klog.ErrorS(err, "failed to validate system variables")
+			return nil, err
+		}
+
+		hubConfig.TLSClientConfig.CAFile = caBundle
```
ryanzhang-oss, line 179: we should only set caBundle or caData

ryanzhang-oss (changes requested):

cmd/memberagent/main.go
ryanzhang-oss: yes, this is only needed when there is  a reasonable "default" value.

cmd/memberagent/main.go
mingqishao: I think that is a copy/paste bug. I fixed that and added unit test in [this commit](https://github.com/Azure/fleet/pull/365/commits/f1f019903055963bbd77be13303157da388d600f).

cmd/memberagent/main.go
mingqishao: fixed in [this commit](https://github.com/Azure/fleet/pull/365/commits/84e6d516bad972a4345acfc5894a9c2bab42998d)

ryanzhang-oss (approved):

PR #366, opened 2023-05-23
feat: Webhook for CRDs

Arvindthiru:

Webhook for fleet CRDs

Created a fleet and joined a member cluster and used kusto query to identify users and their groups. Max size of groups for any user was 3.

image

All users were either part of system:authenticated or system:masters

I have:

Added E2E


michaelawyu:
Hi Arvind! I feel that there are a few general issues about this PR, which would be best if we resolve them first. I'll add them below; we can sync at a convenient time to go through them together if you'd like:

* At this moment the webhook is validating CRD objects; but to cover the cases in the design doc, it should work on all objects
* We should check in the webhook API object along with the code

pkg/webhook/customresourcedefinition/customresourcedefinition_webhook.go
```diff
@@ -0,0 +1,65 @@
...
+	hookServer.Register(ValidationPath, &webhook.Admission{Handler: &customResourceDefintionValidator{Client: mgr.GetClient()}})
+	return nil
+}
+
+type customResourceDefintionValidator struct {
+	Client  client.Client
+	decoder *admission.Decoder
+}
+
+func (v *customResourceDefintionValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
```
michaelawyu, line 37: A very minor concern here: 

the controller runtime package is now recommending the new CustomValidator/Validator interface; though I really do not have a preference here, the traditional Handle interface would work just fine.

pkg/webhook/validation/crdvalidation.go
```diff
@@ -0,0 +1,12 @@
+package validation
+
+import "k8s.io/utils/strings/slices"
+
+var (
+	validObjectGroups = []string{"networking.fleet.azure.com", "fleet.azure.com", "multicluster.x-k8s.io"}
+)
+
+// CheckCRDGroup checks to see if the input CRD group is a fleet CRD group.
+func CheckCRDGroup(group string) bool {
```
michaelawyu, line 10: Hi Arvind! I think for this one, as mentioned in the design doc, we might need to use a vague match rather than an exact match.

pkg/webhook/validation/uservalidation.go
```diff
@@ -0,0 +1,20 @@
...
+
+const (
+	authenticatedGroup = "system:authenticated"
+	mastersGroup       = "system:masters"
+)
+
+// TODO: Get valid user names as flag and check to validate those user names.
+
+// ValidateUserGroups checks to see if user is authorized to make a request to the hub cluster's api-server.
+func ValidateUserGroups(groups []string) bool {
```
michaelawyu, line 15: This is a tricky one; As Arvind you have pointed out in our previous sync, the `authenticated` group might not be a good choice. Let's discuss this further if it is OK with you.

Arvindthiru: We only allow user from system:masters group to make changes to fleet CRDs

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	group := regexp.MustCompile(groupMatch).FindStringSubmatch(crd.Name)[1]
+	if validation.CheckCRDGroup(group) && !validation.ValidateUserForCRD(req.UserInfo) {
+		return admission.Denied(fmt.Sprintf("failed to validate user: %s in groups: %v to modify fleet CRD: %s", req.UserInfo.Username, req.UserInfo.Groups, crd.Name))
+	}
+	return admission.Allowed("")
+}
+
+func retrieveCRDGVK() string {
```
michaelawyu, line 75: Hi Arvind! A nit here: it might be simpler to directly build a GVK here.

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -0,0 +1,85 @@
...
+type fleetResourceValidator struct {
+	Client  client.Client
+	decoder *admission.Decoder
+}
+
+func (v *fleetResourceValidator) Handle(_ context.Context, req admission.Request) admission.Response {
+	var response admission.Response
+	if req.Operation == admissionv1.Create || req.Operation == admissionv1.Update || req.Operation == admissionv1.Delete {
+		switch req.Kind.String() {
+		case retrieveCRDGVK():
```
michaelawyu, line 42: Hi Arvind! This is just a nit, but we can compare GVKs directly; casting them to string has some minor overhead (string concatenation).

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -0,0 +1,85 @@
...
+			klog.V(2).InfoS("handling CRD resource", "GVK", retrieveCRDGVK())
+			response = v.handleCRD(req)
+		default:
+			klog.V(2).InfoS("resource is not monitored by fleet resource validator webhook", "GVK", req.Kind.String())
+			response = admission.Allowed("")
```
michaelawyu, line 47: Though this message is probably not read by anyone :D, it might still be better to have a short description here

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -0,0 +1,85 @@
...
+			return admission.Errored(http.StatusBadRequest, err)
+		}
+	} else {
+		if err := v.decoder.Decode(req, &crd); err != nil {
+			klog.ErrorS(err, "failed to decode request object for create/update operation", "userName", req.UserInfo.Username, "groups", req.UserInfo.Groups)
+			return admission.Errored(http.StatusBadRequest, err)
+		}
+	}
+
+	group := regexp.MustCompile(groupMatch).FindStringSubmatch(crd.Name)[1]
```
michaelawyu, line 68: Hi Arvind! Apologies if I misunderstand anything here, but the intention here was to match against GVKs (specifically groups), right? But the code is now matching against object names.

pkg/webhook/validation/crdvalidation.go
```diff
@@ -0,0 +1,12 @@
+package validation
+
+import "k8s.io/utils/strings/slices"
+
+var (
+	validObjectGroups = []string{"networking.fleet.azure.com", "fleet.azure.com", "multicluster.x-k8s.io"}
```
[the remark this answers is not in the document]
michaelawyu: Sorry for not being able to relay this to you earlier; I will ping you in the chat to schedule a regular 1:1 to improve the situation.

pkg/webhook/webhook.go
```diff
@@ -178,6 +180,28 @@ func (w *Config) createFleetWebhookConfiguration(ctx context.Context) error {
...
+				AdmissionReviewVersions: []string{"v1", "v1beta1"},
+				Rules: []admv1.RuleWithOperations{
+					{
+						Operations: []admv1.OperationType{
+							admv1.Create,
+							admv1.Update,
+							admv1.Delete,
+						},
+						Rule: admv1.Rule{
+							APIGroups:   []string{"apiextensions.k8s.io"},
```
michaelawyu, line 197: Hi Arvind! Another nit here: since we are using references instead of hard-coded values in the actual handler logic; do you want to apply it here as well? Or course I am also OK with top-level constants if you may.

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
michaelawyu: Of course it's not wrong considering that the name always follows the same pattern, but we do have direct access to GVK in admission request.

test/e2e/webhook_test.go
```diff
@@ -460,3 +463,145 @@ var _ = Describe("Fleet's Hub cluster webhook tests", func() {
 		})
 	})
 })
+
+var _ = Describe("Fleet's CRD Resource Handler webhook tests", func() {
```
michaelawyu, line 467: Hi Arvind! Just a few nits:

* There are many repeated constants in the code right now (e.g, `test-user`, etc.); it could help if they are extracted out
* Though there's nothing wrong comparing error messages, it might help to check against [status codes (reasons)](https://pkg.go.dev/k8s.io/apimachinery@v0.27.2/pkg/apis/meta/v1#StatusReason)
* Style-wise, gomega has a To(Succeed()) matcher, which might be simpler than checking for nil errors.

pkg/webhook/validation/crdvalidation.go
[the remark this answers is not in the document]
Arvindthiru: The API group change is not finalized can we handle that in a separate PR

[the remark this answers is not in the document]
michaelawyu: Hi Arvind! I think the PR is just merged earlier today, PTAL.

pkg/webhook/add_customresourcedefinition.go
```diff
@@ -0,0 +1,8 @@
+package webhook
```
michaelawyu, line 1: Hi Arvind! If I understand it correctly, should this file be named `add_resourcehandler.go` instead?

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
michaelawyu: Hi Arvind! For this one, is there any concern over checking the GVK directly?

michaelawyu:
Hi Arvind! I think everything looks good; for the remaining comments, if you have any concern, please let me know.

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
Arvindthiru: The GVK and GVR contained in the request contains the info fro CustomResourceDefintiion itself

pkg/webhook/validation/crdvalidation.go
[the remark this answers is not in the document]
Arvindthiru: Added placement.karavel.io to list of fleet CRD groups checked

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
michaelawyu: Hi Arvind! In this case, would you mind adding a comment here to explain the reasoning, specifically the naming pattern CRD uses? Thanks!

michaelawyu (approved):
LGTM ;) Thanks!

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// This regex works because every CRD name in kubernetes follows this pattern <plural>.<group>.
+	group := regexp.MustCompile(groupMatch).FindStringSubmatch(crd.Name)[1]
+	if validation.CheckCRDGroup(group) && !validation.ValidateUserForCRD(req.UserInfo) {
+		return admission.Denied(fmt.Sprintf("failed to validate user: %s in groups: %v to modify fleet CRD: %s", req.UserInfo.Username, req.UserInfo.Groups, crd.Name))
+	}
+	return admission.Allowed(fmt.Sprintf("user: %s in groups: %v is allowed to modify CRD: %s", req.UserInfo.Username, req.UserInfo.Groups, crd.Name))
+}
+
+func retrieveCRDGVK() metav1.GroupVersionKind {
```
ryanzhang-oss, line 78: what does "retrieve" mean here?

pkg/webhook/add_fleetresourcehandler.go
```diff
@@ -0,0 +1,8 @@
+package webhook
```
ryanzhang-oss, line 1: not for this PR but we probably should combine those files together next time.

ryanzhang-oss (approved):

PR #367, opened 2023-05-24
interface: Policy scheduling API

ryanzhang-oss:

The definition of the improved policy scheduling API.

Fixes #

I have:

 No test needed

The code base does not pass the latest line anymore. Will need a separate PR to fix that.


apis/v1/policySnapshot_types.go
```diff
@@ -0,0 +1,134 @@
...
+	// scheduling decision based on the preferred affinity selector.
+	// An affinity score may not present if the cluster does not meet the required affinity.
+	// +optional
+	AffinityScore int32 `json:"affinityScore,omitempty"`
+
+	// TopologySpreadScore represents the priority score of the cluster calculated by the last
+	// scheduling decision based on the topology spread applied to the cluster.
+	// A priority score may not present if the cluster does not meet the topology spread.
+	// +optional
+	TopologySpreadScore int32 `json:"priorityScore,omitempty"`
```
zhiying-lin, line 106: should we create a separate structure, eg ClusterScore for them?

later, it will be easy to extend.

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,101 @@
...
+// +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ResourceSnapShot is used to store a snapshot of selected resources by a resource placement policy.
+// It is immutable. We may need to produce more than one resourceSnapshot for one ResourcePlacement to
+// get around the 1MB size limit of k8s objects.
+// Each snapshot must have a label "fleet.azure.com/snapshotGroup" with value as the snapshot index.
+// Each must have an annotation "fleet.azure.com/resource-hash" with value as the sha-256 hash
+// value of all the snapshots belong to the same snapshot index.
+type ResourceSnapShot struct {
```
zhiying-lin, line 29:
```suggestion
type ResourceSnapshot struct {
```

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,101 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:JSONPath=`.metadata.generation`,name="Gen",type=string
+// +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ResourceSnapShot is used to store a snapshot of selected resources by a resource placement policy.
+// It is immutable. We may need to produce more than one resourceSnapshot for one ResourcePlacement to
+// get around the 1MB size limit of k8s objects.
+// Each snapshot must have a label "fleet.azure.com/snapshotGroup" with value as the snapshot index.
+// Each must have an annotation "fleet.azure.com/resource-hash" with value as the sha-256 hash
```
zhiying-lin, line 27: should we put the resource hash as a field instead of annotation?

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,101 @@
...
+	Status ResourceSnapShotStatus `json:"status,omitempty"`
+}
+
+// ResourceSnapShotSpec	defines the desired state of ResourceSnapShot.
+type ResourceSnapShotSpec struct {
+	// Index is the parent index of this resource snapshot. Each index can have multiple snapshots.
+	// All the snapshots with the same index have the same label "fleet.azure.com/snapshotGroup" that
+	// points to the index.
+	// +required
+	Index int32 `json:"index"`
```
zhiying-lin, line 48: what's the index for the parent? is it a random number?
Can we use the parent name instead?

Why do we need the label in addition to this filed?

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,101 @@
...
+	// SelectedResources contains a list of resources selected by ResourceSelectors.
+	// +required
+	SelectedResources []ResourceContent `json:"selectedResources"`
+
+	// IsLatest indicates if the resourceSnapshot is the latest or not.
+	// +required
+	IsLatest bool `json:"isLatest"`
+
+	// PolicySnapShotName is the name of the policy snapshot that this resource snapshot is pointing to.
+	PolicySnapShotName int32 `json:"policySnapShotName"`
```
zhiying-lin, line 59:
```suggestion
	PolicySnapShotName string `json:"policySnapShotName"`
```

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,101 @@
...
+	// +required
+	Index int32 `json:"index"`
+
+	// SelectedResources contains a list of resources selected by ResourceSelectors.
+	// +required
+	SelectedResources []ResourceContent `json:"selectedResources"`
+
+	// IsLatest indicates if the resourceSnapshot is the latest or not.
+	// +required
+	IsLatest bool `json:"isLatest"`
```
zhiying-lin, line 56: if the resources are split into multiple snapshots, how many snapshots will be set as "latest", only parent or including its children?

apis/v1/policySnapshot_types.go
```diff
@@ -0,0 +1,134 @@
...
+// PolicySnapShotSpec defines the desired state of PolicySnapShot.
+type PolicySnapShotSpec struct {
+	// Policy defines how to select member clusters to place the selected resources.
+	// If unspecified, all the joined member clusters are selected.
+	// +optional
+	Policy *PlacementPolicy `json:"policy,omitempty"`
+
+	// IsLatest indicates if the policy is the latest or not.
+	// +required
+	IsLatest bool `json:"isLatest"`
```
michaelawyu, line 46: To reliably calculate the currently active policy snapshot, we would need ranks (unless we use sort by time, which is less reliable) + an isActive mark. Should we use labels for both marks?

apis/v1/resourceSnapshot_types.go
michaelawyu, line 48: I am wondering, would it be easier if we just use selected resources hash + rank + isActive mark?

michaelawyu: We need hash to easier determine if the snapshot is latest. The rank and isActive marks are needed for identifying latest snapshot.

apis/v1/clusterresourceplacement_types.go
```diff
@@ -0,0 +1,366 @@
...
+const (
+	// PickAllPlacementType Pick all clusters that satisfy the rules
+	PickAllPlacementType PlacementType = "PickAll"
+
+	// PickNPlacementType Pick N clusters
+	PickNPlacementType PlacementType = "PickN"
+)
+
+// RolloutStrategy describes how to replace existing application with new ones.
+// +enum
```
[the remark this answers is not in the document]
ryanzhang-oss: Our current behavior is "Recreate" so I make it explicit with this filed. We could make "recreate" default to make the entire experience backward compatible.

apis/v1/clusterresourceplacement_types.go
```diff
@@ -0,0 +1,366 @@
...
+// RolloutStrategyType identifies the type of rollout strategy
+// +enum
+type RolloutStrategyType string
+
+const (
+	// RecreateDeploymentStrategyType Remove all existing resources from the clusters before creating new ones.
+	RecreateDeploymentStrategyType RolloutStrategyType = "Recreate"
+
+	// RollingUpdateDeploymentStrategyType Create a new application only after an old instance is deleted.
+	RollingUpdateDeploymentStrategyType RolloutStrategyType = "OnDelete"
```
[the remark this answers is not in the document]
ryanzhang-oss: Sorry, I messed up here. The two types are "OnDelete" and "Recreate"

apis/v1/clusterresourceplacement_types.go
```diff
@@ -0,0 +1,366 @@
...
+	Type RolloutStrategyType `json:"type,omitempty"`
+}
+
+// RolloutStrategyType identifies the type of rollout strategy
+// +enum
+type RolloutStrategyType string
+
+const (
+	// RecreateDeploymentStrategyType Remove all existing resources from the clusters before creating new ones.
+	RecreateDeploymentStrategyType RolloutStrategyType = "Recreate"
```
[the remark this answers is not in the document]
ryanzhang-oss: I messed up the type definition. The types I want to support are in comment on line 328.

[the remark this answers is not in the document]
ryanzhang-oss: The pros of make "OnDelete" default is that it is the safer mode. The cons is that it's a breaking change from the current behavior.

apis/v1/policySnapshot_types.go
ryanzhang-oss: yes, we need label. I used a field to make it part of the schema. I wonder if we will lose the ability to do schema enforcement if we  use a label to replace a field? I know we can do validation webhook, is that the same? I am fine either way.

apis/v1/policySnapshot_types.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// Conditions is an array of current observed conditions for PolicySnapShot.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// ClusterDecisions contains a list of names of member clusters considered by the scheduler.
+	// Note that all the selected clusters must present in the list while not all the
+	// member clusters are guaranteed to be listed.
+	// +optional
+	ClusterDecisions []ClusterDecision `json:"targetClusters,omitempty"`
```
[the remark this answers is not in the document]
ryanzhang-oss: yes, definitely.

apis/v1/policySnapshot_types.go
ryanzhang-oss: good point

apis/v1/policySnapshot_types.go
```diff
@@ -0,0 +1,134 @@
...
+type ClusterDecision struct {
+	// ClusterName is the name of the ManagedCluster. If it is not empty, its value should be unique cross all
+	// placement decisions for the Placement.
+	// +kubebuilder:validation:Required
+	// +required
+	ClusterName string `json:"clusterName"`
+
+	// Selected indicates if this cluster is selected by the scheduler.
+	// +required
+	Selected bool `json:"selected"`
```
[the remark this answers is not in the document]
ryanzhang-oss: My rational is that users often wonder why a cluster is not selected. i.e. I wanted to pick A but you picked B. Why? I feel that just the selected cluster is not the whole picture.

apis/v1/resourceSnapshot_types.go
ryanzhang-oss: The tricky thing is that this is not the value of this snapshot so I am not sure if this will be confusing for a field.

apis/v1/resourceSnapshot_types.go
ryanzhang-oss: We need the label so that we can select all the snapshot belongs to the same parent.  I think I have explained the pros vs cons about index vs name in the binding definition.

apis/v1/resourceSnapshot_types.go
ryanzhang-oss: all the children are marked as "latest".  The "parent" is a logic concept. I just realized that we can use the first snapshot as "parent" and the rest are basically a linked list.

Just thinking out loud. 
My mental model for a "logic" parent is that all the snapshots name starts with the CRP name + index (i.e.  resource-index1-0, resource-index1-1)

for the list model the name will be (resource-index1, resource-index1-1, resource-index1-2)

Given that most of the snapshots will have only 1 object, it seems that the second one is better.

apis/v1/binding_types.go
```diff
@@ -0,0 +1,80 @@
...
+	// The observed status of MemberCluster.
+	// +optional
+	Status ResourceBindingStatus `json:"status,omitempty"`
+}
+
+// ResourceBindingSpec defines the desired state of ResourceBinding.
+type ResourceBindingSpec struct {
+	// ResourceSnapshotNames is the names of the resource snapshot to which
+	// that this resource binding points.
+	ResourceSnapshotIndex int32 `json:"resourceSnapshotIndex"`
```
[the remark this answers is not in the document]
ryanzhang-oss: I don't think `ResourceSnapshotName` is right though. It sounds like it's own name.

apis/v1/clusterresourceplacement_types.go
```diff
@@ -0,0 +1,366 @@
...
+	// Note that we only include 1000 failed resource placements even if there are more than 1000.
+	// +optional
+	FailedResourcePlacements []FailedResourcePlacement `json:"failedPlacements,omitempty"`
+}
+
+// ResourceIdentifier identifies one Kubernetes resource.
+type ResourceIdentifier struct {
+	// Group is the group name of the selected resource.
+	// +required
+	Group string `json:"group,omitempty"`
```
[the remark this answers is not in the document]
ryanzhang-oss: I don't think so.

apis/v1/resourceSnapshot_types.go
ryanzhang-oss: what does "rank" mean here?

apis/v1/resourceSnapshot_types.go
```diff
@@ -11,6 +11,15 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+const (
+	// ResourceIndexLabel is the label that indicate the resource snapshot index of a cluster policy.
+	ResourceIndexLabel = labelPrefix + "policyIndex"
```
zhiying-lin, line 16: does this label store the index of the resourceSnapshot for the pagination, or policySnapshot info?

i'm confused by the comment and the label value "policyIndex".

apis/v1/binding_types.go
```diff
@@ -9,72 +9,79 @@ import (
...
+// ResourceBindingSpec defines the desired state of ClusterResourceBinding.
 type ResourceBindingSpec struct {
-	// ResourceSnapshotNames is the names of the resource snapshot to which
-	// that this resource binding points.
+	// ResourcePolicyName is the name of the resource policy that this resource binding points to.
+	ResourcePolicyName string `json:"resourcePolicyName"`
+
+	// ResourceSnapshotIndex is the index of the resource snapshot to which
+	// that this resource binding points to.
 	ResourceSnapshotIndex int32 `json:"resourceSnapshotIndex"`
```
zhiying-lin: is ResourcePolicyName the policy snapshot name?

Should we only need the ResourceSnapshotName (parent/active)?
The dispatcher will fetch all the resourceSnapshots owned by the same parent and create the corresponding work.

apis/v1/binding_types.go
```diff
@@ -9,72 +9,79 @@ import (
...
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Joined")].status`,name="Joined",type=string
+// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="ResourceBindingApplied")].status`,name="Applied",type=string
 // +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
 
-// ResourceBinding is a resource created in the hub cluster to represent a member cluster within a fleet.
-type ResourceBinding struct {
+// ClusterResourceBinding is represents a scheduling decision that binds a group of resources to a cluster.
+// it must have CRPTrackingLabel that points to the cluster resource policy that creates it.
+type ClusterResourceBinding struct {
```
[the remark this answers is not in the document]
ryanzhang-oss: no, we will add a ResourceBinding since it needs to be namespace scoped. I didn't change the spec name so they can actually share the same spec.

apis/v1/policySnapshot_types.go
```diff
@@ -19,9 +27,10 @@ import (
 // +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// PolicySnapShot is used to store a snapshot of placement policy
+// ClusterPolicySnapshot is used to store a snapshot of cluster placement policy.
 // It is immutable.
-type PolicySnapShot struct {
+// It must have `CRPTrackingLabel`, `PolicyIndexLabel` and `IsLatestSnapshotLabel` labels.
+type ClusterPolicySnapshot struct {
```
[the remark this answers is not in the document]
ryanzhang-oss: I had the same thoughts but I realized they can't as the resourcePlacement is namespace scoped so they can't be the same.  AFAIK, one CRD cannot be both cluster and namespace scoped.

apis/v1/resourceSnapshot_types.go
```diff
@@ -20,13 +29,12 @@ import (
...
-// ResourceSnapShot is used to store a snapshot of selected resources by a resource placement policy.
+// ClusterResourceSnapshot is used to store a snapshot of selected resources by a resource placement policy.
 // It is immutable. We may need to produce more than one resourceSnapshot for one ResourcePlacement to
 // get around the 1MB size limit of k8s objects.
-// Each snapshot must have a label "fleet.azure.com/snapshotGroup" with value as the snapshot index.
-// Each must have an annotation "fleet.azure.com/resource-hash" with value as the sha-256 hash
-// value of all the snapshots belong to the same snapshot index.
-type ResourceSnapShot struct {
+// Each snapshot must have `CRPTrackingLabel`, `ResourceIndexLabel` and `IsLatestSnapshotLabel`
+// Each snapshot must have an annotation "fleet.azure.com/" with value as the s
```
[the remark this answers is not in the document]
ryanzhang-oss: the hash is an annotation since we don't need to select it.

apis/v1/binding_types.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// ResourceBindingSpec defines the desired state of ClusterResourceBinding.
+type ResourceBindingSpec struct {
+	// ResourcePolicyName is the name of the resource policy that this resource binding points to.
+	ResourcePolicyName string `json:"resourcePolicyName"`
+
+	// ResourceSnapshotIndex is the index of the resource snapshot to which
+	// that this resource binding points to.
+	ResourceSnapshotIndex int32 `json:"resourceSnapshotIndex"`
```
zhiying-lin, line 44:
```suggestion
	// ResourceSnapshotName is the name of the resource snapshot that this resource binding points to. If the resources are divided into multiple snapshots because of the resource size limit, it will point to the name of the parent snapshot.
	ResourceSnapshotName string `json:"resourceSnapshotName"`
```
you may miss this comment? https://github.com/Azure/fleet/pull/367#discussion_r1206637501

apis/v1/policySnapshot_types.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// PolicySnapShotSpec defines the desired state of PolicySnapShot.
+type PolicySnapShotSpec struct {
+	// Policy defines how to select member clusters to place the selected resources.
+	// If unspecified, all the joined member clusters are selected.
+	// +optional
+	Policy *PlacementPolicy `json:"policy,omitempty"`
+
+	// PolicyHash is the sha-256 hash value of the Policy field
```
zhiying-lin, line 53:
```suggestion
	// PolicyHash is the sha-256 hash value of the Policy field.
```

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,107 @@
...
+// +kubebuilder:resource:scope="Cluster",shortName=rss,categories={fleet-workload}
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:JSONPath=`.metadata.generation`,name="Gen",type=string
+// +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ClusterResourceSnapshot is used to store a snapshot of selected resources by a resource placement policy.
+// It is immutable. We may need to produce more than one resourceSnapshot for one ResourcePlacement to
+// get around the 1MB size limit of k8s objects.
+// Each snapshot must have `CRPTrackingLabel`, `ResourceIndexLabel` and `IsLatestSnapshotLabel`
```
zhiying-lin, line 35:
```suggestion
// Each snapshot must have `CRPTrackingLabel`, `ResourceIndexLabel` and `IsLatestSnapshotLabel`. If there are multiple resource snapshots for a resourcePlacement, the parent one whose name is {CRPName}-{resourceIndex} will have a label "NumberOfResourceSnapshots" to store the total number of resource snapshots.
```

apis/v1/resourceSnapshot_types.go
zhiying-lin:
```suggestion
	ResourceIndexLabel = labelPrefix + "resourceIndex"
```
missed the comment?

apis/v1/resourceSnapshot_types.go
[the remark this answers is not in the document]
zhiying-lin:
```suggestion
// Each snapshot must have an annotation "fleet.azure.com/resourceHash" with value as the the sha-256 hash value of all the snapshots belong to the same snapshot index.
```

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,105 @@
...
+	Status ResourceSnapShotStatus `json:"status,omitempty"`
+}
+
+// ResourceSnapShotSpec	defines the desired state of ResourceSnapShot.
+type ResourceSnapShotSpec struct {
+	// Index is the parent index of this resource snapshot. Each index can have multiple snapshots.
+	// All the snapshots with the same index have the same label "fleet.azure.com/snapshotGroup" that
+	// points to the index.
+	// +required
+	Index int32 `json:"index"`
```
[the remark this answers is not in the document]
zhiying-lin: forget to address this comment? we don't need the parent index any more?

We already have the ResourceIndexLabel. All the resourceSnapshots could be retrieved by querying this label and the parent snapshot (contains extra metadata, totalSnapshots) could be retrieved by using name, eg {crpName}-{resource-index}.

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,106 @@
...
+	Status ResourceSnapShotStatus `json:"status,omitempty"`
+}
+
+// ResourceSnapShotSpec	defines the desired state of ResourceSnapShot.
+type ResourceSnapShotSpec struct {
+	// Index is the parent index of this resource snapshot. Each index can have multiple snapshots.
+	// All the snapshots with the same index have the same label "fleet.azure.com/snapshotGroup" that
+	// points to the index.
+	// +required
+
```
zhiying-lin, line 56:
```suggestion
```

apis/v1/resourceSnapshot_types.go
```diff
@@ -0,0 +1,106 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:JSONPath=`.metadata.generation`,name="Gen",type=string
+// +kubebuilder:printcolumn:JSONPath=`.metadata.creationTimestamp`,name="Age",type=date
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ClusterResourceSnapshot is used to store a snapshot of selected resources by a resource placement policy.
+// It is immutable. We may need to produce more than one resourceSnapshot for one ResourcePlacement to
+// get around the 1MB size limit of k8s objects.
+// Each snapshot must have `CRPTrackingLabel`, `ResourceIndexLabel` and `IsLatestSnapshotLabel`. If there are multiple resource snapshots for a resourcePlacement, the parent one whose name is {CRPName}-{resourceIndex} will have a label "NumberOfResourceSnapshots" to store the total number of resource snapshots.
+// Each snapshot must have an annotation "fleet.azure.com/resourceHash" with value as the the sha-256 hash value of all the snapshots belong to the same snapshot index.
```
zhiying-lin, line 36:
```suggestion
// Each snapshot must have an annotation "fleet.azure.com/resourceHash" with value as the sha-256 hash value of all the snapshots belong to the same snapshot index.
```

zhiying-lin (approved):
minor comments, LGTM :)

PR #370, opened 2023-06-02
feat: scheduler (1/): add schedule queue implementation

michaelawyu:

It features the a simple scheduling queue for the scheduler and related components to use.

I have:

- [x] Unit tests

### DIFF
--- a/pkg/scheduler/queue/queue.go
+++ b/pkg/scheduler/queue/queue.go
@@ -20,14 +20,16 @@ type PolicySnapshotKeySchedulingQueueWriter interface {
 	Add(cpsKey PolicySnapshotKey)
 }
 
-// SchedulingQueue is an interface which queues PolicySnapshots for the scheduler to schedule.
+// PolicySnapshotSchedulingQueue is an interface which queues PolicySnapshots for the scheduler to schedule.
 type PolicySnapshotKeySchedulingQueue interface {
 	PolicySnapshotKeySchedulingQueueWriter
 
 	// Run starts the scheduling queue.
 	Run()
-	// Close closes the scheduling queue.
+	// Close closes the scheduling queue immediately.
 	Close()
+	// CloseWithDrain closes the scheduling queue after all items in the queue are processed.
+	CloseWithDrain()
 	// NextClusterPolicySnapshotKey returns the next-in-line PolicySnapshot key for the scheduler to schedule.
 	NextClusterPolicySnapshotKey() PolicySnapshotKey
 	// Done marks a PolicySnapshot key as done.
@@ -83,8 +85,13 @@ func WithWorkqueueName(name string) Option {
 // appropriate.
 func (sq *simplePolicySnapshotKeySchedulingQueue) Run() {}
 
-// Close shuts down the scheduling queue.
+// Close shuts down the scheduling queue immediately.
 func (sq *simplePolicySnapshotKeySchedulingQueue) Close() {
+	sq.policySanpshotWorkQueue.ShutDown()
+}
+
+// CloseWithDrain shuts down the scheduling queue and returns until all items are processed.
+func (sq *simplePolicySnapshotKeySchedulingQueue) CloseWithDrain() {
 	sq.policySanpshotWorkQueue.ShutDownWithDrain()
 }
 
--- a/pkg/scheduler/queue/queue.go
+++ b/pkg/scheduler/queue/queue.go
@@ -110,7 +110,7 @@ func (sq *simplePolicySnapshotKeySchedulingQueue) Add(cpsKey PolicySnapshotKey) 
 }
 
 // NewSimplePolicySnapshotKeySchedulingQueue returns a simplePolicySnapshotKeySchedulingQueue.
-func NewSimplePolicySnapshotKeySchedulingQueue(opts ...Option) *simplePolicySnapshotKeySchedulingQueue {
+func NewSimplePolicySnapshotKeySchedulingQueue(opts ...Option) PolicySnapshotKeySchedulingQueue {
 	options := defaultSimplePolicySnapshotKeySchedulingQueueOptions
 	for _, opt := range opts {
 		opt(&options)
--- a/pkg/scheduler/queue/queue.go
+++ b/pkg/scheduler/queue/queue.go
@@ -3,15 +3,15 @@ Copyright (c) Microsoft Corporation.
 Licensed under the MIT license.
 */
 
-// Package schedulingqueue features a scheduling queue, which keeps track of all placements for the scheduler
+// Package queue features a scheduling queue, which keeps track of all placements for the scheduler
 // to schedule.
-package schedulingqueue
+package queue
 
 import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-// PolicySnapshotKey is the unique identifier for a PolicySnapshot stored in a scheduling queue.
+// PolicySnapshotKey is the unique identifier (the name of the policy) for a PolicySnapshot stored in a scheduling queue.
 type PolicySnapshotKey string
 
 // PolicySnapshotKeySchedulingQueueWriter is an interface which allows sources, such as controllers, to add
--- a/pkg/scheduler/queue/queue.go
+++ b/pkg/scheduler/queue/queue.go
@@ -1,0 +1,122 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+// Package schedulingqueue features a scheduling queue, which keeps track of all placements for the scheduler
+// to schedule.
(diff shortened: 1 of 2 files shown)


pkg/scheduler/queue/queue.go
```diff
@@ -0,0 +1,122 @@
...
+// Run starts the scheduling queue.
+//
+// At this moment, Run is an no-op as there is only one queue present; in the future,
+// when more queues are added, Run would start goroutines that move items between queues as
+// appropriate.
+func (sq *simplePolicySnapshotKeySchedulingQueue) Run() {}
+
+// Close shuts down the scheduling queue.
+func (sq *simplePolicySnapshotKeySchedulingQueue) Close() {
+	sq.policySanpshotWorkQueue.ShutDownWithDrain()
```
ryanzhang-oss, line 88: IIRC, this call will be blocked by existing reconcile activities. Shall we have a non-blocking call too?

pkg/scheduler/queue/queue.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Package schedulingqueue features a scheduling queue, which keeps track of all placements for the scheduler
+// to schedule.
+package schedulingqueue
+
+import (
+	"k8s.io/client-go/util/workqueue"
+)
+
+// PolicySnapshotKey is the unique identifier for a PolicySnapshot stored in a scheduling queue.
```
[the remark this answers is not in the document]
michaelawyu: Hi Ryan! For the queue part only the name is needed.

pkg/scheduler/queue/queue.go
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+// Package schedulingqueue features a scheduling queue, which keeps track of all placements for the scheduler
+// to schedule.
+package schedulingqueue
```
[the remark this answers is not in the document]
michaelawyu: Hi Ryan! My guess would be that we will not grant users the capabilities to specify a queue implementation via profiles in the future, but this is just my hunch. So far there needs to be only one queue implementation I think.

pkg/scheduler/queue/queue.go
```diff
@@ -0,0 +1,122 @@
...
+	// Done marks a PolicySnapshot key as done.
+	Done(cpsKey PolicySnapshotKey)
+}
+
+// simplePolicySnapshotKeySchedulingQueue is a simple implementation of PolicySnapshotKeySchedulingQueue.
+//
+// At this moment, one single workqueue would suffice, as sources such as the cluster watcher,
+// the binding watcher, etc., can catch all changes that need the scheduler's attention.
+// In the future, when more features, e.g., inter-placement affinity/anti-affinity, are added,
+// more queues, such as a backoff queue, might become necessary.
```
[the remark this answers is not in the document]
michaelawyu: Hi Ryan! Yeah, the PolicySnapshotKeySchedulingQueue may consist of multiple queues in the future, i.e., an active queue + a back-off queue. My thought at this moment is that the regardless of the actual # of queues within, the writer should care only about putting items into the queue(s); exactly how items are moved between specific queues are the sole responsibilities of the queue implementation.

pkg/scheduler/queue/queue.go
michaelawyu: Hi Ryan! Yeah, this call will block until all work items are Done()'d. I will add a non-blocking call as well.

pkg/scheduler/queue/queue.go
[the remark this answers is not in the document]
ryanzhang-oss: btw,  why is the package name different from the directory name?

pkg/scheduler/queue/queue.go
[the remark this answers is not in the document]
ryanzhang-oss:
```suggestion
// PolicySnapshotKey is the unique identifier,  which is the name of the policy, for a PolicySnapshot stored in a scheduling queue.
```

pkg/scheduler/queue/queue.go
```diff
@@ -0,0 +1,129 @@
...
+type PolicySnapshotKey string
+
+// PolicySnapshotKeySchedulingQueueWriter is an interface which allows sources, such as controllers, to add
+// PolicySnapshots to the scheduling queue.
+type PolicySnapshotKeySchedulingQueueWriter interface {
+	Add(cpsKey PolicySnapshotKey)
+}
+
+// PolicySnapshotSchedulingQueue is an interface which queues PolicySnapshots for the scheduler to schedule.
+type PolicySnapshotKeySchedulingQueue interface {
```
zhiying-lin, line 24: is the queue designed to handle the crp policy only?

how do we handle the resourcePlacement in the future?

michaelawyu: This is a scheduler only queue (in the scheduler package), and the scheduler cares only for PolicySnapshots IIRC.

zhiying-lin: we don't need to implement it now but still need to think about it when designing the interface and at least make the queue interface easy to extend in the future.

Right now, some of the definitions are generic, eg, "PolicySnapshotKey" and some of them are specific, eg, "NextClusterPolicySnapshotKey".

Should we make them generic and later we could have two queues?

WDYT?

michaelawyu: I am not sure I followed. The API is literally called `ClusterPolicySnapshot`, and at this moment it is the only policy related API we have; do you mean that we will have a different API for `ResourcePlacement` (if we use that name) policy snapshots? have we made that decision yet?

I don't particularly mind changing the names, but do allow me to explain my stance a little bit clearer:

Firstly, we do not have `ResourcePlacement` yet. We do not know how it will evolve, and we definitely haven't thought it through. There might be many unexpected inconsistencies we need to address, and right now it is not even on the list. Premature accommodations can be harmful.

Secondly, this is an internal API and we can easily change it will a simple PR if the situation arrives. This is not a forward compatibility issue per se. I do believe that right now we should focus on the tasks on hand; I am happy to send another PR to add an interface or make the current one more generic if appropriate.

zhiying-lin: `do you mean that we will have a different API for ResourcePlacement (if we use that name) policy snapshots? `

yeah, i think so, cause `ClusterPolicySnapshot` is a cluster scope resource and later we need a namespace scope resource for the "ResourcePlacement".

@ryanzhang-oss What's your preference?

michaelawyu: I probably haven't made myself clear enough, sorry about that, but my point is not about whether we will have a new API, but that we do NOT need to think about this in a premature fashion. We can easily make changes when the situation arises with no side effects at all.

ryanzhang-oss: I think there is a possibility that we will support namespaced placement. However, the interesting thing here is that the interface is pretty much generic except one function "NextClusterPolicySnapshotKey". Everything else uses PolicySnapshot instead of ClusterPolicySnapshot. 

I think there is a chance we can unify ClusterPolicySnapshot and PolicySnapshot as both will have the exact same spec (I didn't change the spec name, just the type name). The only difference is the policyKey would have a namespace in it. With that said, I think it's also Okey to not do forward design but the current interface seems to be in between.

michaelawyu: In that case, would it be OK if I use the full name `ClusterPolicySnapshot` all along? We can then make changes (e.g., make things more generic if needed) when the situation matures.

zhiying-lin: i;m ok with using full name ClusterPolicySnapshot all along and enhancing later. if the code state is in the middle, it will cause some confusions.

pkg/scheduler/queue/queue.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// CloseWithDrain shuts down the scheduling queue and returns until all items are processed.
+func (sq *simpleClusterPolicySnapshotKeySchedulingQueue) CloseWithDrain() {
+	sq.clusterPolicySanpshotWorkQueue.ShutDownWithDrain()
+}
+
+// NextClusterPolicySnapshotKey returns the next ClusterPolicySnapshot key in the work queue for
+// the scheduler to process.
+func (sq *simpleClusterPolicySnapshotKeySchedulingQueue) NextClusterPolicySnapshotKey() ClusterPolicySnapshotKey {
```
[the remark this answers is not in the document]
michaelawyu: Hi Ryan! I have updated the code, PTAL. For now it will go with the simple route (simple wrapper); when there are more queues it will manage its own state.

ryanzhang-oss (approved):

PR #371, opened 2023-06-02
feat: scheduler (2/): add ClusterScore related structs, etc.

michaelawyu:

It features some structs for keeping scores in the scheduling process.

I have:

- [x] Unit tests

### DIFF
--- a/pkg/scheduler/framework/score.go
+++ b/pkg/scheduler/framework/score.go
@@ -1,0 +1,65 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package framework
+
+import (
+	fleetv1 "go.goms.io/fleet/apis/v1"
+)
+
+// ClusterScore is the scores the scheduler assigns to a cluster.
+type ClusterScore struct {
+	// TopologySpreadScore determines how much a binding would satisfy the topology spread
+	// constraints specified by the user.
+	TopologySpreadScore int
+	// AffinityScore determines how much a binding would satisfy the affinity terms
+	// specified by the user.
+	AffinityScore int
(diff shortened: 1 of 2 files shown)


pkg/scheduler/framework/score.go
```diff
@@ -0,0 +1,65 @@
...
+type ClusterScore struct {
+	// TopologySpreadScore determines how much a binding would satisfy the topology spread
+	// constraints specified by the user.
+	TopologySpreadScore int
+	// AffinityScore determines how much a binding would satisfy the affinity terms
+	// specified by the user.
+	AffinityScore int
+	// PriorityScore determines how much a binding would satisfy the priority terms
+	// specified by the user.
+	PriorityScore int
```
ryanzhang-oss, line 22: we don't have that now, right?

michaelawyu: Yeah. Though the PriorityScore is mentioned in the PRD and the design doc, none of the P0 plugins touch this score :(

ryanzhang-oss: please remove them for this PR then

pkg/scheduler/framework/score.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Add adds a ClusterScore to another ClusterScore.
+func (s1 *ClusterScore) Add(s2 ClusterScore) {
```
zhiying-lin, line 26: just curious when we'll need this func

michaelawyu: We need them as multiple plugins will return their own scores and it is up to the scheduler to add them together.

zhiying-lin: oh, i see, thank for the info

zhiying-lin (approved):

PR #372, opened 2023-06-02
feat: scheduler (3/): add cycle state implementation

michaelawyu:

It features the the cycle state struct which the scheduler and its plugins use to keep states.

I have:

- [x] Unit tests

Some fields are commented out as they are unused in this split PR; will uncomment in a following PR.

### DIFF
--- a/pkg/scheduler/framework/cyclestate.go
+++ b/pkg/scheduler/framework/cyclestate.go
@@ -8,8 +8,6 @@ package framework
 import (
 	"fmt"
 	"sync"
-
-	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // StateKey is the key for a state value stored in a CycleState.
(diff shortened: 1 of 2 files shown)


pkg/scheduler/framework/cyclestate.go
```diff
@@ -0,0 +1,74 @@
...
+	// skippedScorePlugins are a set of Score plugins that will be skipped during the scheduling
+	// cycle.
+	skippedScorePlugins sets.String
+
+	// expectedBatchSize is the number of bindings that the scheduler should create.
+	expectedBatchSize int
+	// batchSizeLimit is the number of bindings that the scheduler can actually create for the
+	// scheduling cycle.
+	batchSizeLimit int
+	**/
```
ryanzhang-oss, line 50: I think we can review them when we add them

ryanzhang-oss (approved):

PR #373, opened 2023-06-02
feat: add handleUpdate func for the CRP controller to support v1 API

zhiying-lin:
Create handleUpdate func for v1 CRP in order to not break the existing e2e tests.

The current handleUpdate func only creates the clusterPolicySnapshot if needed.

I have:

Add Unit tests

Will add other functionality, eg, create the clusterResourceSnapshot in the coming PR.


pkg/controllers/clusterresourceplacement/placement_controller.go
```diff
@@ -333,3 +339,116 @@ func (r *Reconciler) updatePlacementAppliedCondition(placement *fleetv1alpha1.Cl
...
+			},
+		}
+		if err := r.Client.Create(ctx, latestPolicySnapshot); err != nil {
+			klog.ErrorS(err, "Failed to create new clusterPolicySnapshot", "clusterPolicySnapshot", klog.KObj(latestPolicySnapshot))
+			return ctrl.Result{}, err
+		}
+	}
+	// create clusterResourceSnapshot
+	// update the status based on the latestPolicySnapshot status
+	// update the status based on the work
```
ryanzhang-oss, line 396: What if the latestPolicy is not "active" but policy Hash is the same as the current? It seems that we won't do anything.

What if there are more than one "active" (not suppose to happen)? Do we want to safe guard that too?

pkg/controllers/clusterresourceplacement/placement_controller.go
```diff
@@ -333,3 +339,116 @@ func (r *Reconciler) updatePlacementAppliedCondition(placement *fleetv1alpha1.Cl
...
+func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1.ClusterResourcePlacement) (ctrl.Result, error) {
+	crpKObj := klog.KObj(crp)
+	policyHash, err := generatePolicyHash(crp.Spec.Policy)
+	if err != nil {
+		klog.ErrorS(err, "Failed to generate policy hash of crp", "clusterResourcePlacement", crpKObj)
+		return ctrl.Result{}, err
+	}
+
+	clusterPolicySnapshotList := &fleetv1.ClusterPolicySnapShotList{}
+	if err := r.Client.List(ctx, clusterPolicySnapshotList, client.MatchingLabels{fleetv1.CRPTrackingLabel: crp.Name}); err != nil {
```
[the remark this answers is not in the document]
zhiying-lin: sg, my original idea is to reduce the API server call. Since it should be rare case there is no active snapshot, it's reasonable  to list with active label first. will address it.

pkg/controllers/clusterresourceplacement/placement_controller.go
```diff
@@ -333,3 +339,116 @@ func (r *Reconciler) updatePlacementAppliedCondition(placement *fleetv1alpha1.Cl
...
+			return ctrl.Result{}, err
+		}
+	}
+	// create clusterResourceSnapshot
+	// update the status based on the latestPolicySnapshot status
+	// update the status based on the work
+	return ctrl.Result{}, nil
+}
+
+func ownerReference(crp *fleetv1.ClusterResourcePlacement) metav1.OwnerReference {
```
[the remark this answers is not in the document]
zhiying-lin: there is an util func in the controllerutil package, called , 

func SetControllerReference(owner, controlled metav1.Object, scheme *runtime.Scheme) error

we don't have the scheme passed into the controller today and there will be some changes there.
The current one looks more simpler to me.

I think this util will be used by the CRP controller only. If there are other controllers, i could move it to a common place.

pkg/controllers/clusterresourceplacement/placement_controller.go
```diff
@@ -333,3 +339,116 @@ func (r *Reconciler) updatePlacementAppliedCondition(placement *fleetv1alpha1.Cl
...
+		if lastPolicyIndex < policyIndex {
+			index = i
+			lastPolicyIndex = policyIndex
+		}
+	}
+	return &list.Items[index], lastPolicyIndex, nil
+}
+
+func parsePolicyIndexFromLabel(s *fleetv1.ClusterPolicySnapshot) (int, error) {
+	indexLabel := s.Labels[fleetv1.PolicyIndexLabel]
```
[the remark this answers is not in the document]
zhiying-lin: if this label does not exist, the indexLabel will be empty.
strconv.Atoi will return error for the empty string.

So we've already checked in this case. :)

pkg/controllers/clusterresourceplacement/placement_controller.go
```diff
@@ -333,3 +339,116 @@ func (r *Reconciler) updatePlacementAppliedCondition(placement *fleetv1alpha1.Cl
...
+	v, err := strconv.Atoi(indexLabel)
+	if err != nil {
+		klog.ErrorS(errors.New("invalid clusterPolicySnapshot policyIndex label"), "Failed to parse the policy index label", "clusterPolicySnapshot", klog.KObj(s), "policyIndexLabel", indexLabel)
+		// should never happen
```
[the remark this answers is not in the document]
zhiying-lin: Assuming we will add the deny policy to reject customer's modification and create alerts for this kind of unexpected errors, our controller/system should be in the wrong state when hitting this issue.

The simple way is to rollback our controllers instead of swallowing the error.

I was thinking skipping the snapshots with invalid labels before but figured out it did not bring much benefits compared with the its complexity.

WDYT?

pkg/controllers/clusterresourceplacement/placement_controller.go
zhiying-lin: good catch, "What if the latestPolicy is not "active" but policy Hash is the same as the current? " should be handled and this corner case may happen. Will address it.

"What if there are more than one "active" (not suppose to happen)? Do we want to safe guard that too?"
Can we rely on the validating webhook & deny policy?

pkg/controllers/clusterresourceplacement/placement_controller.go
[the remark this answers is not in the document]
ryanzhang-oss: yeah, those cases should not happen especially when we will have webhooks in place. But the question is still what should do about those? Here are some options

- Don't check those "should never happen" case in the code at all. Let the system crash which will trigger alerts.
- Check and raise alerts in the code directly (not sure if there is a way)
- Check and add an error metrics and alert based on those metrics

I am just not sure that returning an error is a good way to handle this type of "should never happen" error since it will stuck and the system can't recover on its own.

pkg/controllers/clusterresourceplacement/placement_controller.go
```diff
@@ -333,3 +339,153 @@ func (r *Reconciler) updatePlacementAppliedCondition(placement *fleetv1alpha1.Cl
...
+	if err != nil {
+		klog.ErrorS(err, "Failed to generate policy hash of crp", "clusterResourcePlacement", crpKObj)
+		return ctrl.Result{}, err
+	}
+
+	latestPolicySnapshot, latestPolicySnapshotIndex, err := r.lookupLatestClusterPolicySnapshot(ctx, crp)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
```
ryanzhang-oss, line 358:
```suggestion
// Mark the last policy snapshot as in active if it is different from what we have now
```

pkg/controllers/clusterresourceplacement/placement_controller.go
```diff
@@ -333,3 +339,153 @@ func (r *Reconciler) updatePlacementAppliedCondition(placement *fleetv1alpha1.Cl
...
+		klog.ErrorS(err, "It should never happen", "clusterResourcePlacement", crpKObj)
+		return nil, -1, err
+	}
+	// When there are no active snapshots, find the one who has the largest policy index.
+	if err := r.Client.List(ctx, snapshotList, client.MatchingLabels{fleetv1.CRPTrackingLabel: crp.Name}); err != nil {
+		klog.ErrorS(err, "Failed to list all clusterPolicySnapshots", "clusterResourcePlacement", crpKObj)
+		return nil, -1, err
+	}
+	if len(snapshotList.Items) == 0 {
+		return nil, -1, nil
```
ryanzhang-oss, line 457: This means our index starts from 0, first policy snapshot will be  "crpxxxx-0". Is this what you intended?

zhiying-lin: yeah, will add a comment to indicate that

pkg/controllers/clusterresourceplacement/placement_controller.go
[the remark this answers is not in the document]
zhiying-lin: Suppose it will be used by the clusterResourceSnapshot too.

Will change to use the controllerutil  package.

pkg/controllers/clusterresourceplacement/placement_controller.go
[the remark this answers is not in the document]
zhiying-lin: agree with you. When we hit these cases, oncall should be involved.
Prefer to take option3 and will add the metrics in a separate PR.

option 1 will impact the whole service and the error may be only caused by one CR.
The other CRs could work correctly.

pkg/utils/controller/controller.go
```diff
@@ -27,6 +28,20 @@ const (
...
+var (
+	// ErrUnexpectedBehavior indicates the current situation is not expected.
+	// There should be something wrong with the system and cannot be recovered by itself.
+	ErrUnexpectedBehavior = errors.New("unexpected behavior which cannot be handled by the controller")
+)
+
+// NewUnexpectedBehaviorError returns ErrUnexpectedBehavior type error.
+func NewUnexpectedBehaviorError(err error) error {
+	if err != nil {
+		return fmt.Errorf("%w: %v", ErrUnexpectedBehavior, err.Error())
```
ryanzhang-oss, line 40: https://pkg.go.dev/errors

Does it matter if %w is at the beginning?   will Is(err, ErrUnexpectedBehavior) return true for NewUnexpectedBehaviorError(err)

pkg/controllers/clusterresourceplacement/placement_controller.go
```diff
@@ -333,3 +340,153 @@ func (r *Reconciler) updatePlacementAppliedCondition(placement *fleetv1alpha1.Cl
...
+		// It means there are multiple active snapshots and should never happen.
+		err := fmt.Errorf("there are %d active clusterPolicySnapshots owned by clusterResourcePlacement %v", len(snapshotList.Items), crp.Name)
+		klog.ErrorS(err, "It should never happen", "clusterResourcePlacement", crpKObj)
+		// TODO(zhiying) emit error metrics or well defined logs
+		return nil, -1, controller.NewUnexpectedBehaviorError(err)
+	}
+	// When there are no active snapshots, find the one who has the largest policy index.
+	if err := r.Client.List(ctx, snapshotList, client.MatchingLabels{fleetv1.CRPTrackingLabel: crp.Name}); err != nil {
+		klog.ErrorS(err, "Failed to list all clusterPolicySnapshots", "clusterResourcePlacement", crpKObj)
+		return nil, -1, err
```
ryanzhang-oss, line 453: as we discussed, this is a dependency (api server) failure. Since we created Unexpected error type, shall we create this too?

zhiying-lin: sure, will address

ryanzhang-oss (approved):

PR #376, opened 2023-06-05
feat: scheduler (4/): add status related structs

michaelawyu:

It features some structs for plugins to report their running status.

I have:

- [x] Unit tests

### DIFF
--- a/pkg/scheduler/framework/status.go
+++ b/pkg/scheduler/framework/status.go
@@ -1,0 +1,134 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package framework
+
+import "strings"
+
+// StatusCode is the status code of a Status, returned by a plugin.
+type StatusCode int
+
+// Pre-defined status codes.
+const (
+	// Success signals that a plugin has completed its run successfully.
+	// Note that a nil *Status is also considered as a Success Status.
+	Success StatusCode = iota
+	// internalError signals that a plugin has encountered an internal error.
+	// Note that this status code is NOT exported; to return an internalError status, use the
+	// FromError() call.
+	internalError
+	// ClusterUnschedulable signals that a plugin has found that a placement should not be bound
+	// to a specific cluster.
+	ClusterUnschedulable
+	// PreSkip signals that a plugin should be skipped in the following stage. This is returned
+	// by plugins at PreFilture or PreScore stages only, to save some overhead.
+	PreSkip
+)
+
+var statusCodeNames = []string{"Success", "InternalError", "ClusterUnschedulable", "PreSkip"}
+
+// Name returns the name of a status code.
+func (sc StatusCode) Name() string {
+	return statusCodeNames[sc]
+}
+
+// Status is the result yielded by a plugin.
+type Status struct {
+	// statusCode is the status code of a Status.
+	statusCode StatusCode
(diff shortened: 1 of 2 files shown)


pkg/scheduler/framework/status.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Pre-defined status codes.
+const (
+	// Success signals that a plugin has completed its run successfully.
+	// Note that a nil *Status is also considered as a Success Status.
+	Success StatusCode = iota
+	// internalError signals that a plugin has encountered an internal error.
+	// Note that this status code is NOT exported; to return an internalError status, use the
+	// FromError() call.
+	internalError
```
ryanzhang-oss, line 21: I wonder why is this internal state  in the middle of all the states?

pkg/scheduler/framework/status.go
```diff
@@ -0,0 +1,137 @@
...
+	ClusterUnschedulable
+	// Skip signals that no action is needed for the plugin to take at the stage.
+	// If this is returned by a plugin at the Pre- stages (PreFilter or PreScore), the associated
+	// plugin will be skipped at the following stages (Filter or Score) as well. This helps
+	// reduce the overhead of having to repeatedly call a plugin that is not needed for every
+	// cluster in the Filter or Score stage.
+	Skip
+)
+
+var statusCodeNames = []string{"Success", "InternalError", "ClusterUnschedulable", "PreSkip"}
```
ryanzhang-oss, line 33: Why are those values not alligned with the state name?

pkg/scheduler/framework/status.go
```diff
@@ -0,0 +1,137 @@
...
+	Success StatusCode = iota
+	// internalError signals that a plugin has encountered an internal error.
+	// Note that this status code is NOT exported; to return an internalError status, use the
+	// FromError() call.
+	internalError
+	// ClusterUnschedulable signals that a plugin has found that a placement should not be bound
+	// to a specific cluster.
+	ClusterUnschedulable
+	// Skip signals that no action is needed for the plugin to take at the stage.
+	// If this is returned by a plugin at the Pre- stages (PreFilter or PreScore), the associated
```
ryanzhang-oss, line 26: we have PreFilter/PreScore?

pkg/scheduler/framework/status.go
michaelawyu: Hi Ryan! This is not an internal state per se, but to indicate that a plugin could not complete its task due to an error. And I couldn't use `error` due to the name collision :(

pkg/scheduler/framework/status.go
michaelawyu: My bad. This inconsistency was left out when sending out the PR (it was originally called `PreSkip`). I will correct it.

pkg/scheduler/framework/status.go
michaelawyu: Yeah, this probably is better explained in the following interface PR, but the general idea is that we need the pre- stages so that some plugins can determine that they do not need to be called early (i.e., if there is no topology spread constraints, there is no need to call topology spread constraints plugin at all), otherwise per our interface these plugins will be called once per cluster ->  inspect the policy -> then terminate its call, which incurs some extra overhead.

zhiying-lin (approved):

PR #378, opened 2023-06-06
feat: add rollout controller skeleton

zhiying-lin:

add rollout controller skeleton

I have:

Will add the integration test to validate Reconcile func in a separate PR.

### DIFF
--- a/apis/v1/binding_types.go
+++ b/apis/v1/binding_types.go
@@ -38,6 +38,10 @@ type ResourceBindingSpec struct {
 	// it points to the name of the leading snapshot of the index group.
 	ResourceSnapshotName string `json:"resourceSnapshotName"`
 
+	// PolicySnapshotName is the name of the policy snapshot that this resource binding & resource snapshot points to.
+	// It is required to decide whether the binding needs to be deleted or not without talking to API server.
+	PolicySnapshotName string `json:"policySnapshotName"`
+
 	// TargetCluster is the name of the cluster that the scheduler assigns the resources to.
 	TargetCluster string `json:"targetCluster"`
 }
--- a/apis/v1/commons.go
+++ b/apis/v1/commons.go
@@ -14,7 +14,7 @@ type ClusterState string
 
 const (
 	// Unprefixed labels/annotations are reserved for end-users
-	// we will add a fleet.azure.com to designate these labels/annotations as official fleet labels/annotations.
+	// we will add a fleet.azure.com to designate these labels/annotations/finalizers as official fleet labels/annotations/finalizers.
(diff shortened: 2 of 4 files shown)


apis/v1/binding_types.go
```diff
@@ -38,6 +38,10 @@ type ResourceBindingSpec struct {
 	// it points to the name of the leading snapshot of the index group.
 	ResourceSnapshotName string `json:"resourceSnapshotName"`
 
+	// PolicySnapshotName is the name of the policy snapshot that this resource binding & resource snapshot points to.
+	// It is required to decide whether the binding needs to be deleted or not without talking to API server.
```
michaelawyu, line 42: Ah, I remember that ResourceSnapshot has a reference to PolicySnapshot, correct?

Additionally, since we cannot tell whether a PolicySnapshot is the latest from looking at the name, we have to talk to the API server (or actually, cache) any way.

zhiying-lin: yeah, but we only store the resourceSnapshot name here, but not include the policySnapshot name. :(

We will get the latest policySnapshot name by calling API server.
This field is to quickly figure out which policySnapshot the resourceSnapsht of the binding points to.
otherwise, we need to call API server by GET resourceSnapshot.

michaelawyu: Ah, but the client in use will query the cache for the GET call, right? It will never reach API server.

ryanzhang-oss: Does the controller needs to get the ResourceSnapshot? If so, then I am not sure we need this field as the policySnapshotName is in the ResourceSnapshot.

zhiying-lin: nope, it does not need to query the resourceSnapshot for each binding.

The PolicySnapshotName will be used in two places:

1. use to check if the current binding is old or not and then delete the old bindings for the ReCreate rollout strategy, in line 144, https://github.com/zhiying-lin/fleet/blob/rollout-controller/pkg/controllers/clusterresourcerollout/clusterresourcerollout_controller.go#L144
2. use to check if the the current binding is using latest policy, then update the resource  of the bindings, in line 160 https://github.com/zhiying-lin/fleet/blob/rollout-controller/pkg/controllers/clusterresourcerollout/clusterresourcerollout_controller.go#L160

michaelawyu: Please see the message in chat. We don't need this field for saving trips to API server (the trip will not happen because of caches), but we do need it for a different reason.

zhiying-lin:
The rollout controller is no longer needed. Closing this PR

PR #381, opened 2023-06-08
feat: scheduler (6/): add scheduler profile + plugin interfaces

michaelawyu:

It features the implementation for scheduler profile + the plugin interfaces to the scheduler.

I have:

- [x] Unit tests

### DIFF
--- a/pkg/scheduler/framework/profile_test.go
+++ b/pkg/scheduler/framework/profile_test.go
@@ -1,0 +1,108 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package framework
+
+import (
+	"context"
+	"testing"
+
+	fleetv1 "go.goms.io/fleet/apis/v1"
+)
+
+const (
+	dummyPluginName = "dummyAllPurposePlugin"
+)
+
+// A no-op, dummy plugin which connects to all extension points.
+type DummyAllPurposePlugin struct{}
+
+// Name returns the name of the dummy plugin.
+func (p *DummyAllPurposePlugin) Name() string {
+	return dummyPluginName
+}
+
+// PostBatch implements the PostBatch interface for the dummy plugin.
+func (p *DummyAllPurposePlugin) PostBatch(ctx context.Context, state CycleStatePluginReadWriter, policy *fleetv1.ClusterPolicySnapshot) (size int, status *Status) { //nolint:revive
+	return 1, nil
+}
+
+// PreFilter implements the PreFilter interface for the dummy plugin.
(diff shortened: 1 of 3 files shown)


pkg/scheduler/framework/profile_test.go
```diff
@@ -0,0 +1,108 @@
...
+	profile.WithPreScorePlugin(dummyAllPurposePlugin)
+	profile.WithScorePlugin(dummyAllPurposePlugin)
+
+	// Same plugin should be registered only once, even if it connects to multiple extension points.
+	if len(profile.registeredPlugins) != 1 {
+		t.Fatalf("registerPlugins len() = %d, want %d", len(profile.registeredPlugins), 1)
+	}
+	if pl, ok := profile.registeredPlugins[dummyPluginName]; !ok || pl != dummyPlugin {
+		t.Fatalf("registeredPlugins[%s] = %v, %t, want %v, %t", dummyPluginName, pl, ok, dummyPlugin, true)
+	}
```
zhiying-lin, line 72: could you please compare the structure directly?

eg,
want := map[string]Plugin{
		dummyPluginName: dummyPlugin,
	}

	if !cmp.Equal(profile.registeredPlugins, want) {
	}

pkg/scheduler/framework/profile_test.go
```diff
@@ -0,0 +1,108 @@
...
+	profile.WithFilterPlugin(dummyAllPurposePlugin)
+	profile.WithPreScorePlugin(dummyAllPurposePlugin)
+	profile.WithScorePlugin(dummyAllPurposePlugin)
+
+	// Same plugin should be registered only once, even if it connects to multiple extension points.
+	if len(profile.registeredPlugins) != 1 {
+		t.Fatalf("registerPlugins len() = %d, want %d", len(profile.registeredPlugins), 1)
+	}
+	if pl, ok := profile.registeredPlugins[dummyPluginName]; !ok || pl != dummyPlugin {
+		t.Fatalf("registeredPlugins[%s] = %v, %t, want %v, %t", dummyPluginName, pl, ok, dummyPlugin, true)
```
zhiying-lin, line 71: prefer to use errorf in this case,

"Even after your test cases encounter a failure, they should keep going for as long as possible in order to print out all of the failed checks in a single run. This way, someone who's fixing the failing test doesn't have to play whac-a-mole, fixing one bug and then re-running the test to find the next bug.

On a practical level, prefer calling t.Error over t.Fatal. When comparing several different properties of a function's output, use t.Error for each of those comparisions.
"

https://github.com/golang/go/wiki/TestComments#keep-going

pkg/scheduler/framework/profile_test.go
michaelawyu: Done.

pkg/scheduler/framework/profile_test.go
michaelawyu: No longer applicable.

zhiying-lin (approved):

PR #384, opened 2023-06-13
fix: rename api group version

zhiying-lin:

Rename API group version to placement.karavel.io/v1beta1

I have:


michaelawyu (approved):
I have some reservations regarding the core path, but other than that I've got no issues.

ryanzhang-oss (commented):
I wonder why it's moved to /apis/core? Normally the directory structure maches the group name.

apis/placement/v1alpha1/clusterresourceplacement_types.go
```diff
@@ -3,7 +3,7 @@ Copyright (c) Microsoft Corporation.
 Licensed under the MIT license.
 */
 
-package v1
+package v1alpha1
```
ryanzhang-oss, line 6: it's better to mark this as the storage version or remove that tag for all the objects. I only moved the new CRDs to storage version when I made the first change to avoid breaking existing tests

apis/placement/v1alpha1/groupversion_info.go
```diff
@@ -14,7 +14,7 @@ import (
 
 var (
 	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "fleet.azure.com", Version: "v1"}
+	GroupVersion = schema.GroupVersion{Group: "placement.fleet.io", Version: "v1alpha1"}
```
ryanzhang-oss, line 17: I am asking the PM if this is a good name

apis/placement/v1alpha1/clusterresourceplacement_types.go
zhiying-lin: updated the storage flag from the api and regenerated the manifests.
All the yaml files set "storage: true" by default.

ryanzhang-oss (approved):

PR #389, opened 2023-06-16
feat: scheduler (8/): add more scheduler framework logic

michaelawyu:

It features parts of the scheduler framework logic.

**This PR has been updated to reflect the latest design refresh; it also renames an API per the latest design doc.**

I have:

- [x] Unit tests

Per earlier agreements, testing on this PR has a limited focus/scope.

**Important**: Please postpone reviewing of this PR until the design/API change has finalized.


pkg/scheduler/framework/framework.go
```diff
@@ -145,6 +151,126 @@ func (f *framework) EventRecorder() record.EventRecorder {
...
+	// Collect all clusters.
+	//
+	// Note that clusters here are listed from the cached client for improved performance. This is
+	// safe in consistency as it is guaranteed that the scheduler will receive all events for cluster
+	// changes eventually.
+	//
+	// TO-DO (chenyu1): assign variable(s) when more logic is added.
+	_, err = f.collectClusters(ctx)
+	if err != nil {
+		klog.ErrorS(err, errorMessage, klog.KObj(policy))
```
zhiying-lin, line 184:
```suggestion
		klog.ErrorS(err, errorMessage, clusterPolicySnapshotRef)
```

use clusterPolicySnapshotRef instead?

pkg/scheduler/framework/framework.go
```diff
@@ -145,6 +151,126 @@ func (f *framework) EventRecorder() record.EventRecorder {
...
+	// Not yet fully implemented.
 	return ctrl.Result{}, nil
 }
+
+// collectClusters lists all clusters in the cache.
+func (f *framework) collectClusters(ctx context.Context) ([]fleetv1beta1.MemberCluster, error) {
+	errorFormat := "failed to collect clusters: %w"
+
+	clusterList := &fleetv1beta1.MemberClusterList{}
+	if err := f.client.List(ctx, clusterList, &client.ListOptions{}); err != nil {
```
zhiying-lin, line 239: filter out the cluster whose state = leave?

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/utils"
+	"go.goms.io/fleet/pkg/utils/controller"
+)
+
+// extractNumOfClustersFromPolicySnapshot extracts the numOfClusters from the policy snapshot.
+func extractNumOfClustersFromPolicySnapshot(policy *fleetv1beta1.ClusterPolicySnapshot) (int, error) {
```
zhiying-lin, line 20: could you please reuse the https://github.com/Azure/fleet/blob/main/pkg/controllers/clusterresourceplacement/placement_controller.go#L531?

move this func into a common place? so both crp controller &scheduler could share this func.

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -0,0 +1,74 @@
...
+	var owner string
+	for _, ownerRef := range policy.OwnerReferences {
+		if ownerRef.Kind == utils.CRPV1Beta1GVK.Kind {
+			owner = ownerRef.Name
+			break
+		}
+	}
+	if len(owner) == 0 {
+		return "", fmt.Errorf("cannot find owner reference for policy snapshot %v", policy.Name)
+	}
```
zhiying-lin, line 46:
```suggestion
	for _, ownerRef := range policy.OwnerReferences {
		if ownerRef.Kind == utils.CRPV1Beta1GVK.Kind {
			return ownerRef.Name, nil
		}
	}
	return "", fmt.Errorf("cannot find owner reference for policy snapshot %v", policy.Name)
```

pkg/scheduler/framework/framework_test.go
```diff
@@ -0,0 +1,358 @@
...
+		},
+	}
+	deletedBindingWithoutDispatcherFinalizer := &fleetv1beta1.ClusterResourceBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:              "deleted-binding-without-dispatcher-finalizer",
+			DeletionTimestamp: &timestamp,
+		},
+	}
+
+	active, deletedWith, deletedWithout := classifyBindings([]fleetv1beta1.ClusterResourceBinding{*activeBinding, *deletedBindingWithDispatcherFinalizer, *deletedBindingWithoutDispatcherFinalizer})
```
zhiying-lin, line 314: add a test for the empty binding list?

pkg/utils/common.go
```diff
@@ -59,6 +60,14 @@ const (
 	// MemberClusterFinalizer is used to make sure that we handle gc of all the member cluster resources on the hub cluster.
 	MemberClusterFinalizer = "work.fleet.azure.com/membercluster-finalizer"
 
+	// DispatcherFinalizer is added by the dispatcher to make sure that a binding can only be deleted if the dispatcher
+	// has removed all selected resources from the bound cluster.
+	DispatcherFinalizer = "fleet.io/dispatcher-cleanup"
+
+	// SchedulerFinalizer is added by the scheduler to make sure that a binding can only be deleted if the scheduler
+	// has relieved it from scheduling consideration.
+	SchedulerFinalizer = "fleet.io/scheduler-cleanup"
```
zhiying-lin, line 69: how about moving these two to v1beta1 package? use the fleetprefix + "..."

or expose the fleetprefix from the v1beta1 package?

the group name has still not been finalized yet.

pkg/utils/common.go
```diff
@@ -141,6 +150,13 @@ var (
 		Version:  corev1.SchemeGroupVersion.Version,
 		Resource: "services",
 	}
+
+	// TO-DO (chenyu1): Add more common GVRs/GVKs here.
+	CRPV1Beta1GVK = schema.GroupVersionKind{
```
zhiying-lin, line 155:
```suggestion
	CRPGVK = schema.GroupVersionKind{
```
to avoid the variable rename in the future when we upgrade the API version.

pkg/scheduler/framework/framework.go
```diff
@@ -145,6 +151,126 @@ func (f *framework) EventRecorder() record.EventRecorder {
...
-	// Not yet implemented.
+	startTime := time.Now()
+	clusterPolicySnapshotRef := klog.KObj(policy)
+	klog.V(2).InfoS("Scheduling cycle starts", "clusterPolicySnapshot", clusterPolicySnapshotRef)
+	defer func() {
+		latency := time.Since(startTime).Milliseconds()
+		klog.V(2).InfoS("Scheduling cycle ends", "clusterPolicySnapshot", clusterPolicySnapshotRef, "latency", latency)
+	}()
+
+	errorMessage := "failed to run scheduling cycle"
```
zhiying-lin, line 162:
```suggestion
	errorMessage := "Failed to run scheduling cycle"
```
Following logging convention in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md#remove-string-formatting-from-log-message

klog.InfoS("Deleted pod", ...)

pkg/scheduler/framework/framework.go
michaelawyu: The filtering was done at a later part of the function.

pkg/scheduler/framework/frameworkutils.go
michaelawyu: This was to guard against the cases where the owner name is unset. IIRC this was not part of the API schema check, though I might have remembered it wrong.

pkg/scheduler/framework/framework_test.go
michaelawyu: As mentioned the in the description, per our earlier agreement, at this moment PRs will be featuring limited test cases. More will be added in future PRs, when the development settles a little bit.

pkg/utils/common.go
michaelawyu: It's called a GVK so different versions = different objects. To avoid breaking it might be easier to add a new one rather than reusing the same.

pkg/scheduler/framework/frameworkutils.go
zhiying-lin: hmm, the suggested change will also return error when the owner reference cannot be found.
The logic should be the same.

Is my understanding correct?

michaelawyu: I mean not the owner reference struct, but the owner name field.

zhiying-lin: oh, got it, here you validate whether owner name is empty or not.

pkg/scheduler/framework/framework.go
michaelawyu: In addition, checking for Leave state will not be enough, as what (most) users want is to propagate resources to ready clusters -> Join != Ready. This is actually a complicated check, which is why I am considering the plugin approach. Besides, this way it can make use of the parallelization setup the scheduler already has.

pkg/scheduler/framework/framework.go
michaelawyu: Done.

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -0,0 +1,56 @@
...
+	return owner, nil
+}
+
+// classifyBindings categorizes bindings into three groups:
```
ryanzhang-oss, line 32: With the new rollout API and design, we might have more types of bindings. e.g. we can have bindings produced by the scheduler that is not active until the rollout controller marks it as.

pkg/scheduler/framework/framework.go
```diff
@@ -145,6 +151,126 @@ func (f *framework) EventRecorder() record.EventRecorder {
...
+	// not started, or is still in the process of, removing the same set of resources from
+	// the cluster, triggered by a recently deleted binding.
+	//
+	// TO-DO (chenyu1): assign variable(s) when more logic is added.
+	klog.V(2).InfoS("Classifying bindings", "clusterPolicySnapshot", clusterPolicySnapshotRef)
+	_, _, deletedWithoutDispatcherFinalizer := classifyBindings(bindings)
+
+	// If a binding has been marked for deletion and no longer has the dispatcher finalizer, the scheduler
+	// removes its own finalizer from it, to clear it for eventual deletion.
+	if err := f.removeSchedulerFinalizerFromBindings(ctx, deletedWithoutDispatcherFinalizer); err != nil {
```
ryanzhang-oss, line 225: should scheduler do this or the rollout controller do this?

pkg/scheduler/framework/framework.go
michaelawyu: Discussed with Ryan offline; this will be implemented as a filter plugin instead for better flexibility and parallelism.

pkg/scheduler/framework/frameworkutils.go
michaelawyu: Updated to reflect the design refresh.

pkg/scheduler/framework/framework.go
michaelawyu: Updated to reflect the design refresh.

apis/placement/v1beta1/binding_types.go
```diff
@@ -10,6 +10,49 @@ import (
...
+	// should place resources to it.
+	//
+	// Note that an active binding may not be associated with the latest scheduling policy snapshot or the latest
+	// resource snapshot. It may be up to another controller, e.g., the rolling update controller, to modify the
+	// association (if applicable). In certain cases (e.g., not enough fitting clusters), the binding may not even
+	// has a target cluster.
+	//
+	// Note also that it is not the scheduler's responsibility to add this label, even though it does
+	// reads this label to inform the scheduling cycle..
+	ActiveBindingLabel = fleetPrefix + "is-active-binding"
```
michaelawyu, line 32: The states are kept as labels for easier queries; a field would also work though. Anti-tempering is needed to make sure that the system functions correctly.

pkg/scheduler/framework/framework.go
```diff
@@ -43,7 +49,7 @@ type Framework interface {
 	Handle
 
 	// RunSchedulerCycleFor performs scheduling for a policy snapshot.
-	RunSchedulingCycleFor(ctx context.Context, policy *fleetv1beta1.ClusterPolicySnapshot, resources *fleetv1beta1.ClusterResourceSnapshot) (result ctrl.Result, err error)
+	RunSchedulingCycleFor(ctx context.Context, crpName string, policy *fleetv1beta1.SchedulingPolicySnapshot, resources *fleetv1beta1.ClusterResourceSnapshot) (result ctrl.Result, err error)
```
ryanzhang-oss, line 52: why does the scheduler need to know ClusterResoruceSnapshot? I just realized that the rollout controller can decide the resource when it activate a "creating" binding. In this way, the scheduler will totally get out of the business of resource (the part that won't affect scheduling decision)

pkg/scheduler/queue/queue.go
```diff
@@ -11,71 +11,71 @@ import (
 	"k8s.io/client-go/util/workqueue"
 )
 
-// ClusterPolicySnapshotKey is the unique identifier (its name) for a ClusterPolicySnapshot stored in a scheduling queue.
-type ClusterPolicySnapshotKey string
+// SchedulingPolicySnapshotKey is the unique identifier (its name) for a SchedulingPolicySnapshot stored in a scheduling queue.
+type SchedulingPolicySnapshotKey string
```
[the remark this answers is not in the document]
michaelawyu: Yeah, we are still holding CRP keys in the queue. Now since the design has evolved maybe I should just change it to be a CRP key queue instead.

pkg/utils/common.go
```diff
@@ -258,3 +267,20 @@ func ShouldPropagateNamespace(namespace string, skippedNamespaces map[string]boo
...
+// on a policy snapshot.
+func ExtractNumOfClustersFromPolicySnapshot(policy *fleetv1beta1.SchedulingPolicySnapshot) (int, error) {
+	numOfClustersStr, ok := policy.Annotations[fleetv1beta1.NumberOfClustersAnnotation]
+	if !ok {
+		return 0, fmt.Errorf("cannot find annotation %s", fleetv1beta1.NumberOfClustersAnnotation)
+	}
+
+	// Cast the annotation to an integer; throw an error if the cast cannot be completed or the value is negative.
+	numOfClusters, err := strconv.Atoi(numOfClustersStr)
+	if err != nil || numOfClusters < 0 {
```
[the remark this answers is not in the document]
michaelawyu: Ah, this is a case that some clarification might be needed: numOfClusters = 0 is interpreted as removing all bindings without deleting the CRP (i.e., pausing), hence a valid case, is it correct?

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -0,0 +1,60 @@
...
+	creatingWithoutTargetCluster = make([]*fleetv1beta1.ClusterResourceBinding, 0, len(bindings))
+	deleted = make([]*fleetv1beta1.ClusterResourceBinding, 0, len(bindings))
+
+	for idx := range bindings {
+		binding := bindings[idx]
+
+		_, activeLabelExists := binding.Labels[fleetv1beta1.ActiveBindingLabel]
+		_, creatingLabelExists := binding.Labels[fleetv1beta1.CreatingBindingLabel]
+		_, obsoleteLabelExists := binding.Labels[fleetv1beta1.ObsoleteBindingLabel]
+		_, noTargetClusterLabelExists := binding.Labels[fleetv1beta1.NoTargetClusterBindingLabel]
```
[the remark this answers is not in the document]
michaelawyu: Yeah, this PR update is sent before the API PR (and the design discussion with Fei) so it is out of date. I will update the PR again when the API PR is merged.

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package framework
```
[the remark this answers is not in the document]
michaelawyu: Originally it is :D. I changed it to this name as I do not want to imply that code here is utilities for all files in the package but for the framework part only.

pkg/scheduler/framework/framework.go
michaelawyu: Yeah, originally it is needed for creating the binding (there is a resource snapshot field, and in our old design there was no rollout controller). Though technically speaking the scheduler does not need to fulfill this field if we make it clear that it is the responsibility for the rollout controller to fulfill this field instead (and that the policy snapshot associated with the resource snapshot does not need to align with the one marked on the binding -> it is the outer field that counts)

michaelawyu:
Closing as the design/API has changed; another PR has been sent.

PR #390, opened 2023-06-19
feat: update the existing selectResources related func to support new API

zhiying-lin:

Differentiate the error types

- user error
- api server error (from cache or api server)
- unexpected behavior error
- expected behavior error

I have:

Need to add more UT in the future

### DIFF
--- a/pkg/controllers/clusterresourceplacement/resource_selector.go
+++ b/pkg/controllers/clusterresourceplacement/resource_selector.go
@@ -6,7 +6,6 @@ Licensed under the MIT license.
 package clusterresourceplacement
 
 import (
-	"context"
 	"fmt"
 	"sort"
 	"strings"
@@ -22,14 +21,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	workv1alpha1 "sigs.k8s.io/work-api/pkg/apis/v1alpha1"
 
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	fleetv1alpha1 "go.goms.io/fleet/apis/v1alpha1"
 	"go.goms.io/fleet/pkg/utils"
+	"go.goms.io/fleet/pkg/utils/controller"
 )
 
 // selectResources selects the resources according to the placement resourceSelectors.
 // It also generates an array of manifests obj based on the selected resources.
-func (r *Reconciler) selectResources(ctx context.Context, placement *fleetv1alpha1.ClusterResourcePlacement) ([]workv1alpha1.Manifest, error) {
-	selectedObjects, err := r.gatherSelectedResource(ctx, placement)
+func (r *Reconciler) selectResources(placement *fleetv1alpha1.ClusterResourcePlacement) ([]workv1alpha1.Manifest, error) {
+	selectedObjects, err := r.gatherSelectedResource(placement.GetName(), convertResourceSelector(placement.Spec.ResourceSelectors))
 	if err != nil {
 		return nil, err
 	}
@@ -56,10 +57,21 @@ func (r *Reconciler) selectResources(ctx context.Context, placement *fleetv1alph
 	return manifests, nil
 }
 
-// getSelectedResource get all the resources according to the resource selector.
-func (r *Reconciler) gatherSelectedResource(ctx context.Context, placement *fleetv1alpha1.ClusterResourcePlacement) ([]runtime.Object, error) {
+// Note: temporary solution to share the same set of utils between v1alpha1 and v1beta1 APIs so that v1alpha1 implementation
+// won't be broken. v1alpha1 implementation should be removed when new API is ready.
+// The clusterResourceSelect has no changes between different versions.
+func convertResourceSelector(old []fleetv1alpha1.ClusterResourceSelector) []fleetv1beta1.ClusterResourceSelector {
+	res := make([]fleetv1beta1.ClusterResourceSelector, len(old))
+	for i, item := range old {
+		res[i] = fleetv1beta1.ClusterResourceSelector(item)
+	}
+	return res
+}
+
+// gatherSelectedResource gets all the resources according to the resource selector.
+func (r *Reconciler) gatherSelectedResource(placement string, selectors []fleetv1beta1.ClusterResourceSelector) ([]runtime.Object, error) {
 	var resources []runtime.Object
-	for _, selector := range placement.Spec.ResourceSelectors {
+	for _, selector := range selectors {
 		gvk := schema.GroupVersionKind{
 			Group:   selector.Group,
 			Version: selector.Version,
@@ -73,12 +85,12 @@ func (r *Reconciler) gatherSelectedResource(ctx context.Context, placement *flee
 		var objs []runtime.Object
 		var err error
 		if gvk == utils.NamespaceGVK {
-			objs, err = r.fetchNamespaceResources(ctx, selector, placement.GetName())
+			objs, err = r.fetchNamespaceResources(selector, placement)
 		} else {
-			objs, err = r.fetchClusterScopedResources(ctx, selector, placement.GetName())
+			objs, err = r.fetchClusterScopedResources(selector, placement)
 		}
 		if err != nil {
-			return nil, fmt.Errorf("selector = %v: %w", selector, err)
+			return nil, err
 		}
 		resources = append(resources, objs...)
 	}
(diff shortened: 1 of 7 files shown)


pkg/utils/controller/controller.go
```diff
@@ -35,6 +35,12 @@ var (
 
 	// ErrAPIServerError indicates the error is returned by the API server.
 	ErrAPIServerError = errors.New("error returned by the API server")
+
+	// ErrInformerError indicates the error is returned by the informer or related to the informer.
+	ErrInformerError = errors.New("errors returned by the informer or related to the informer")
+
+	// ErrClientError indicates the error is caused by the client side and customer needs to take the action.
+	ErrClientError = errors.New("failed to process the request due to a client error")
```
michaelawyu, line 43: This should be ErrUserError? ClientError might be a confusing name.

pkg/utils/controller/controller.go
```diff
@@ -35,6 +35,12 @@ var (
 
 	// ErrAPIServerError indicates the error is returned by the API server.
 	ErrAPIServerError = errors.New("error returned by the API server")
+
+	// ErrInformerError indicates the error is returned by the informer or related to the informer.
+	ErrInformerError = errors.New("errors returned by the informer or related to the informer")
```
michaelawyu, line 40: I am not sure this is a valid type. You mean like a Lister error? Note that it also overlaps with the APIServerError.

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -56,10 +57,18 @@ func (r *Reconciler) selectResources(ctx context.Context, placement *fleetv1alph
 	return manifests, nil
 }
 
+func convertResourceSelector(old []fleetv1alpha1.ClusterResourceSelector) []fleetv1beta1.ClusterResourceSelector {
```
michaelawyu, line 60: Just to make sure I understand the purpose correctly, this is for temporarily adding a wrapper layer to satisfy the tests? It will be removed when the resource conversion webhook is ready?

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -120,19 +129,22 @@ func (r *Reconciler) fetchClusterScopedResources(_ context.Context, selector fle
...
 	// TODO: validator should enforce the mutual exclusiveness between the `name` and `labelSelector` fields
 	if len(selector.Name) != 0 {
 		obj, err := lister.Get(selector.Name)
 		if err != nil {
+			if errors.IsNotFound(err) {
+				return nil, nil
+			}
 			klog.ErrorS(err, "cannot get the resource", "gvr", gvr, "name", selector.Name)
-			return nil, client.IgnoreNotFound(err)
+			return nil, controller.NewInformerError(err)
```
michaelawyu, line 147: IgnoreNotFound has its purposes, I don't think it should be removed here.

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -221,19 +233,23 @@ func (r *Reconciler) fetchNamespaceResources(ctx context.Context, selector fleet
...
+		return nil, controller.NewClientError(err)
 	}
 
 	klog.V(2).InfoS("start to fetch all the resources inside a namespace", "namespace", namespaceName)
 	// select the namespace object itself
 	obj, err := r.InformerManager.Lister(utils.NamespaceGVR).Get(namespaceName)
 	if err != nil {
 		klog.ErrorS(err, "cannot get the namespace", "namespace", namespaceName)
-		return nil, client.IgnoreNotFound(err)
+		if errors.IsNotFound(err) {
```
michaelawyu, line 249: I do think a better way is to let the New function handle nil in a special manner.

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -250,18 +266,19 @@ func (r *Reconciler) fetchAllResourcesInOneNamespace(_ context.Context, namespac
...
 		}
 		if !r.InformerManager.IsInformerSynced(gvr) {
-			return nil, fmt.Errorf("informer cache for %+v is not synced yet", gvr)
+			return nil, controller.NewInformerError(fmt.Errorf("informer cache for %+v is not synced yet", gvr))
 		}
 		lister := r.InformerManager.Lister(gvr)
 		objs, err := lister.ByNamespace(namespaceName).List(labels.Everything())
 		if err != nil {
-			return nil, fmt.Errorf("cannot list all the objects of type %+v in namespace %s: %w", gvr, namespaceName, err)
+			return nil, controller.NewInformerError(fmt.Errorf("cannot list all the objects of type %+v in namespace %s: %w", gvr, namespaceName, err))
```
michaelawyu, line 274: This is where it gets puzzling. This error is a typically read error, while the one above (269) is an initialization error which indicates that the handler hasn't received all pre-sync related updates. Should they be considered of the same category?

pkg/utils/controller/controller.go
```diff
@@ -54,6 +60,22 @@ func NewAPIServerError(err error) error {
 	return ErrAPIServerError
 }
 
+// NewInformerError returns ErrInformerError type error.
+func NewInformerError(err error) error {
```
michaelawyu, line 64: Even if it receives a nil it still wraps it as an error?

pkg/controllers/clusterresourceplacement/resource_selector.go
ryanzhang-oss: are we converting CRD between api groups?

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -338,15 +354,55 @@ func generateManifest(object *unstructured.Unstructured) (*workv1alpha1.Manifest
...
+	if err != nil {
+		return nil, err
+	}
+
+	resources := make([]fleetv1beta1.ResourceContent, len(selectedObjects))
+	for i, obj := range selectedObjects {
+		unstructuredObj := obj.DeepCopyObject().(*unstructured.Unstructured)
+		rc, err := generateResourceContent(unstructuredObj)
+		if err != nil {
+			return nil, err
```
ryanzhang-oss, line 403: when will this encounter error? Should this be an "UnExpected" error?

pkg/controllers/clusterresourceplacement/resource_selector.go
zhiying-lin: nope, just temporary solution in order to not break the existing tests & logic.

i don't think we'll have the resource conversation webhook as we changed the group version.

pkg/controllers/clusterresourceplacement/resource_selector.go
zhiying-lin: moved the unexpected error from internal func to here.

pkg/utils/controller/controller.go
```diff
@@ -33,25 +34,56 @@ var (
 	// There should be something wrong with the system and cannot be recovered by itself.
 	ErrUnexpectedBehavior = errors.New("unexpected behavior which cannot be handled by the controller")
 
+	// ErrExpectedBehavior indicates the current situation is expected, which can be recovered by itself after retries.
+	ErrExpectedBehavior = errors.New("expected behavior which can be recovered by itself")
```
michaelawyu, line 38: I am not 100% sure if I understand this part, and a little bit clarification would be greatly appreciated.

What would be if there's a resource version conflict (API error)? Is this an API error or an expected error? Similarly, what would be if an error is expected with sporadic occurrences and unexpected if repeated consistently?

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -120,10 +132,10 @@ func (r *Reconciler) fetchClusterScopedResources(_ context.Context, selector fle
 		Kind:    selector.Kind,
 	}
 	if !r.InformerManager.IsClusterScopedResources(gvk) {
-		return nil, fmt.Errorf("%+v is not a cluster scoped resource", restMapping.Resource)
+		return nil, controller.NewUserError(fmt.Errorf("%+v is not a cluster scoped resource", restMapping.Resource))
 	}
 	if !r.InformerManager.IsInformerSynced(gvr) {
-		return nil, fmt.Errorf("informer cache for %+v is not synced yet", restMapping.Resource)
+		return nil, controller.NewExpectedBehaviorError(fmt.Errorf("informer cache for %+v is not synced yet", restMapping.Resource))
```
michaelawyu, line 138: This is an example of my confusion specified in an earlier comment. It could be, at this point, that the cache is simply not working fast enough and will get sync'd soon; or it could also be that there is a connection/RBAC issue that will never recover without intervention and needs attention.

pkg/utils/controller/controller.go
ryanzhang-oss: I think one way to interpret this is we can alert on too many ExpectedBehaviorError which will be treated the same as "UnExpectedError" while APIServerError will clearly say that there is something wrong with the api-server. We will look more into cluster status.

zhiying-lin: For now there are api server errors, user errors and the rest are expected/unexpected logic errors. They should have no overlap.

agree with Ryan, the alert could be set based on the occurrences or status code returned by the API server.

ryanzhang-oss (approved):

PR #393, opened 2023-06-21
feat: Webhook for member cluster

Arvindthiru:

Fixes #

I have:

UT and E2E tests


pkg/webhook/clusterresourceplacement/clusterresourceplacement_webhook.go
```diff
@@ -16,6 +16,6 @@ const (
 	ValidationPath = "/validate-fleet-azure-com-v1alpha1-clusterresourceplacement"
 )
 
-func Add(mgr manager.Manager) error {
+func Add(mgr manager.Manager, _ []string) error {
```
Arvindthiru, line 19: _ []string is used as a placeholder for now, this file will eventually be removed in follow up PRs, same goes for the separate handler files for pods and replicasets

pkg/webhook/validation/uservalidation.go
```diff
@@ -1,17 +1,42 @@
...
 
-// TODO:(Arvindthiru) Get valid usernames as flag and allow those usernames.
-
 // ValidateUserForCRD checks to see if user is authenticated to make a request to modify fleet CRDs.
 func ValidateUserForCRD(userInfo authenticationv1.UserInfo) bool {
 	return slices.Contains(userInfo.Groups, mastersGroup)
 }
+
+// ValidateUserForFleetCR checks to see if user is authenticated to make a request to modify Fleet CRs.
+func ValidateUserForFleetCR(ctx context.Context, client client.Client, whiteListedUsers []string, userInfo authenticationv1.UserInfo) bool {
```
ryanzhang-oss, line 24: is this used anywhere?

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -9,41 +9,48 @@ import (
...
 	hookServer := mgr.GetWebhookServer()
-	hookServer.Register(ValidationPath, &webhook.Admission{Handler: &fleetResourceValidator{Client: mgr.GetClient()}})
+	hookServer.Register(ValidationPath, &webhook.Admission{Handler: &FleetResourceValidator{Client: mgr.GetClient(), whiteListedUsers: whiteListedUsers}})
 	return nil
 }
 
-type fleetResourceValidator struct {
-	Client  client.Client
-	decoder *admission.Decoder
+type FleetResourceValidator struct {
```
ryanzhang-oss, line 38: Is there any reason this is now public?

pkg/webhook/validation/uservalidation.go
```diff
@@ -1,17 +1,42 @@
...
+	if slices.Contains(userInfo.Groups, mastersGroup) {
+		return true
+	}
+	if slices.Contains(whiteListedUsers, userInfo.Username) {
+		return true
+	}
+	var memberClusterList fleetv1alpha1.MemberClusterList
+	if err := client.List(ctx, &memberClusterList); err != nil {
+		klog.V(2).ErrorS(err, "failed to list member clusters")
+		return false
```
ryanzhang-oss, line 34: is this list from the cache?

pkg/webhook/validation/uservalidation.go
ryanzhang-oss: is the whiteList only useful fro FleetCR?

pkg/webhook/webhook.go
```diff
@@ -211,6 +212,28 @@ func (w *Config) createFleetWebhookConfiguration(ctx context.Context) error {
 					},
 				},
 			},
+			{
+				Name:                    "fleet.membercluster.validating",
+				ClientConfig:            w.createClientConfig(fleetv1alpha1.MemberCluster{}),
+				FailurePolicy:           &failPolicy,
```
ryanzhang-oss, line 218: We have decided to fail open in the design doc, right?

pkg/webhook/validation/uservalidation.go
Arvindthiru: This client from the manager that's constructed in hub agent's main.go where we don't pass a NewClient func, since we are not passing it looks like we get new delegating client which uses the cache on reads 
image

pkg/webhook/validation/uservalidation.go
Arvindthiru: Using whiteList for CRD check as well

ryanzhang-oss (approved):

pkg/webhook/validation/uservalidation.go
ryanzhang-oss: we probably should upgrade our controller-runtime now.
I looked at the code, I think we are reading it from cache.

PR #395, opened 2023-06-22
feat: Add fleet resource label to fleet namespace

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/apis/placement/v1beta1/commons.go
+++ b/apis/placement/v1beta1/commons.go
@@ -24,6 +24,7 @@ const (
 	// IsLatestSnapshotLabel tells if the snapshot is the latest one.
 	IsLatestSnapshotLabel = fleetPrefix + "isLatestSnapshot"
 
+	// FleetResourceLabelKey is that label that indicates the resource is a fleet resource.
 	FleetResourceLabelKey = fleetPrefix + "isFleetResource"
 )
 
--- a/apis/placement/v1beta1/commons.go
+++ b/apis/placement/v1beta1/commons.go
@@ -16,13 +16,15 @@ const (
 	// Unprefixed labels/annotations are reserved for end-users
 	// we will add a placement.azure.com to designate these labels/annotations as official fleet labels/annotations.
 	// See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions
-	FleetPrefix = "placement.azure.com/"
+	fleetPrefix = "placement.azure.com/"
 
 	// CRPTrackingLabel is the label that points to the cluster resource policy that creates a resource binding.
-	CRPTrackingLabel = FleetPrefix + "parentCRP"
+	CRPTrackingLabel = fleetPrefix + "parentCRP"
 
 	// IsLatestSnapshotLabel tells if the snapshot is the latest one.
-	IsLatestSnapshotLabel = FleetPrefix + "isLatestSnapshot"
+	IsLatestSnapshotLabel = fleetPrefix + "isLatestSnapshot"
+
+	FleetResourceLabelKey = fleetPrefix + "isFleetResource"
 )
 
 const (
--- a/apis/placement/v1beta1/commons.go
+++ b/apis/placement/v1beta1/commons.go
@@ -16,13 +16,13 @@ const (
 	// Unprefixed labels/annotations are reserved for end-users
 	// we will add a placement.azure.com to designate these labels/annotations as official fleet labels/annotations.
 	// See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions
-	fleetPrefix = "placement.azure.com/"
+	FleetPrefix = "placement.azure.com/"
 
 	// CRPTrackingLabel is the label that points to the cluster resource policy that creates a resource binding.
-	CRPTrackingLabel = fleetPrefix + "parentCRP"
+	CRPTrackingLabel = FleetPrefix + "parentCRP"
 
 	// IsLatestSnapshotLabel tells if the snapshot is the latest one.
-	IsLatestSnapshotLabel = fleetPrefix + "isLatestSnapshot"
+	IsLatestSnapshotLabel = FleetPrefix + "isLatestSnapshot"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/controllers/membercluster/membercluster_controller.go
```diff
@@ -47,6 +47,11 @@ const (
 	reasonMemberClusterJoined         = "MemberClusterJoined"
 	reasonMemberClusterLeft           = "MemberClusterLeft"
 	reasonMemberClusterUnknown        = "MemberClusterUnknown"
+
+	fleetLabelKey              = "fleet-resource"
```
ryanzhang-oss, line 51: We now have a fleet wide "prefix" https://github.com/ryanzhang-oss/fleet/blob/9434986b703d8f45ae601957629e96e9b624495c/apis/placement/v1beta1/commons.go
https://github.com/Azure/fleet/blob/9434986b703d8f45ae601957629e96e9b624495c/apis/placement/v1beta1/commons.go#L19

pkg/controllers/membercluster/membercluster_controller.go
```diff
@@ -232,6 +237,7 @@ func (r *Reconciler) syncNamespace(ctx context.Context, mc *fleetv1alpha1.Member
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            namespaceName,
 			OwnerReferences: []metav1.OwnerReference{*toOwnerReference(mc)},
+			Labels:          map[string]string{fleetLabelKey: fleetNamespaceValue},
```
ryanzhang-oss, line 240: is this the expectedNS or actually the new NS?

ryanzhang-oss: we need to do fix the // TODO now

Arvindthiru:
Misunderstood the NamespaceSelector in ValidatingWebhook Object it selects the object that our webhook should based on the namespace's labels so no need to add labels on Role and Rolebindings we create

pkg/controllers/membercluster/membercluster_controller.go
```diff
@@ -251,8 +257,17 @@ func (r *Reconciler) syncNamespace(ctx context.Context, mc *fleetv1alpha1.Member
 		return namespaceName, nil
 	}
 
-	// TODO: Update namespace if currentNS != expectedNS.
-
+	// Updates namespace if currentNS != expectedNS.
+	if cmp.Equal(currentNS.Labels, expectedNS.Labels) {
```
ryanzhang-oss, line 261: why compare only lables?

ryanzhang-oss:
How do we select Role and RoleBindings then?

> Misunderstood the NamespaceSelector in ValidatingWebhook Object it selects the object that our webhook should handle based on the object's namespace and labels on the namespace, so no need to add labels on Role and Rolebindings we create

ryanzhang-oss:
I just realized that we have new version of API but the webhooks only works for v1alpha1. How should we handle that?

Arvindthiru:
> How do we select Role and RoleBindings then?
> 
> > Misunderstood the NamespaceSelector in ValidatingWebhook Object it selects the object that our webhook should handle based on the object's namespace and labels on the namespace, so no need to add labels on Role and Rolebindings we create

Since we add the label on the namespace we can ask it only handle roles and rolebindings within the namespace with the labels we added

Arvindthiru:
> I just realized that we have new version of API but the webhooks only works for v1alpha1. How should we handle that?

We can do something like this https://github.com/Azure/fleet/pull/397

apis/placement/v1beta1/commons.go
```diff
@@ -13,16 +13,16 @@ import (
 type ClusterState string
 
 const (
-	// Unprefixed labels/annotations are reserved for end-users
-	// we will add a placement.karavel.io to designate these labels/annotations as official fleet labels/annotations.
+	// FleetPrefix placement.karavel.io will be used to designate these labels/annotations as official fleet labels/annotations.
 	// See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions
-	fleetPrefix = "placement.karavel.io/"
+	// Non-prefixed labels/annotations are reserved for end-users
+	FleetPrefix = "placement.karavel.io/"
```
michaelawyu, line 19: Hi Arvind! I am thinking, rather than exposing this variable, would it be a better option to simply add the common label here?

pkg/controllers/membercluster/membercluster_controller.go
```diff
@@ -47,6 +49,9 @@ const (
 	reasonMemberClusterJoined         = "MemberClusterJoined"
 	reasonMemberClusterLeft           = "MemberClusterLeft"
 	reasonMemberClusterUnknown        = "MemberClusterUnknown"
+
+	fleetResourceLabelKey = fleetv1beta1.FleetPrefix + "isFleetResource"
+	fleetNamespaceValue   = "fleet-namespace"
```
michaelawyu, line 54: Hi Arvind! Maybe it would make more sense, considering the name of the label, to use `true` here.

apis/placement/v1beta1/zz_generated.deepcopy.go
```diff
@@ -12,7 +12,7 @@ package v1beta1
 
 import (
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
Arvindthiru, line 15: make reviewable added this change

michaelawyu: Yeah, it happens all the time in random 😣

pkg/controllers/membercluster/membercluster_controller.go
```diff
@@ -251,8 +255,15 @@ func (r *Reconciler) syncNamespace(ctx context.Context, mc *fleetv1alpha1.Member
 		return namespaceName, nil
 	}
 
-	// TODO: Update namespace if currentNS != expectedNS.
-
+	// migration: To add new label to all existing member cluster namespaces.
+	klog.V(2).InfoS("patching namespace", "memberCluster", klog.KObj(mc), "namespace", namespaceName)
```
michaelawyu, line 259: Hi Arvind! Just a nit here, but would you mind adding a check here to make sure that the patch will not happen if the label is already present?

I understand that patch will not happen if the patch data bytes are empty, but the log emitted here might be a little confusing.

michaelawyu (dismissed):
Aside from the nit everything looks great ✨

PR #398, opened 2023-06-27
feat: create clusterResourceSnapshots

zhiying-lin:

get or create cluster resource snapshots if needed.

I have:

Unit tests

### DIFF
--- a/pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
+++ b/pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
@@ -17,8 +17,9 @@ import (
 	"go.goms.io/fleet/pkg/utils/controller"
 )
 
-func (r *Reconciler) Reconcile(_ context.Context, _ controller.QueueKey) (ctrl.Result, error) {
-	return ctrl.Result{}, nil
+func (r *Reconciler) Reconcile(ctx context.Context, _ controller.QueueKey) (ctrl.Result, error) {
+	// TODO workaround to bypass lint check
+	return r.handleUpdate(ctx, nil)
 }
 
 // handleUpdate handles the create/update clusterResourcePlacement event.
@@ -26,20 +27,52 @@ func (r *Reconciler) Reconcile(_ context.Context, _ controller.QueueKey) (ctrl.R
 // clusterPolicySnapshot status and work status.
 // If the error type is ErrUnexpectedBehavior, the controller will skip the reconciling.
 func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement) (ctrl.Result, error) {
+	latestPolicySnapshot, err := r.getOrCreateClusterPolicySnapshot(ctx, crp)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	selectedResources, err := r.selectResourcesForPlacement(crp)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	resourceSnapshotSpec := fleetv1beta1.ResourceSnapshotSpec{
+		SelectedResources:  selectedResources,
+		PolicySnapshotName: latestPolicySnapshot.Name,
+	}
+	_, err = r.getOrCreateClusterResourceSnapshot(ctx, crp, &resourceSnapshotSpec)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// update the status based on the latestPolicySnapshot status
+	// update the status based on the work
+	return ctrl.Result{}, nil
+}
+
+func (r *Reconciler) getOrCreateClusterPolicySnapshot(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement) (*fleetv1beta1.ClusterSchedulingPolicySnapshot, error) {
 	crpKObj := klog.KObj(crp)
 	schedulingPolicy := *crp.Spec.Policy // will exclude the numberOfClusters
 	schedulingPolicy.NumberOfClusters = nil
 	policyHash, err := generatePolicyHash(&schedulingPolicy)
 	if err != nil {
 		klog.ErrorS(err, "Failed to generate policy hash of crp", "clusterResourcePlacement", crpKObj)
-		return ctrl.Result{}, controller.NewUnexpectedBehaviorError(err)
+		return nil, controller.NewUnexpectedBehaviorError(err)
 	}
 
+	// latestPolicySnapshotIndex should be -1 when there is no snapshot.
 	latestPolicySnapshot, latestPolicySnapshotIndex, err := r.lookupLatestClusterPolicySnapshot(ctx, crp)
 	if err != nil {
-		return ctrl.Result{}, err
+		return nil, err
 	}
 
+	if latestPolicySnapshot != nil && string(latestPolicySnapshot.Spec.PolicyHash) == policyHash {
+		if err := r.ensureLatestPolicySnapshot(ctx, crp, latestPolicySnapshot); err != nil {
+			return nil, err
+		}
+		return latestPolicySnapshot, nil
+	}
+
+	// Need to create new snapshot when 1) there is no snapshots or 2) the latest snapshot hash != current one.
 	// mark the last policy snapshot as inactive if it is different from what we have now
 	if latestPolicySnapshot != nil &&
 		string(latestPolicySnapshot.Spec.PolicyHash) != policyHash &&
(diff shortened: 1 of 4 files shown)


pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -89,18 +113,77 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
+	latestResourceSnapshotHash := ""
+	if latestResourceSnapshot != nil {
+		resourceSnapshotKObj := klog.KObj(latestResourceSnapshot)
+		latestResourceSnapshotHash, err = parseResourceGroupHashFromAnnotation(latestResourceSnapshot)
+		if err != nil {
+			klog.ErrorS(err, "Failed to get the ResourceGroupHashAnnotation", "clusterResourceSnapshot", resourceSnapshotKObj)
+			return nil, controller.NewUnexpectedBehaviorError(err)
+		}
+		// mark the last resource snapshot as inactive if it is different from what we have now
+		if latestResourceSnapshotHash != resourceHash &&
```
michaelawyu, line 142: The logic here is a little bit difficult to read IMO. The case for lastestResourceSnapshot != nil && latestHash != resourceHash is handled to two separate branches.

michaelawyu, line 142: The logic here is a little bit difficult to read IMO. The case for lastestResourceSnapshot != nil && latestHash != resourceHash is handled in two separate branches.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -89,18 +113,77 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
+		return nil, controller.NewUnexpectedBehaviorError(err)
 	}
-	// update the status based on the latestPolicySnapshot status
-	// update the status based on the work
-	return ctrl.Result{}, nil
+	latestResourceSnapshot, latestResourceSnapshotIndex, err := r.lookupLatestResourceSnapshot(ctx, crp)
+	if err != nil {
+		return nil, err
+	}
+	latestResourceSnapshotHash := ""
```
michaelawyu, line 133: If you are already retrieving latest resource snapshot and its index in the lookup function, why not retrieve hashes there? It would simplify the logic a little bit.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -89,18 +113,77 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
+		}
+	}
+
+	if latestResourceSnapshot != nil && latestResourceSnapshotHash == resourceHash {
+		if err := r.ensureLatestResourceSnapshot(ctx, latestResourceSnapshot); err != nil {
+			return nil, err
+		}
+	} else {
+		// create a new resource snapshot
+		latestResourceSnapshotIndex++
```
michaelawyu, line 159: It implicitly couples with the returns of the lookup function (-1 -> 0).

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -17,27 +17,51 @@ import (
...
+func (r *Reconciler) Reconcile(ctx context.Context, _ controller.QueueKey) (ctrl.Result, error) {
+	// TODO workaround to bypass lint check
+	return r.handleUpdate(ctx, nil)
 }
 
 // handleUpdate handles the create/update clusterResourcePlacement event.
 // It creates corresponding clusterPolicySnapshot and clusterResourceSnapshot if needed and updates the status based on
 // clusterPolicySnapshot status and work status.
 // If the error type is ErrUnexpectedBehavior, the controller will skip the reconciling.
 func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement) (ctrl.Result, error) {
```
michaelawyu, line 29: If CRP has snapshots, shouldn't it have finalizers?

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -198,6 +298,63 @@ func (r *Reconciler) lookupLatestClusterPolicySnapshot(ctx context.Context, crp
...
+		}
+		return &snapshotList.Items[0], resourceIndex, nil
+	} else if len(snapshotList.Items) > 1 {
+		// It means there are multiple active snapshots and should never happen.
+		err := fmt.Errorf("there are %d active clusterResourceSnapshots owned by clusterResourcePlacement %v", len(snapshotList.Items), crp.Name)
+		klog.ErrorS(err, "It should never happen", "clusterResourcePlacement", crpKObj)
+		return nil, -1, controller.NewUnexpectedBehaviorError(err)
+	}
+	// When there are no active snapshots, find the one who has the largest resource index.
+	if err := r.Client.List(ctx, snapshotList, client.MatchingLabels{fleetv1beta1.CRPTrackingLabel: crp.Name}); err != nil {
```
michaelawyu, line 334: Why list it twice? The checkup here can done with a loop.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -198,6 +298,63 @@ func (r *Reconciler) lookupLatestClusterPolicySnapshot(ctx context.Context, crp
...
+		resourceIndex, err := parseResourceIndexFromLabel(&snapshotList.Items[0])
+		if err != nil {
+			klog.ErrorS(err, "Failed to parse the resource index label", "clusterResourceSnapshot", klog.KObj(&snapshotList.Items[0]))
+			return nil, -1, controller.NewUnexpectedBehaviorError(err)
+		}
+		return &snapshotList.Items[0], resourceIndex, nil
+	} else if len(snapshotList.Items) > 1 {
+		// It means there are multiple active snapshots and should never happen.
+		err := fmt.Errorf("there are %d active clusterResourceSnapshots owned by clusterResourcePlacement %v", len(snapshotList.Items), crp.Name)
+		klog.ErrorS(err, "It should never happen", "clusterResourcePlacement", crpKObj)
```
michaelawyu, line 330: Maybe it's just me but I don't think "It should never happen" is a valid error message.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -211,6 +368,19 @@ func parsePolicyIndexFromLabel(s *fleetv1beta1.ClusterPolicySnapshot) (int, erro
...
 }
 
+// parseResourceIndexFromLabel returns error when parsing the label which should never return error in production.
+func parseResourceIndexFromLabel(s *fleetv1beta1.ClusterResourceSnapshot) (int, error) {
+	indexLabel := s.Labels[fleetv1beta1.ResourceIndexLabel]
+	v, err := strconv.Atoi(indexLabel)
+	if err != nil {
+		return -1, err
+	}
+	if v < 0 {
```
michaelawyu, line 378: It's just a nit, but the two branches can be merged into one "invalid resource index" error handling I believe.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
zhiying-lin: will add it in the future PR.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
zhiying-lin: The lookup func is little complicated for now, so prefer to get the hash outside.

Refactored the code. PTAL to see if the readability is better.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
zhiying-lin: no active snapshot will be the edge case. So did not lookup the active snapshot in the local.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
zhiying-lin: refactored the code

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
zhiying-lin: added a comment there. The UT should catch the regression if any. :)

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -198,15 +308,79 @@ func (r *Reconciler) lookupLatestClusterPolicySnapshot(ctx context.Context, crp
...
+	latestSnapshotLabelMatcher := client.MatchingLabels{
+		fleetv1beta1.CRPTrackingLabel:      crp.Name,
+		fleetv1beta1.IsLatestSnapshotLabel: strconv.FormatBool(true),
+	}
+	crpKObj := klog.KObj(crp)
+	if err := r.Client.List(ctx, snapshotList, latestSnapshotLabelMatcher); err != nil {
+		klog.ErrorS(err, "Failed to list active clusterResourceSnapshots", "clusterResourcePlacement", crpKObj)
+		return nil, -1, controller.NewAPIServerError(false, err)
+	}
+	if len(snapshotList.Items) == 1 {
```
ryanzhang-oss, line 330: do we handle the multiple resourceSnapshots objects per index?

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
```diff
@@ -48,59 +81,119 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
+// TODO handle all the resources selected by placement larger than 1MB size limit of k8s objects.
+func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement, resourceSnapshotSpec *fleetv1beta1.ResourceSnapshotSpec) (*fleetv1beta1.ClusterResourceSnapshot, error) {
+	resourceHash, err := generateResourceHash(resourceSnapshotSpec)
+	if err != nil {
+		klog.ErrorS(err, "Failed to generate resource hash of crp", "clusterResourcePlacement", klog.KObj(crp))
+		return nil, controller.NewUnexpectedBehaviorError(err)
+	}
+
+	// latestResourceSnapshotIndex should be -1 when there is no snapshot.
+	latestResourceSnapshot, latestResourceSnapshotIndex, err := r.lookupLatestResourceSnapshot(ctx, crp)
```
ryanzhang-oss, line 135: the resourceSnapshot and schedulingPolicySnapshot are generated following the same rule. Do we need to create two functions to look up the last? Will generic help unite them?

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
zhiying-lin: not now, will handle it in the second phase.

pkg/controllers/clusterresourceplacement/clusterresourceplacement_controller.go
zhiying-lin: the implementation will diverge when we handle the 1MB limit for the resourceSnapshot.

So prefer not to unify them, otherwise there will be too many "if else" conditions.

ryanzhang-oss (approved):

PR #400, opened 2023-06-28
interface: updated ResourceBinding API

michaelawyu:

This PR updates the ResourceBinding API.

I have:

N/A


apis/placement/v1beta1/binding_types.go
```diff
@@ -42,8 +42,16 @@ type ResourceBindingSpec struct {
 	// it points to the name of the leading snapshot of the index group.
 	ResourceSnapshotName string `json:"resourceSnapshotName"`
 
+	// PolicySnapshotName is the name of the scheduling policy snapshot that this resource binding
+	// points to; more specifically, the scheduler creates this bindings in accordance with this
+	// scheduling policy snapshot.
+	PolicySnapshotName string `json:"policySnapshotName"`
```
ryanzhang-oss, line 48:
```suggestion
	SchedulingPolicySnapshotName string `json:"schedulingPolicySnapshotName"`
```

apis/placement/v1beta1/binding_types.go
```diff
@@ -42,8 +42,16 @@ type ResourceBindingSpec struct {
 	// it points to the name of the leading snapshot of the index group.
 	ResourceSnapshotName string `json:"resourceSnapshotName"`
 
+	// PolicySnapshotName is the name of the scheduling policy snapshot that this resource binding
```
ryanzhang-oss, line 45:
```suggestion
	// SchedulingPolicySnapshotName is the name of the scheduling policy snapshot that this resource binding
```

apis/placement/v1beta1/binding_types.go
```diff
@@ -42,8 +42,16 @@ type ResourceBindingSpec struct {
...
+	// points to; more specifically, the scheduler creates this bindings in accordance with this
+	// scheduling policy snapshot.
+	PolicySnapshotName string `json:"policySnapshotName"`
+
 	// TargetCluster is the name of the cluster that the scheduler assigns the resources to.
 	TargetCluster string `json:"targetCluster"`
+
+	// ClusterDecision explains why the scheduler makes this binding.
```
ryanzhang-oss, line 53:
```suggestion
	// ClusterDecision explains why the scheduler select this cluster.
```

apis/placement/v1beta1/binding_types.go
```diff
@@ -42,8 +42,16 @@ type ResourceBindingSpec struct {
...
 
+	// SchedulingPolicySnapshotName is the name of the scheduling policy snapshot that this resource binding
+	// points to; more specifically, the scheduler creates this bindings in accordance with this
+	// scheduling policy snapshot.
+	SchedulingPolicySnapshotName string `json:"schedulingPolicySnapshotName"`
+
 	// TargetCluster is the name of the cluster that the scheduler assigns the resources to.
 	TargetCluster string `json:"targetCluster"`
+
+	// ClusterDecision explains why the scheduler select this cluster.
```
ryanzhang-oss, line 53:
```suggestion
	// ClusterDecision explains why the scheduler selected this cluster.
```

ryanzhang-oss (approved):

ryanzhang-oss (approved):

apis/placement/v1beta1/resourcesnapshot_types.go
```diff
@@ -66,10 +66,6 @@ type ResourceSnapshotSpec struct {
 	// SelectedResources contains a list of resources selected by ResourceSelectors.
 	// +required
 	SelectedResources []ResourceContent `json:"selectedResources"`
-
-	// PolicySnapshotName is the name of the policy snapshot that this resource snapshot is pointing to.
-	// +required
-	PolicySnapshotName string `json:"policySnapshotName"`
```
zhiying-lin, line 72: just wonder why we remove this field?

For example, CRP has two policySnapshots & two resourceSnapshots. policySnapshot A selects resources A and policy Snapshot B selects resource B. policySnapshot B and resourceSnapshot B are active.

Scheduler has not created the new bindings for policyA and marked the old bindings as deleted yet.
 In this case, should rollout controller avoid updating the these old bindings by checking the policySnapshotName in the resourceSnapshot? 

If we remove this field, we cannot distinguish whether the resourceSnapshot is created for policyA or policyB. 

For the future rollback feature, should we also rollback the policy?

ryanzhang-oss: The main reason we removed this field is to completely decouple the schedulingPolicy from resource. 

> Scheduler has not created the new bindings for policyA and marked the old bindings as deleted yet.
In this case, should rollout controller avoid updating the these old bindings by checking the policySnapshotName in the resourceSnapshot?

 I am not sure if I follow this case. The policyA is old so why is scheduler creating new bindings for it? What are the old bindings for? 

> If we remove this field, we cannot distinguish whether the resourceSnapshot is created for policyA or policyB.
Isn't  the latest resourceSnapshot always point to the latest policy?

> For the future rollback feature, should we also rollback the policy?
Given now that we have two revisions. It seems that the future rollback policy has to contain both of them?

PR #402, opened 2023-06-29
feat: create a crp watcher

zhiying-lin:

The current resourceWatcher will detect the CRP changes and queue the key into the CRPController queue.

Create a separate CRP watcher to monitor the CRP changes so that resourceWatcher will be only responsible for the dynamicResource changes.

Other controllers can access the cache using k8 runtime client instead of informer manager, which is more friendly than informer manager lister func (need to do the structure converstion).

I have:

Will add integration test in the next PR


michaelawyu (approved):
LGTM.

PR #407, opened 2023-06-30
feat: work generator code

ryanzhang-oss:

Check in the overrider controller with all the test

Fixes #

I have:


apis/placement/v1beta1/commons.go
```diff
@@ -23,6 +23,9 @@ const (
 
 	// IsLatestSnapshotLabel tells if the snapshot is the latest one.
 	IsLatestSnapshotLabel = fleetPrefix + "isLatestSnapshot"
+
+	// OverriderFinalizer is used to make sure that the binding is
```
michaelawyu, line 27: Hi Ryan! This comment seems to be incomplete.

pkg/controllers/overrider/reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	workv1alpha1 "sigs.k8s.io/work-api/pkg/apis/v1alpha1"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/utils/controller"
+)
+
+// Reconciler watches new cluster policy snapshot objects creation events
```
michaelawyu, line 33: Hi Ryan! This controller should watch for binding events (and work deletion) events, if I understand it correctly.

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+		klog.ErrorS(err, "failed to get the resource snapshot from resource resourceBinding",
+			"resource resourceBinding", klog.KObj(resourceBinding), "resource snapshot name", resourceBinding.Spec.ResourceSnapshotName)
+		return nil, controller.NewAPIServerError(true, err)
+	}
+	resourceSnapshots[0] = &resourceSnapshot
+
+	// check if there are more snapshot in the same index group
+	countAnnotation := resourceSnapshot.Annotations[fleetv1beta1.NumberOfResourceSnapshotsAnnotation]
+	snapshotCount, err := strconv.Atoi(countAnnotation)
+	if err != nil {
```
zhiying-lin, line 193:
```suggestion
	if err != nil || snapshotCount < 0 {
```

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+		return nil, controller.NewUnexpectedBehaviorError(err)
```
zhiying-lin, line 194: add a log here to record the resourceSnapshot name?

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	if snapshotCount > 1 {
+		index, exist := resourceSnapshot.Labels[fleetv1beta1.ResourceIndexLabel]
+		if !exist {
+			unexpectedErr := fmt.Errorf("resource snapshot %s has no index label", resourceSnapshot.Name)
```
zhiying-lin, line 199: add a klog.ErrorS here?

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+		for i := 0; i < snapshotCount; i++ {
+			resourceSnapshots = append(resourceSnapshots, &resourceSnapshotList.Items[i])
+		}
+	}
+	return resourceSnapshots, nil
+}
+
+// upsertWork creates or updates the work for the corresponding resource snapshot
+func (r *Reconciler) upsertWork(ctx context.Context, work *workv1alpha1.Work, resourceSnapshot *fleetv1beta1.ClusterResourceSnapshot, resourceBinding *fleetv1beta1.ClusterResourceBinding) error {
+	needCreate := false
```
zhiying-lin, line 219: you could use the controllerutil.CreateOrUpdate to archive this in the 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil" package

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+	if len(works) == 0 {
+		controllerutil.RemoveFinalizer(resourceBinding, fleetv1beta1.WorkFinalizer)
+		if err = r.Client.Update(ctx, resourceBinding); err != nil {
+			klog.ErrorS(err, "failed to remove the work finalizer from resource binding", "resource binding", klog.KObj(resourceBinding))
+			return ctrl.Result{}, controller.NewAPIServerError(false, err)
+		}
+	}
+	klog.V(4).InfoS("the resource binding still has undeleted work", "resource binding", klog.KObj(resourceBinding),
+		"number of associated work", len(works))
+	return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
zhiying-lin, line 104: we've already watched the work delete event. We requeue the request for backup? 5 min is good enough?
Could you please add a note there?

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+
+// SetupWithManager sets up the controller with the Manager.
+// It watches binding events and also delete events for work.
+func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.recorder = mgr.GetEventRecorderFor("work generator")
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&fleetv1beta1.ClusterResourceBinding{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Watches(&source.Kind{Type: &workv1alpha1.Work{}}, &handler.Funcs{
+			// we only care about work delete
+			DeleteFunc: func(evt event.DeleteEvent, queue workqueue.RateLimitingInterface) {
```
zhiying-lin, line 280: Could you please add a comment to explain why?

If i understand it correctly, customer cannot create/update/delete the resources in the hub cluster reserved namespaces, so we don't need to sync the work spec.
Here we watch the work delete event to handle the binding deletion.

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+		return nil, controller.NewAPIServerError(true, err)
+	}
+	for _, work := range workList.Items {
+		existingWork[work.Name] = work.DeepCopy()
+	}
+	return existingWork, nil
+}
+
+// syncAllWork generates all the work for the resourceBinding and apply them to the
+func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding *fleetv1beta1.ClusterResourceBinding) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
michaelawyu: The concern is, what would the member agent react to this change when both works involve the same set of resources?

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+	resourceBindingObj := klog.KObj(resourceBinding)
+	// reset the manifest list regardless
+	work.Spec.Workload.Manifests = make([]workv1alpha1.Manifest, 0)
+	for _, selectedResource := range resourceSnapshot.Spec.SelectedResources {
+		work.Spec.Workload.Manifests = append(work.Spec.Workload.Manifests, workv1alpha1.Manifest{
+			RawExtension: selectedResource.RawExtension,
+		})
+	}
+
+	// upsert the work
```
michaelawyu, line 252: Hi Ryan! controller runtime package has this CreateOrUpdate method which may be helpful: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate.

pkg/controllers/workgenerator/reconciler.go
ryanzhang-oss: I looked into this, the update function seems to be more complicated to write than this version

pkg/controllers/workgenerator/reconciler.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	existingWork := make(map[string]*workv1alpha1.Work)
+	workList := &workv1alpha1.WorkList{}
+	if err := r.Client.List(ctx, workList, parentBindingLabelMatcher, nameSpaceMatcher); err != nil {
+		klog.ErrorS(err, "failed to list all the work associated with the resourceBinding", "resourceBinding", klog.KObj(resourceBinding))
+		return nil, controller.NewAPIServerError(true, err)
+	}
+	for _, work := range workList.Items {
+		existingWork[work.Name] = work.DeepCopy()
+	}
```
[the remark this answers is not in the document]
ryanzhang-oss: those are from cache, I don't want to directly modify them.

pkg/controllers/workgenerator/reconciler.go
zhiying-lin: hmm, i think it should be fine. we just need to ensure the labels, ownerreference and spec are set.
If i remember it correctly, member agent will only update the status of the work?

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,329 @@
...
+	// create/update the corresponding work for each snapshot
+	activeWork := make(map[string]bool, len(resourceSnapshots))
+	for _, snapshot := range resourceSnapshots {
+		workName := getWorkNameFromSnapshotName(snapshot.Name)
+		activeWork[workName] = true
+		// TODO(RZ): issue those requests in parallel to speed up the process
+		err = r.upsertWork(ctx, works[workName], snapshot, resourceBinding)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
zhiying-lin, line 162:
```suggestion
		if r.upsertWork(ctx, works[workName], snapshot, resourceBinding); err != nil {
			return ctrl.Result{}, err
		}
```

pkg/utils/controller/controller.go
```diff
@@ -86,6 +86,24 @@ func NewUserError(err error) error {
 	return nil
 }
 
+// NewCreateIgnoreAlreadyExistError returns ErrExpectedBehavior type error if the error is already exist.
+// Otherwise, returns ErrAPIServerError type error.
+func NewCreateIgnoreAlreadyExistError(err error) error {
+	if !apierrors.IsAlreadyExists(err) {
+		return NewAPIServerError(false, err)
+	}
+	return NewExpectedBehaviorError(err)
```
zhiying-lin, line 95: when it's alreadyExists, should we ignore the error?

pkg/utils/controller/controller.go
```diff
@@ -86,6 +86,24 @@ func NewUserError(err error) error {
...
+}
+
+// NewUpdateIgnoreConflictError returns ErrExpectedBehavior type error if the error is conflict.
+// Otherwise, returns ErrAPIServerError type error.
+func NewUpdateIgnoreConflictError(err error) error {
+	if !apierrors.IsConflict(err) {
+		return NewAPIServerError(false, err)
+	}
+	return NewExpectedBehaviorError(err)
+}
```
zhiying-lin, line 105: make it generic?

for example,
```
type IsExpectedErrorFn func(error) bool

// NewAPIServerError returns error types when accessing data from cache or API server.
func NewAPIServerOrExpectedBehaviorError(fromCache bool, err error, f IsExpectedErrorFn) error {
	if err != nil {
		if f(err) {
			return NewExpectedBehaviorError(err)
		}

		if fromCache && isUnexpectedCacheError(err) {
			return NewUnexpectedBehaviorError(err)
		}
		return fmt.Errorf("%w: %v", ErrAPIServerError, err.Error())
	}
	return nil
}
```

You can call it using NewAPIServerOrExpectedBehaviorError(false, err, apierrors.IsConflict)

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,329 @@
...
+	// check if there are more snapshot in the same index group
+	countAnnotation := resourceSnapshot.Annotations[fleetv1beta1.NumberOfResourceSnapshotsAnnotation]
+	snapshotCount, err := strconv.Atoi(countAnnotation)
+	if err != nil {
+		return nil, controller.NewUnexpectedBehaviorError(err)
+	}
+	if snapshotCount > 1 {
+		index, exist := resourceSnapshot.Labels[fleetv1beta1.ResourceIndexLabel]
+		if !exist {
+			unexpectedErr := fmt.Errorf("resource snapshot %s has no index label", resourceSnapshot.Name)
```
zhiying-lin, line 203: add log?

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,329 @@
...
+		klog.ErrorS(err, "failed to get the resource snapshot from resource resourceBinding",
+			"resource resourceBinding", klog.KObj(resourceBinding), "resource snapshot name", resourceBinding.Spec.ResourceSnapshotName)
+		return nil, controller.NewAPIServerError(true, err)
+	}
+	resourceSnapshots[resourceSnapshot.Name] = &resourceSnapshot
+
+	// check if there are more snapshot in the same index group
+	countAnnotation := resourceSnapshot.Annotations[fleetv1beta1.NumberOfResourceSnapshotsAnnotation]
+	snapshotCount, err := strconv.Atoi(countAnnotation)
+	if err != nil {
```
zhiying-lin, line 197: add log here?

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,329 @@
...
+		if !exist {
+			unexpectedErr := fmt.Errorf("resource snapshot %s has no index label", resourceSnapshot.Name)
+			return nil, controller.NewUnexpectedBehaviorError(unexpectedErr)
+		}
+		resourceIndexLabelMatcher := client.MatchingLabels{
+			fleetv1beta1.ResourceIndexLabel: index,
+		}
+		resourceSnapshotList := &fleetv1beta1.ClusterResourceSnapshotList{}
+		if err := r.Client.List(ctx, resourceSnapshotList, resourceIndexLabelMatcher); err != nil {
+			klog.ErrorS(err, "failed to list all the resource snapshot associated with the resourceBinding", "resourceBinding", klog.KObj(resourceBinding))
```
zhiying-lin, line 211:
```suggestion
			klog.ErrorS(err, "Failed to list all the resource snapshot associated with the resourceBinding", "resourceBinding", klog.KObj(resourceBinding))
```

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,329 @@
...
+		}
+	}
+
+	klog.V(2).InfoS("Successfully synced all the work associated with the resourceBinding", "resourceBinding", klog.KObj(resourceBinding))
+	return ctrl.Result{}, nil
+}
+
+// fetchAllResourceSnapshots gathers all the resource snapshots for the resource binding.
+func (r *Reconciler) fetchAllResourceSnapshots(ctx context.Context, resourceBinding *fleetv1beta1.ClusterResourceBinding) (map[string]*fleetv1beta1.ClusterResourceSnapshot, error) {
+	resourceSnapshots := make(map[string]*fleetv1beta1.ClusterResourceSnapshot, 1)
```
zhiying-lin, line 185: just curious why we need the map here? array is good enough?

pkg/utils/controller/controller.go
zhiying-lin: nvm, should proceed to update the spec

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: map is the easiest way to dedup since the "master" snapshot will be included in the following list operation again.

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: I thought error should not start with capital letter

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: The reason I didn't was that the framework will print out a log at the end of any failed reconcile so it's kinda a dup

pkg/utils/controller/controller.go
ryanzhang-oss: SG except those are only for write operations in my mind

pkg/controllers/workgenerator/controller.go
```diff
@@ -59,7 +58,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		if errors.IsNotFound(err) {
 			return ctrl.Result{}, nil
 		}
-		klog.ErrorS(err, "Failed to get the resource binding", "resource binding", bindingRef)
+		klog.ErrorS(err, "Failed to get the resource binding", "resourcebinding", bindingRef)
```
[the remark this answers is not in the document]
ryanzhang-oss: can there be resourceBinding (none cluster scoped) even if the resources are namespaced?

pkg/utils/controller/controller.go
ryanzhang-oss: btw, these two are the only cases I can think of now. If that's the case, does it still make sense to have a generic one?

pkg/controllers/workgenerator/controller.go
zhiying-lin: oh, this rule does not apply to the log strings.

`so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.`

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/controllers/workgenerator/controller.go
```diff
@@ -97,24 +96,26 @@ func (r *Reconciler) handleDelete(ctx context.Context, resourceBinding *fleetv1b
...
-// ensureFinalizer makes sure that the resourceBinding CR has a finalizer on it.
+// ensureFinalizer makes sure that the resourceSnapshot CR has a finalizer on it.
 func (r *Reconciler) ensureFinalizer(ctx context.Context, resourceBinding client.Object) error {
 	if controllerutil.ContainsFinalizer(resourceBinding, fleetv1beta1.WorkFinalizer) {
 		return nil
 	}
 	controllerutil.AddFinalizer(resourceBinding, fleetv1beta1.WorkFinalizer)
 	if err := r.Client.Update(ctx, resourceBinding); err != nil {
-		klog.ErrorS(err, "failed to add the work finalizer to resourceBinding", "resourceBinding", klog.KObj(resourceBinding))
+		klog.ErrorS(err, "failed to add the work finalizer to resourceSnapshot", "resourceSnapshot", klog.KObj(resourceBinding))
```
zhiying-lin, line 115: by mistake?

The finalizer should be added to the resourceBinding?

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,347 @@
...
+// according to the information in the binding objects.
+// TODO: incorporate an overriding policy if one exists
+type Reconciler struct {
+	client.Client
+	recorder record.EventRecorder
+}
+
+// Reconcile triggers a single binding reconcile round.
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	klog.V(2).InfoS("Start to reconcile a ClusterResourceBinding", "cluster resource binding", req.Name)
```
zhiying-lin, line 48:
```suggestion
	klog.V(2).InfoS("Start to reconcile a ClusterResourceBinding", "clusterResourceBinding", req.Name)
```

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,347 @@
...
+	startTime := time.Now()
+	bindingRef := klog.KRef(req.Namespace, req.Name)
+	// add latency log
+	defer func() {
+		klog.V(2).InfoS("ClusterResourceBinding reconciliation loop ends", "resource binding", bindingRef, "latency", time.Since(startTime).Milliseconds())
```
zhiying-lin, line 53:
```suggestion
		klog.V(2).InfoS("ClusterResourceBinding reconciliation loop ends", "resourceBinding", bindingRef, "latency", time.Since(startTime).Milliseconds())
```

apis/placement/v1beta1/binding_types.go
```diff
@@ -10,6 +10,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// WorkFinalizer is used to make sure that the binding is not deleted until the work objects it generates are all deleted.
+	WorkFinalizer = fleetPrefix + "work-cleanup"
+
+	// ParentBindingLabel is the label that contains the name of the binding that generates the work.
+	ParentBindingLabel = fleetPrefix + "resourceHash"
```
zhiying-lin, line 18:
```suggestion
	ParentBindingLabel = fleetPrefix + "resourceBinding"
```

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,347 @@
...
+	existingWork := make(map[string]*workv1alpha1.Work)
+	workList := &workv1alpha1.WorkList{}
+	if err := r.Client.List(ctx, workList, parentBindingLabelMatcher, namespaceMatcher); err != nil {
+		klog.ErrorS(err, "failed to list all the work associated with the resourceSnapshot", "resourceSnapshot", klog.KObj(resourceBinding))
+		return nil, controller.NewAPIServerError(true, err)
+	}
+	for _, work := range workList.Items {
+		existingWork[work.Name] = work.DeepCopy()
+	}
+	klog.V(4).InfoS("Get all the work associated", "numOfWork", len(existingWork), "resourceSnapshot", klog.KObj(resourceBinding))
```
zhiying-lin, line 137:
```suggestion
	klog.V(4).InfoS("Get all the work associated", "numOfWork", len(existingWork), "resourceBinding", klog.KObj(resourceBinding))
```

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,347 @@
...
+		workName, err := getWorkNameFromSnapshotName(snapshot)
+		if err != nil {
+			klog.ErrorS(err, "Encountered a mal-formatted resource binding", "resourceSnapshot", klog.KObj(resourceBinding))
+			return ctrl.Result{}, err
+		}
+		activeWork[workName] = true
+		err = r.upsertWork(ctx, works[workName], workName, snapshot, resourceBinding)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
zhiying-lin, line 167:
```suggestion
		if err := r.upsertWork(ctx, works[workName], workName, snapshot, resourceBinding); err != nil {
			return ctrl.Result{}, err
		}
```

pkg/controllers/workgenerator/controller.go
zhiying-lin: oh i see, forget we'll put the first snapshot twice.

pkg/utils/controller/controller.go
zhiying-lin: I'm fine and we could do the refactor later.

pkg/controllers/workgenerator/suite_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"context"
+	"flag"
+	"os"
+	"path/filepath"
+	"testing"
+
+	kruisev1alpha1 "github.com/openkruise/kruise/apis/apps/v1alpha1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	workv1alpha1 "sigs.k8s.io/work-api/pkg/apis/v1alpha1"
+
```
zhiying-lin, line 18: put them in the same group?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -0,0 +1,494 @@
...
+				}
+				diff = cmp.Diff(expectedOwnerReference, work.OwnerReferences)
+				Expect(diff).Should(BeEmpty(), fmt.Sprintf("work owner reference(%s) mismatch (-want +got):\n%s", work.Name, diff))
+				//inspect the work labels
+				expectedLabels := map[string]string{
+					fleetv1beta1.CRPTrackingLabel:   testCRPName,
+					fleetv1beta1.ParentBindingLabel: binding.Name,
+				}
+				diff = cmp.Diff(expectedLabels, work.Labels)
+				Expect(diff).Should(BeEmpty(), fmt.Sprintf("work label(%s) mismatch (-want +got):\n%s", work.Name, diff))
```
zhiying-lin, line 157: it's better to compare the full structure of the work by ignoring other predictable fields.

https://github.com/golang/go/wiki/TestComments#compare-full-structures

zhiying-lin (commented):
The logic LGTM but the logs needs to be updated.

- the param key should be camel
- some of the logs are missing log level
- log string should be capitalized.

pkg/controllers/workgenerator/controller.go
```diff
@@ -174,63 +173,65 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding *fleetv1be
...
+	// fetch the master snapshot first
 	resourceSnapshots := make(map[string]*fleetv1beta1.ClusterResourceSnapshot, 1)
-	resourceSnapshot := fleetv1beta1.ClusterResourceSnapshot{}
-	if err := r.Client.Get(ctx, client.ObjectKey{Name: resourceBinding.Spec.ResourceSnapshotName}, &resourceSnapshot); err != nil {
-		klog.ErrorS(err, "failed to get the resource snapshot from resource resourceSnapshot",
-			"resource resourceSnapshot", klog.KObj(resourceBinding), "resource snapshot name", resourceBinding.Spec.ResourceSnapshotName)
+	masterResourceSnapshot := fleetv1beta1.ClusterResourceSnapshot{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: resourceBinding.Spec.ResourceSnapshotName}, &masterResourceSnapshot); err != nil {
+		klog.ErrorS(err, "Failed to get the resource snapshot from resource masterResourceSnapshot",
+			"resource masterResourceSnapshot", klog.KObj(resourceBinding), "masterResourceSnapshot", resourceBinding.Spec.ResourceSnapshotName)
```
zhiying-lin, line 193:
```suggestion
			"resourceBinding", klog.KObj(resourceBinding), "masterResourceSnapshot", resourceBinding.Spec.ResourceSnapshotName)
```

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -0,0 +1,118 @@
...
+					},
+					Annotations: map[string]string{
+						fleetv1beta1.SubindexOfResourceSnapshotAnnotation: "2",
+					},
+				},
+			},
+			wantErr:    nil,
+			wantedName: "placement-2",
+		},
+		"test return error if the resource snapshot has invaide subindex": {
```
zhiying-lin, line 76:
```suggestion
		"test return error if the resource snapshot has invalid subindex": {
```

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -0,0 +1,118 @@
...
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			workName, err := getWorkNameFromSnapshotName(tt.resourceSnapshot)
+			if !errors.Is(err, tt.wantErr) {
+				t.Errorf("failed getWorkNameFromSnapshotName test `%s` error = %v, wantErr %v", name, err, tt.wantErr)
+				return
+			}
+			if workName != tt.wantedName {
+				t.Errorf("getWorkNameFromSnapshotName test `%s` workName = `%v`, wantedName `%v`", name, workName, tt.wantedName)
+				return
```
zhiying-lin, line 114: nit, redundant return here

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -0,0 +1,118 @@
+package workgenerator
```
zhiying-lin, line 1: add copyright license header?

pkg/controllers/workgenerator/controller_integration_test.go
zhiying-lin: @ryanzhang-oss  do you want to address it?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -0,0 +1,501 @@
...
```
zhiying-lin, line 269: compare the full structure instead?

pkg/controllers/workgenerator/controller.go
```diff
@@ -0,0 +1,346 @@
...
+		klog.ErrorS(misMatchErr, "Resource snapshot associated with the binding are not ready", "resourceBinding", klog.KObj(resourceBinding))
+		// TODO(RZ): return a specific error to allow creating works for the existing snapshots but don't delete the existing works
+		// make sure the reconcile requeue the request
+		return nil, controller.NewExpectedBehaviorError(misMatchErr)
+	}
+	klog.V(4).InfoS("Get all the resource snapshot associated with the binding", "numOfSnapshot", len(resourceSnapshots), "resourceBinding", klog.KObj(resourceBinding))
+	return resourceSnapshots, nil
+}
+
+// upsertWork creates or updates the work for the corresponding resource snapshot
```
zhiying-lin, line 238:
```suggestion
// upsertWork creates or updates the work for the corresponding resource snapshot.
```

pkg/controllers/workgenerator/controller_integration_test.go
ryanzhang-oss: I gave it a try but found out that I need to ignore a lot of fields if I compare the entire work such as
metadata.ResourceVersion, UID, Generation, CreationTimestamp, Annotations, ManagedFields.

Here is what the section says.

>  If your struct needs to be compared for approximate equality or some other kind of semantic equality, or it contains fields that cannot be compared for equality (e.g. if one of the fields is an io.Reader), tweaking a [cmp.Diff](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) or [cmp.Equal](https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal) comparison with [cmpopts](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts) options such as [cmpopts.IgnoreInterfaces](https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreInterfaces) may meet your needs ([example](https://go.dev/play/p/vrCUNVfxsvF)); otherwise, this technique just won't work, so do whatever works.

zhiying-lin: That's true, you have to ignore these runtime fields. The code will become,

```
                                wantWork := v1alpha1.Work{
						Name:            fmt.Sprintf(fleetv1beta1.FirstWorkNameFmt, testCRPName),
						Namespace:       namespaceName,
						OwnerReferences: expectedOwnerReference,
						Labels:          expectedLabels,
					},
					Spec: v1alpha1.WorkSpec{
						Workload: v1alpha1.WorkloadTemplate{Manifests: expectedManifest},
					},
				}
				ignoreOption := cmpopts.IgnoreFields(metav1.ObjectMeta{},
					"UID", "ResourceVersion", "ManagedFields", "CreationTimestamp", "Generation")
				diff = cmp.Diff(wantWork, work, ignoreOption)
				Expect(diff).Should(BeEmpty(), fmt.Sprintf("work(%s) mismatch (-want +got):\n%s", work.Name, diff))
```
which i think it has better readability. but will leave the decision to you.

zhiying-lin (approved):

pkg/controllers/workgenerator/controller_integration_test.go
ryanzhang-oss: I think I didn't get it right the first time.

PR #412, opened 2023-07-03
test: add integration test for clusterschedulingpolicysnapshot controller

zhiying-lin:

Add integration test

Fixes #

I have:

Ran integration tests


pkg/controllers/clustershedulingpolicysnapshot/suite_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	By("starting the controller manager")
+	klog.InitFlags(flag.CommandLine)
+	flag.Parse()
+
+	mgr, err = ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:             scheme.Scheme,
+		MetricsBindAddress: "0",
+		Logger:             klogr.NewWithOptions(klogr.WithFormat(klogr.FormatKlog)),
+		Port:               4848,
```
michaelawyu, line 79: This field has been deprecated I am afraid. For test we don't need this field any way I assume?

pkg/controllers/clustershedulingpolicysnapshot/suite_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}()
+})
+
+var _ = AfterSuite(func() {
+	defer klog.Flush()
+
+	cancel()
+	By("tearing down the test environment")
+	err := testEnv.Stop()
+	Expect(err).NotTo(HaveOccurred())
```
michaelawyu, line 104: A nit: should we use `Expect(...).To(Succeed())`?

test/utils/controller/controller.go
```diff
@@ -0,0 +1,48 @@
+/*
```
michaelawyu, line 1: Apologies for not getting the point, but is it kind of defeat the purpose of integration tests if we use this mock? Is there any reason why we cannot use the Controller with the API server for testing?

zhiying-lin: The clusterschedulingpolicysnapshot integration test is to test the real snapshot controller only. The util is to fake its dependency, crp controller.

we can add integration tests for the whole crp and its related watcher(dependency) in test/integration location. And we could use the real controller implementation.

michaelawyu: Oh, I mean, in this case, `NewController` with a no-op keyFunc/reconcileFunc would also work, right?

michaelawyu: For this one I don't have a strong preference though; a little bit more real would be better, but I guess it is true that the yield diminishes very quickly.

pkg/controllers/clustershedulingpolicysnapshot/controller_integration_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	Context("When creating new clusterSchedulingPolicySnapshot", func() {
+		It("Should ignore the event", func() {
+			By("By checking placement controller queue")
+			Consistently(func() bool {
+				return fakePlacementController.Key() == ""
+			}, duration, interval).Should(BeTrue(), "controller should ignore the create event and not enqueue the request into the placementController queue")
+
+			By("By deleting snapshot")
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: testSnapshotName}, createdSnapshot)).Should(Succeed())
```
michaelawyu, line 66: Hi Zhiying! A nit here: maybe it's not necessary to get before the delete here.

pkg/controllers/clustershedulingpolicysnapshot/controller_integration_test.go
```diff
@@ -0,0 +1,101 @@
...
+			Expect(k8sClient.Delete(ctx, createdSnapshot)).Should(Succeed())
```
michaelawyu, line 67: Also, to avoid flakiness, it might be a good idea to check that the object is gone before moving onto the next test (which will call create again).

michaelawyu: I understand that the next test is a little bit different, but would it be better if we move this cleanup to its own AfterEach block?

michaelawyu (approved):
A few minor comments; the rest LGTM.

pkg/controllers/clustershedulingpolicysnapshot/controller_integration_test.go
zhiying-lin: done

pkg/controllers/clustershedulingpolicysnapshot/controller_integration_test.go
```diff
@@ -0,0 +1,123 @@
...
+		createdSnapshot = &fleetv1beta1.ClusterSchedulingPolicySnapshot{}
+	)
+
+	BeforeEach(func() {
+		fakePlacementController.ResetQueue()
+		By("By creating a new clusterSchedulingPolicySnapshot")
+		snapshot := policySnapshot()
+		Expect(k8sClient.Create(ctx, snapshot)).Should(Succeed())
+	})
+
```
michaelawyu, line 58: Hi Zhiying! Sorry for the nitpicking, but I am wondering if something alone the lines:

```
BeforeEach(...)

AfterEach(func() {
			By("By deleting snapshot")
			createdSnapshot := policySnapshot()
			Expect(client.IgnoreNotFound(k8sClient.Delete(ctx, createdSnapshot))).Should(Succeed())

			By("By checking snapshot")
			Eventually(func() bool {
				return errors.IsNotFound(k8sClient.Get(ctx, types.NamespacedName{Name: testSnapshotName}, createdSnapshot))
			}, duration, interval).Should(BeTrue(), "snapshot should be deleted")
		)
```

This way all tests could use the same setup.

michaelawyu: This isn't particularly important in any way though.

michaelawyu (approved):
LGTM.

pkg/controllers/clustershedulingpolicysnapshot/controller_integration_test.go
zhiying-lin: i have a test to test the delete event, so that i did not put the delete func in the common afterEach.

PR #415, opened 2023-07-04
feat: enforce revisionHistoryLimit for scheduling policy snapshots

zhiying-lin:

respect the revisionHistoryLimit field before creating a new policy snapshot.

Fixes #

I have:

Added unit tests


michaelawyu:
A little bit running out of time today; will review it tomorrow. Apologies.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -116,7 +117,7 @@ func (r *Reconciler) getOrCreateClusterPolicySnapshot(ctx context.Context, crp *
 	}
 
 	// latestPolicySnapshotIndex should be -1 when there is no snapshot.
-	latestPolicySnapshot, latestPolicySnapshotIndex, err := r.lookupLatestClusterPolicySnapshot(ctx, crp)
+	latestPolicySnapshot, latestPolicySnapshotIndex, sortedList, err := r.lookupLatestClusterPolicySnapshot(ctx, crp)
```
michaelawyu, line 120: Considering that the cleanup is not a very frequent op, would it be better to do it when the size limit is actually over rather than doing it for each list?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -179,6 +186,37 @@ func (r *Reconciler) getOrCreateClusterPolicySnapshot(ctx context.Context, crp *
...
+		if err != nil {
+			return err
+		}
+	}
+	crpKObj := klog.KObj(crp)
+	// respect the revisionHistoryLimit field
+	revisionLimit := fleetv1beta1.RevisionHistoryLimitDefaultValue
+	if crp.Spec.RevisionHistoryLimit != nil {
+		revisionLimit = *crp.Spec.RevisionHistoryLimit
+		if revisionLimit <= 0 {
```
michaelawyu, line 202: This check should be done on the schema end; and it would be better to use the default value if the setting is invalid rather than throwing an error here IMO.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -179,6 +186,37 @@ func (r *Reconciler) getOrCreateClusterPolicySnapshot(ctx context.Context, crp *
 	return latestPolicySnapshot, nil
 }
 
+func (r *Reconciler) deleteRedundantSchedulingPolicySnapshots(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement, sortedList *fleetv1beta1.ClusterSchedulingPolicySnapshotList) error {
+	var err error
+	if sortedList == nil {
```
michaelawyu, line 191: This will happen every time if the function earlier does not return a sortedList?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -324,47 +363,67 @@ func (r *Reconciler) lookupLatestClusterPolicySnapshot(ctx context.Context, crp
...
+		// CRP controller needs a scheduling policy snapshot watcher to enqueue the CRP request.
+		// So the snapshots should be read from cache.
+		return nil, controller.NewAPIServerError(true, err)
+	}
+	sort.Slice(snapshotList.Items, func(i, j int) bool {
+		ii, err := parsePolicyIndexFromLabel(&snapshotList.Items[i])
 		if err != nil {
-			klog.ErrorS(err, "Failed to parse the policy index label", "clusterPolicySnapshot", klog.KObj(&snapshotList.Items[i]))
-			return nil, -1, controller.NewUnexpectedBehaviorError(err)
+			klog.ErrorS(controller.NewUnexpectedBehaviorError(err), "Failed to parse the policy index label", "clusterSchedulingPolicySnapshot", klog.KObj(&snapshotList.Items[i]))
```
michaelawyu, line 415: If an error occurs here, the sort result is unclear. I am not sure if simply logging the error and leave the default value for comparison is the best behavior here. It's better to have a check beforehand.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -324,47 +363,67 @@ func (r *Reconciler) lookupLatestClusterPolicySnapshot(ctx context.Context, crp
...
 		}
-		if lastPolicyIndex < policyIndex {
-			index = i
-			lastPolicyIndex = policyIndex
+		ji, err := parsePolicyIndexFromLabel(&snapshotList.Items[j])
+		if err != nil {
```
michaelawyu, line 418: Same as above.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: The list will only happen when it needs to create an new snapshot. Removed this return to clear the confusion.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: done, returned the unexpected error instead of error swallowing.

michaelawyu (approved):
A nit comment, others LGTM.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -324,12 +362,14 @@ func (r *Reconciler) lookupLatestClusterPolicySnapshot(ctx context.Context, crp
...
-		return nil, -1, controller.NewAPIServerError(false, err)
+		// CRP controller needs a scheduling policy snapshot watcher to enqueue the CRP request.
+		// So the snapshots should be read from cache.
+		return nil, -1, controller.NewAPIServerError(true, err)
 	}
 	if len(snapshotList.Items) == 1 {
 		policyIndex, err := parsePolicyIndexFromLabel(&snapshotList.Items[0])
 		if err != nil {
-			klog.ErrorS(err, "Failed to parse the policy index label", "clusterPolicySnapshot", klog.KObj(&snapshotList.Items[0]))
+			klog.ErrorS(err, "Failed to parse the policy index label", "clusterResourcePlacement", crpKObj, "clusterPolicySnapshot", klog.KObj(&snapshotList.Items[0]))
```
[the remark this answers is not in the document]
zhiying-lin: will merge it first and address it in my next PR, thanks

PR #417, opened 2023-07-05
interface: update crp status api and add validation markers

zhiying-lin:

-  Remove the duplicate clusterName from FailedResourcePlacement as it's already in its parent, ResourcePlacementStatus.
- Add validation markers

Fixes #

I have:

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -218,6 +218,7 @@ type TopologySpreadConstraint struct {
 	// to topologies that satisfy it.
 	// It's an optional field. Default value is 1 and 0 is not allowed.
 	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
 	// +optional
 	MaxSkew *int32 `json:"maxSkew,omitempty"`
 
@@ -365,6 +366,7 @@ type ResourceIdentifier struct {
 // ResourcePlacementStatus represents the placement status of selected resources for one target cluster.
 type ResourcePlacementStatus struct {
 	// ClusterName is the name of the cluster this resource is assigned to.
+	// +required
 	ClusterName string `json:"clusterName"`
 
 	// +kubebuilder:validation:MaxItems=100
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -79,7 +79,7 @@ type ClusterResourcePlacementSpec struct {
 	// This is a pointer to distinguish between explicit zero and not specified.
 	// Defaults to 10.
 	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Maximum=20
+	// +kubebuilder:validation:Maximum=1000
 	// +kubebuilder:default=10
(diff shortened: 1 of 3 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -209,6 +217,7 @@ type TopologySpreadConstraint struct {
 	// When `whenUnsatisfiable=ScheduleAnyway`, it is used to give higher precedence
 	// to topologies that satisfy it.
 	// It's an optional field. Default value is 1 and 0 is not allowed.
+	// +kubebuilder:default=1
```
michaelawyu, line 220: If zero is not allowed there should be a minimum check right?

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -347,8 +365,6 @@ type ResourceIdentifier struct {
 // ResourcePlacementStatus represents the placement status of selected resources for one target cluster.
 type ResourcePlacementStatus struct {
 	// ClusterName is the name of the cluster this resource is assigned to.
-	// If it is not empty, its value should be unique cross all placement decisions for the Placement.
-	// +optional
 	ClusterName string `json:"clusterName"`
```
michaelawyu, line 368: This should be a required field?

config/crd/bases/placement.azure.com_clusterschedulingpolicysnapshots.yaml
```diff
@@ -144,6 +144,8 @@ spec:
                                     corresponding clusterSelectorTerm, in the range
                                     [-100, 100].
                                   format: int32
+                                  maximum: 100
```
ryanzhang-oss, line 147: why is there changes in the snapshot CRD but I don't see change in its definition?

zhiying-lin: i added the validation marker for the PreferredClusterSelector and it's part of PlacementPolicy in the spec of schedulingPolicySnapshot.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -78,6 +78,9 @@ type ClusterResourcePlacementSpec struct {
 	// The number of old ClusterSchedulingPolicySnapshot or ClusterResourceSnapshot resources to retain to allow rollback.
 	// This is a pointer to distinguish between explicit zero and not specified.
 	// Defaults to 10.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=20
```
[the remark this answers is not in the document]
zhiying-lin: I searched the k8 repo and they assume the max limit is math.MaxInt32 to keep all the revisions.

I cannot think of the reason that customers would like to keep the large number of revisions and 20 should be a reasonable number to start with. It would be easier that we increase the limit if needed compared with adding the limit.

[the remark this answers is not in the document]
ryanzhang-oss: 20 seems to be a bit low. Some clients may want to keep a long history. why not 1000?

ryanzhang-oss (approved):

PR #419, opened 2023-07-05
feat: use patches in scheduler to avoid races + add some more unit tests

michaelawyu:

This PR updates the scheduler logic to use patches (instead of updates) to avoid races.

It also adds a number of unit tests that wasn't included in previous scheduler PRs.

I have:

- [x] Unit tests

### DIFF
--- a/pkg/scheduler/framework/framework_test.go
+++ b/pkg/scheduler/framework/framework_test.go
@@ -18,6 +18,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/framework/frameworkutils.go
```diff
@@ -136,10 +150,13 @@ func crossReferencePickedCustersAndObsoleteBindings(
...
 		// Update the binding so that it is associated with the lastest scheduling policy.
-		binding.Spec.SchedulingPolicySnapshotName = policy.Name
+		updated.Spec.SchedulingPolicySnapshotName = policy.Name
 
 		// Add the binding to the toUpdate list.
-		toUpdate = append(toUpdate, binding)
+		toPatch = append(toPatch, &bindingWithPatch{
+			updated: updated,
+			patch:   patch,
+		})
```
zhiying-lin, line 159:
```suggestion
		toPatch = append(toPatch, &bindingWithPatch{
			updated: updated,
			patch:   client.MergeFrom(binding),
		})
```
nit, no need to use the variable or move the variable closer to where it's used

pkg/scheduler/framework/framework_test.go
```diff
@@ -1393,3 +1448,205 @@ func TestCrossReferencePickedCustersAndObsoleteBindings(t *testing.T) {
...
+	}
+
+	// Check if the requested binding has been deleted.
+	deletedBinding := &fleetv1beta1.ClusterResourceBinding{}
+	if err := fakeClient.Get(ctx, types.NamespacedName{Name: anotherBindingName}, deletedBinding); err != nil {
+		t.Errorf("Get() binding %s = %v, want no error", anotherBindingName, err)
+	}
+	if deletedBinding.Spec.State != fleetv1beta1.BindingStateUnscheduled {
+		t.Errorf("binding %s state = %s, want unscheduled", anotherBindingName, deletedBinding.Spec.State)
+	}
```
zhiying-lin, line 1651: compare the full structure to test we only update the state and other fields should not change?

michaelawyu: Done.

zhiying-lin (approved):
LGTM :)

PR #420, opened 2023-07-07
test: fix the crp watcher integration test flakiness

zhiying-lin:

make sure the event has arrived before resetting the queue to fix the test flakiness

Fixes #

I have:

run the unit test ci

### DIFF
--- a/pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
+++ b/pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
@@ -82,6 +82,13 @@ var _ = Describe("Test ClusterResourcePlacement Watcher", func() {
 
 	Context("When updating clusterResourcePlacement", func() {
 		BeforeEach(func() {
+			By("By checking the placement queue before resetting")
+			// The event could arrive after the resetting, which causes the flakiness.
+			// It makes sure the queue is clear before proceed.
+			Eventually(func() bool {
+				return fakePlacementController.Key() == testCRPName
+			}, eventuallyTimeout, interval).Should(BeTrue(), "placementController should receive the CRP name")
+
 			By("By resetting the placement queue")
 			fakePlacementController.ResetQueue()
 		})
@@ -132,6 +139,13 @@ var _ = Describe("Test ClusterResourcePlacement Watcher", func() {
 
 	Context("When deleting clusterResourcePlacement", func() {
 		BeforeEach(func() {
+			By("By checking the placement queue before resetting")
+			// The event could arrive after the resetting, which causes the flakiness.
+			// It makes sure the queue is clear before proceed.
+			Eventually(func() bool {
+				return fakePlacementController.Key() == testCRPName
+			}, eventuallyTimeout, interval).Should(BeTrue(), "placementController should receive the CRP name")
+
 			By("By resetting the placement queue")
 			fakePlacementController.ResetQueue()
 		})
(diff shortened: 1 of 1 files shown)


pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
```diff
@@ -82,6 +82,13 @@ var _ = Describe("Test ClusterResourcePlacement Watcher", func() {
 
 	Context("When updating clusterResourcePlacement", func() {
 		BeforeEach(func() {
+			By("By checking the placement queue before resetting")
+			// The event could arrive after the resetting, which causes the flakiness.
+			// It makes sure the queue is clear before proceed.
+			Eventually(func() bool {
+				return fakePlacementController.Key() == testCRPName
+			}, eventuallyTimeout, interval).Should(BeTrue(), "placementController should receive the CRP name")
```
ryanzhang-oss, line 90: I don't fully get it. If we want to get a clean start for all the test, why create the CPR in the first place at line 60 and then reset it before each test?

ryanzhang-oss: one can move the create to the test in line 75

ryanzhang-oss: and here

ryanzhang-oss: or move the test in line 75 to the BeforeEach for all.

zhiying-lin: moved the create test into the BeforeEach.

michaelawyu (approved):
LGTM.

pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
```diff
@@ -58,33 +58,19 @@ var _ = Describe("Test ClusterResourcePlacement Watcher", func() {
...
 	})
 
 	Context("When updating clusterResourcePlacement", func() {
-		BeforeEach(func() {
-			By("By resetting the placement queue")
-			fakePlacementController.ResetQueue()
-		})
 		AfterEach(func() {
 			By("By deleting crp")
 			createdCRP = clusterResourcePlacementForTest()
```
ryanzhang-oss, line 76: why creating this again?

pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
```diff
@@ -58,33 +58,19 @@ var _ = Describe("Test ClusterResourcePlacement Watcher", func() {
 		By("By creating a new clusterResourcePlacement")
 		crp := clusterResourcePlacementForTest()
```
ryanzhang-oss, line 59: why not assign this to createdCRP so we don't need to create it again when we delete it?

ryanzhang-oss (approved):

pkg/controllers/clusterresourceplacementwatcher/watcher_integration_test.go
zhiying-lin: oh, it's not creating a new CRP by sending to the API server and it's building a CRP structure. the name is not good. Will rename this variable.

ryanzhang-oss (approved):

PR #421, opened 2023-07-07
feat: add the subindex annotation

zhiying-lin:

- Add this annotation to help CRP to sort the snapshots in the group and work generator to create the work name.
- Fix the log to use the final policy snapshot name.

Now the first snapshot name will always be {CRPName}-{resourceIndex}-snapshot.
It helps to solve the duplication of resourceSnapshotName and now the name format is fixed and there is no ambiguity. 

This approach is simpler compared with other options' complexity otherwise the CRP name cannot contain "-" and not supporting it will hurt customer's experience.

Fixes #

I have:

### DIFF
--- a/apis/placement/v1beta1/resourcesnapshot_types.go
+++ b/apis/placement/v1beta1/resourcesnapshot_types.go
@@ -22,7 +22,7 @@ const (
 	// NumberOfResourceSnapshotsAnnotation is the annotation that contains the total number of resource snapshots.
 	NumberOfResourceSnapshotsAnnotation = fleetPrefix + "numberOfResourceSnapshots"
 
-	// SubindexOfResourceSnapshotAnnotation is the annotation to store the index of resource snapshot in the group.
+	// SubindexOfResourceSnapshotAnnotation is the annotation to store the subindex of resource snapshot in the group.
 	SubindexOfResourceSnapshotAnnotation = fleetPrefix + "subindexOfResourceSnapshot"
 
 	// ResourceSnapshotNameFmt is resourcePolicySnapshot name format: {CRPName}-{resourceIndex}.
--- a/apis/placement/v1beta1/resourcesnapshot_types.go
+++ b/apis/placement/v1beta1/resourcesnapshot_types.go
@@ -27,6 +27,9 @@ const (
 
 	// ResourceSnapshotNameFmt is resourcePolicySnapshot name format: {CRPName}-{resourceIndex}-snapshot.
 	ResourceSnapshotNameFmt = "%s-%d-snapshot"
+
+	// ResourceSnapshotNameWithSubindexFmt is resourcePolicySnapshot name with subindex format: {CRPName}-{resourceIndex}-{subindex}.
+	ResourceSnapshotNameWithSubindexFmt = " %s-%d-%d"
 )
 
 // +genclient
@@ -42,7 +45,7 @@ const (
 // Its spec is immutable.
 // We may need to produce more than one resourceSnapshot for all the resources a ResourcePlacement selected to get around the 1MB size limit of k8s objects.
 // We assign an ever-increasing index for each such group of resourceSnapshots.
-// The name convention of a clusterResourceSnapshot is {CRPName}-{resourceIndex}-{snapshot|-{subindex}}
+// The naming convention of a clusterResourceSnapshot is {CRPName}-{resourceIndex}-{subindex}
 // where the name of the first snapshot of a group has no subindex part so its name is {CRPName}-{resourceIndex}-snapshot.
 // resourceIndex will begin with 0.
 // Each snapshot MUST have the following labels:
--- a/apis/placement/v1beta1/resourcesnapshot_types.go
+++ b/apis/placement/v1beta1/resourcesnapshot_types.go
@@ -25,7 +25,7 @@ const (
 	// SubindexOfResourceSnapshotAnnotation is the annotation to store the subindex of resource snapshot in the group.
 	SubindexOfResourceSnapshotAnnotation = fleetPrefix + "subindexOfResourceSnapshot"
 
-	// ResourceSnapshotNameFmt is resourcePolicySnapshot name format: {CRPName}-{resourceIndex}.
+	// ResourceSnapshotNameFmt is resourcePolicySnapshot name format: {CRPName}-{resourceIndex}-snapshot.
 	ResourceSnapshotNameFmt = "%s-%d-snapshot"
 )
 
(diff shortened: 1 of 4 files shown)


apis/placement/v1beta1/resourcesnapshot_types.go
```diff
@@ -22,6 +22,9 @@ const (
 	// NumberOfResourceSnapshotsAnnotation is the annotation that contains the total number of resource snapshots.
 	NumberOfResourceSnapshotsAnnotation = fleetPrefix + "numberOfResourceSnapshots"
 
+	// SubindexOfResourceSnapshotAnnotation is the annotation to store the index of resource snapshot in the group.
```
ryanzhang-oss, line 25:
```suggestion
	// SubindexOfResourceSnapshotAnnotation is the annotation to store the sub-index of resource snapshot in the group.
```

apis/placement/v1beta1/resourcesnapshot_types.go
```diff
@@ -39,9 +42,8 @@ const (
 // Its spec is immutable.
 // We may need to produce more than one resourceSnapshot for all the resources a ResourcePlacement selected to get around the 1MB size limit of k8s objects.
 // We assign an ever-increasing index for each such group of resourceSnapshots.
-// The name convention of a clusterResourceSnapshot is {CRPName}-{resourceIndex}(-{subindex})*
-// where the name of the first snapshot of a group has no subindex part so its name is {CRPName}-{resourceIndex}.
-// resourceIndex will begin with 0.
+// The name convention of a clusterResourceSnapshot is {CRPName}-{resourceIndex}-{subindex} where resourceIndex and
+// subindex will begin with 0.
```
ryanzhang-oss, line 46: I guess this is to avoid conflict between different resource snapshot's name? Obviously this is not ideal. What are our other options?

ryanzhang-oss: I get that restricting CRP name to not having "-" is not ideal either.

ryanzhang-oss:
There is a UT failure that seems to be interesting

apis/placement/v1beta1/resourcesnapshot_types.go
zhiying-lin: The purpose of renaming the first snapshot is to keep snapshot names consistent no matter it's a sub-snapshot or not as mentioned in the "special note" of the PR description in order to avoid special handling in some cases.

To solve the duplication of snapshots (policySnapshot/resourceSnapshot), i'm thinking when we create a snapshot, we create a CRP + {random-suffix} as the a new CRP name and store it in each snapshot. The policySnapshot name becomes {CRPNameWithRandomSuffix}- {index} and resourceSnapshot name becomes {CRPNameWithRandomSuffix}-{resourceIndex}-{subIndex}. The CRPNameWithRandomSuffix should be unchanged during the CRP lifecycle.

WDYT?

ryanzhang-oss (dismissed):

michaelawyu:
The actual name in use for resource snapshots is not updated; is this expected?

apis/placement/v1beta1/resourcesnapshot_types.go
```diff
@@ -22,6 +22,9 @@ const (
 	// NumberOfResourceSnapshotsAnnotation is the annotation that contains the total number of resource snapshots.
 	NumberOfResourceSnapshotsAnnotation = fleetPrefix + "numberOfResourceSnapshots"
 
+	// SubindexOfResourceSnapshotAnnotation is the annotation to store the subindex of resource snapshot in the group.
+	SubindexOfResourceSnapshotAnnotation = fleetPrefix + "subindexOfResourceSnapshot"
+
 	// ResourceSnapshotNameFmt is resourcePolicySnapshot name format: {CRPName}-{resourceIndex}.
```
ryanzhang-oss, line 28: if this PR determines the  naming convention then should we change this too?

apis/placement/v1beta1/resourcesnapshot_types.go
```diff
@@ -22,8 +22,11 @@ const (
...
```
ryanzhang-oss, line 28:
```suggestion
	// ResourceSnapshotNameFmt is resourcePolicySnapshot name format: {CRPName}-{resourceIndex}-snapshot
```

ryanzhang-oss (dismissed):

apis/placement/v1beta1/resourcesnapshot_types.go
```diff
@@ -39,8 +42,8 @@ const (
 // Its spec is immutable.
 // We may need to produce more than one resourceSnapshot for all the resources a ResourcePlacement selected to get around the 1MB size limit of k8s objects.
 // We assign an ever-increasing index for each such group of resourceSnapshots.
-// The name convention of a clusterResourceSnapshot is {CRPName}-{resourceIndex}(-{subindex})*
-// where the name of the first snapshot of a group has no subindex part so its name is {CRPName}-{resourceIndex}.
+// The name convention of a clusterResourceSnapshot is {CRPName}-{resourceIndex}-{snapshot|-{subindex}}
```
michaelawyu, line 45: Nit, naming convention?

michaelawyu: Also a nit: since the special case has been explained, here we can just use {CRPName}-{resourceIndex}-{subindex} to avoid confusion.

apis/placement/v1beta1/resourcesnapshot_types.go
```diff
@@ -22,8 +22,11 @@ const (
 	// NumberOfResourceSnapshotsAnnotation is the annotation that contains the total number of resource snapshots.
 	NumberOfResourceSnapshotsAnnotation = fleetPrefix + "numberOfResourceSnapshots"
 
-	// ResourceSnapshotNameFmt is resourcePolicySnapshot name format: {CRPName}-{resourceIndex}.
-	ResourceSnapshotNameFmt = "%s-%d"
+	// SubindexOfResourceSnapshotAnnotation is the annotation to store the subindex of resource snapshot in the group.
+	SubindexOfResourceSnapshotAnnotation = fleetPrefix + "subindexOfResourceSnapshot"
+
+	// ResourceSnapshotNameFmt is resourcePolicySnapshot name format: {CRPName}-{resourceIndex}-snapshot.
```
michaelawyu, line 28: Maybe add a comment here to explain that this is for temporary usage?

zhiying-lin: it's not temporary. Added another format to address your concern.

michaelawyu (approved):
LGTM ;)

PR #426, opened 2023-07-11
feat: enforce revisionHistoryLimit for resource snapshots

zhiying-lin:

- delete redundant resourceSnapshots before creating one
- add the NumberOfResourceSnapshots annotation

Fixes #

I have:

UT


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -463,29 +523,76 @@ func (r *Reconciler) lookupLatestResourceSnapshot(ctx context.Context, crp *flee
...
+		if err != nil {
+			klog.ErrorS(err, "Failed to parse the subindex index", "clusterResourcePlacement", crpKObj, "clusterResourceSnapshot", klog.KObj(&snapshotList.Items[j]))
+			errs = append(errs, err)
+		}
+		jDoesExist, jSubindex, err := utils.ExtractSubindexFromClusterResourceSnapshot(&snapshotList.Items[j])
+		if err != nil {
+			klog.ErrorS(err, "Failed to parse the subindex index", "clusterResourcePlacement", crpKObj, "clusterResourceSnapshot", klog.KObj(&snapshotList.Items[j]))
+			errs = append(errs, err)
+		}
+		if !iDoesExist { // check if it's the first snapshot
```
michaelawyu, line 582: This lessFunc already compares the major index at an earlier step; so any two items can reach this point must have the same major index. If both of them do not have subindexes, it should be an error.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -192,25 +201,69 @@ func (r *Reconciler) deleteRedundantSchedulingPolicySnapshots(ctx context.Contex
...
+
+// deleteRedundantResourceSnapshots handles multiple snapshots in a group.
+func (r *Reconciler) deleteRedundantResourceSnapshots(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement) error {
+	sortedList, err := r.listSortedResourceSnapshots(ctx, crp)
+	if err != nil {
+		return err
+	}
+
+	// respect the revisionHistoryLimit field
+	revisionLimit := parseRevisionLimit(crp)
```
michaelawyu, line 240: This limit is parsed multiple times in the code; since both snapshot types use the same limit, it might be better to extract it out to a common place.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -192,25 +201,69 @@ func (r *Reconciler) deleteRedundantSchedulingPolicySnapshots(ctx context.Contex
...
+	if len(sortedList.Items) < revisionLimit {
```
michaelawyu, line 241: The semantics does not match with the deletion logic down there.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -192,25 +201,69 @@ func (r *Reconciler) deleteRedundantSchedulingPolicySnapshots(ctx context.Contex
...
 	}
-	for i := 0; i <= len(sortedList.Items)-int(revisionLimit); i++ {
+
+	crpKObj := klog.KObj(crp)
+	lastGroupIndex := -1
+	groupCounter := 0
+
+	// delete the snapshots from the end as there are could be multiple snapshots in a group in order to keep the latest
+	// snapshots from the end.
+	for i := len(sortedList.Items) - 1; i >= 0; i-- {
```
michaelawyu, line 251: Here it deletes snapshots by group, but the check above counts all snapshots as individuals?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -192,25 +201,69 @@ func (r *Reconciler) deleteRedundantSchedulingPolicySnapshots(ctx context.Contex
...
+		ii, err := parseResourceIndexFromLabel(&sortedList.Items[i])
+		if err != nil {
+			klog.ErrorS(err, "Failed to parse the resource index label", "clusterResourcePlacement", crpKObj, "clusterResourceSnapshot", snapshotKObj)
+			return controller.NewUnexpectedBehaviorError(err)
+		}
+		if ii != lastGroupIndex {
+			groupCounter++
+			lastGroupIndex = ii
+		}
+		if groupCounter < revisionLimit { // need to reserve one slot for the new snapshot
```
michaelawyu, line 262: If the group count limit is already found, and the array is sorted, any further check does not need to performed any more, if I understand it correctly?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -272,6 +330,8 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
 			},
 			Annotations: map[string]string{
 				fleetv1beta1.ResourceGroupHashAnnotation: resourceHash,
+				// TODO need to updated once we support multiple snapshots
```
ryanzhang-oss, line 333: looks like the PR assumes more than one resourceSnapshot in an index but the create part only handles one?

ryanzhang-oss:
Can we delete a snapshot that still has bindings pointing to it?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -187,38 +207,79 @@ func (r *Reconciler) getOrCreateClusterSchedulingPolicySnapshot(ctx context.Cont
...
-	// respect the revisionHistoryLimit field
-	revisionLimit := fleetv1beta1.RevisionHistoryLimitDefaultValue
-	if crp.Spec.RevisionHistoryLimit != nil {
-		revisionLimit = *crp.Spec.RevisionHistoryLimit
-		if revisionLimit <= 0 {
-			err := fmt.Errorf("invalid clusterResourcePlacement %s: invalid revisionHistoryLimit %d", crpKObj, revisionLimit)
-			klog.ErrorS(controller.NewExpectedBehaviorError(err), "Invalid revisionHistoryLimit value and using default value instead", "clusterResourcePlacement", crpKObj)
-			// use the default value instead
-			revisionLimit = fleetv1beta1.RevisionHistoryLimitDefaultValue
+	if len(sortedList.Items)-revisionHistoryLimit > 0 {
```
michaelawyu, line 219: Hi Zhiying! The earlier check has deemed that the `revisionHistoryLimit` has been exceeded; this if block is not necessary I think.

michaelawyu: And we still need to perform cleanup when len(sorted) == revisionHistoryLimit if I understand it correctly.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -187,38 +207,79 @@ func (r *Reconciler) getOrCreateClusterSchedulingPolicySnapshot(ctx context.Cont
...
+			return controller.NewUnexpectedBehaviorError(err)
+		}
+		if ii != lastGroupIndex {
+			groupCounter++
+			lastGroupIndex = ii
+		}
+		if groupCounter < revisionHistoryLimit { // need to reserve one slot for the new snapshot
+			continue
+		}
 		if err := r.Client.Delete(ctx, &sortedList.Items[i]); err != nil && !errors.IsNotFound(err) {
```
michaelawyu, line 268: Hi Zhiying! Would you mind adding a comment here explaining that if the number of groups is less than revisionHistoryLimit, no groups will be deleted?

michaelawyu: I just feel that it would make things a little bit more clear if we explicitly mention that the history limit is enforced on groups.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -187,38 +207,79 @@ func (r *Reconciler) getOrCreateClusterSchedulingPolicySnapshot(ctx context.Cont
...
-			klog.ErrorS(err, "Failed to delete clusterSchedulingPolicySnapshot", "clusterResourcePlacement", crpKObj, "clusterSchedulingPolicySnapshot", klog.KObj(&sortedList.Items[i]))
+			klog.ErrorS(err, "Failed to delete clusterResourceSnapshot", "clusterResourcePlacement", crpKObj, "clusterResourceSnapshot", snapshotKObj)
 			return controller.NewAPIServerError(false, err)
 		}
 	}
+	if groupCounter-revisionHistoryLimit > 0 {
+		// We always delete before creating a new snapshot, the snapshot group size should never exceed the limit
+		// as there is no finalizer added and the object should be deleted immediately.
+		klog.Warningf("The number of clusterResourceSnapshot groups exceeds the revisionHistoryLimit and it should never happen", "clusterResourcePlacement", klog.KObj(crp), "numberOfSnapshotGroup", groupCounter, "revisionHistoryLimit", revisionHistoryLimit)
```
michaelawyu, line 276: A nit: `numberOfSnapshotGroups`. Sorry for the nitpicking.

michaelawyu:
Other than the comments, everything LGTM ;)

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: if len(sorted) == limit, that's the expected situation. So we don't need this warning but still need to delete one snapshot.

Added a comment in the code.

michaelawyu: My concern is, is this really a warning? And the equality is not an expected situation. The limit is just for cleaning thing up.

michaelawyu: But anyway, this is a minor point.

michaelawyu (approved):
LGTM.

PR #428, opened 2023-07-12
feat: add generation related fields to policy snapshot status

michaelawyu:

This PR adds generation related fields to the policy snapshot status API. This helps resolve some inconsistencies regarding status updates.

I have:

N/A

N/A

### DIFF
--- a/apis/placement/v1beta1/policysnapshot_types.go
+++ b/apis/placement/v1beta1/policysnapshot_types.go
@@ -19,10 +19,6 @@ const (
 
 	// NumberOfClustersAnnotation is the annotation that indicates how many clusters should be selected for selectN placement type.
 	NumberOfClustersAnnotation = fleetPrefix + "numberOfClusters"
-
-	// CRPGenerationAnnotation is the annotation that indicates the generation from which the
-	// policy snapshot is created or last updated.
-	CRPGenerationAnnotation = fleetPrefix + "CRPGeneration"
 )
 
 // +genclient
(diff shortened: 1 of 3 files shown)


apis/placement/v1beta1/policysnapshot_types.go
```diff
@@ -19,6 +19,10 @@ const (
 
 	// NumberOfClustersAnnotation is the annotation that indicates how many clusters should be selected for selectN placement type.
 	NumberOfClustersAnnotation = fleetPrefix + "numberOfClusters"
+
+	// CRPGenerationAnnotation is the annotation that indicates the generation from which the
+	// policy snapshot is created or last updated.
+	CRPGenerationAnnotation = fleetPrefix + "CRPGeneration"
```
zhiying-lin, line 25: just come to my mind, does resourceSnapshot & work also needs this annotation?

zhiying-lin (approved):

apis/placement/v1beta1/policysnapshot_types.go
zhiying-lin: @ryanzhang-oss do we want to track the crp generation for work? if so, we need another annotation on the work, something like, statusObservedCRPGeneration.

@michaelawyu raised a good point and there is a case, when resource itself is changed, crp will create a new resourceSnapshot but crp generation won't change. The work status may not be accurate to reflect the latest change in some cases.

in this situation, do we still want to track the CRP generation for the resources?

zhiying-lin: yeah, that's true. discussed with Ryan offline and resourceSnapshot creation is not only triggered by the CRP changes. So we will not track the CRP generation here.

PR #429, opened 2023-07-12
feat: build crp status based on policySnapshot & resourceSnapshot

zhiying-lin:

build the CRP status

Fixes #

I have:

UT

The PR is only to build the status and updating CRP will be in a separate PR.
appliedCondition for both CRP status & resourcePlacementStatus by collecting work status will be in a separate PR.

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -439,11 +439,11 @@ const (
 	// - "Unknown" means we haven't started the apply yet.
 	ResourcesAppliedConditionType ResourcePlacementConditionType = "ResourceApplied"
 
-	// PlacementScheduledConditionType indicates whether we have successfully scheduled the placement.
+	// ResourceScheduledConditionType indicates whether we have successfully scheduled the selected resources.
 	// Its condition status can be one of the following:
-	// - "True" means we have successfully scheduled the resources to fully satisfy the placement requirement.
+	// - "True" means we have successfully scheduled the resources to satisfy the placement requirement.
 	// - "False" means we didn't fully satisfy the placement requirement. We will fill the Message field.
-	PlacementScheduledConditionType ResourcePlacementConditionType = "PlacementScheduled"
+	ResourceScheduledConditionType ResourcePlacementConditionType = "ResourceScheduled"
 )
 
 // PlacementType identifies the type of placement.
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -438,6 +438,12 @@ const (
 	// - "False" means some of them have failed.
 	// - "Unknown" means we haven't started the apply yet.
 	ResourcesAppliedConditionType ResourcePlacementConditionType = "ResourceApplied"
+
+	// PlacementScheduledConditionType indicates whether we have successfully scheduled the placement.
+	// Its condition status can be one of the following:
+	// - "True" means we have successfully scheduled the resources to fully satisfy the placement requirement.
+	// - "False" means we didn't fully satisfy the placement requirement. We will fill the Message field.
+	PlacementScheduledConditionType ResourcePlacementConditionType = "PlacementScheduled"
 )
 
 // PlacementType identifies the type of placement.
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -685,6 +685,7 @@ func (r *Reconciler) buildPlacementStatus(crp *fleetv1beta1.ClusterResourcePlace
 		Conditions:        []metav1.Condition{},
 	}
 	// TODO query the sub resourceSnapshots by using latestResourceSnapshot and then update the status.
+	// Work generator also needs to do this and we need to refactor the code to share the same functionality.
 	for i := range latestResourceSnapshot.Spec.SelectedResources {
 		obj, err := r.decodeResourceContent(&latestResourceSnapshot.Spec.SelectedResources[i])
 		if err != nil {
@@ -703,6 +704,11 @@ func (r *Reconciler) buildPlacementStatus(crp *fleetv1beta1.ClusterResourcePlace
 	scheduledCondition := buildScheduledCondition(crp, latestSchedulingPolicySnapshot)
 	meta.SetStatusCondition(&status.Conditions, scheduledCondition)
 
+	// When scheduledCondition is unknown, appliedCondition should be unknown too.
+	// Note: If the scheduledCondition is failed, it means the placement requirement cannot be satisfied fully. For example,
+	// pickN deployment requires 5 clusters and scheduler schedules the resources on 3 clusters. And the appliedCondition
+	// could be true when resources are applied successfully on these 3 clusters and the detailed the resourcePlacementStatuses
+	// need to be populated.
 	if scheduledCondition.Status == metav1.ConditionUnknown {
 		appliedCondition := metav1.Condition{
 			Status:             metav1.ConditionUnknown,
@@ -716,7 +722,7 @@ func (r *Reconciler) buildPlacementStatus(crp *fleetv1beta1.ClusterResourcePlace
 		return status, nil
 	}
 
-	// TODO collect work status and build appliedCondition & resourcePlacementStatus
+	// TODO check the resourceSnapshot status (set by work generator) and build appliedCondition & resourcePlacementStatus
 	status.PlacementStatuses = buildResourcePlacementStatus(crp, latestSchedulingPolicySnapshot)
 	return status, nil
 }
@@ -750,22 +756,15 @@ func buildScheduledCondition(crp *fleetv1beta1.ClusterResourcePlacement, latestS
 
 func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot) []fleetv1beta1.ResourcePlacementStatus {
 	res := make([]fleetv1beta1.ResourcePlacementStatus, 0, len(latestSchedulingPolicySnapshot.Status.ClusterDecisions))
-	for i, c := range latestSchedulingPolicySnapshot.Status.ClusterDecisions {
-		scheduledCondition := metav1.Condition{
-			Status:             metav1.ConditionTrue,
-			Type:               string(fleetv1beta1.PlacementScheduledConditionType),
-			Reason:             "ScheduleSucceeded",
-			Message:            fmt.Sprintf(resourcePlacementConditionScheduleSucceededMessageFormat, c.ClusterName, c.Reason),
-			ObservedGeneration: crp.Generation,
-		}
+	for _, c := range latestSchedulingPolicySnapshot.Status.ClusterDecisions {
+		var scheduledCondition metav1.Condition
 		var rp fleetv1beta1.ResourcePlacementStatus
-		if !latestSchedulingPolicySnapshot.Status.ClusterDecisions[i].Selected {
+		if !c.Selected {
 			// TODO: we could improve the message by summarizing the failure reasons from all of the unselected clusters.
 			// For now, it starts from adding some sample failures of unselected clusters.
-			scheduledCondition.Status = metav1.ConditionFalse
 			scheduledCondition = metav1.Condition{
 				Status:             metav1.ConditionFalse,
-				Type:               string(fleetv1beta1.PlacementScheduledConditionType),
+				Type:               string(fleetv1beta1.ResourceScheduledConditionType),
 				Reason:             "ScheduleFailed",
 				Message:            fmt.Sprintf(resourcePlacementConditionScheduleFailedMessageFormat, c.ClusterName, c.Reason),
 				ObservedGeneration: crp.Generation,
@@ -774,6 +773,13 @@ func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, la
 				scheduledCondition.Message = fmt.Sprintf(resourcePlacementConditionScheduleFailedWithScoreMessageFormat, c.ClusterName, c.ClusterScore, c.Reason)
 			}
 		} else {
+			scheduledCondition = metav1.Condition{
+				Status:             metav1.ConditionTrue,
+				Type:               string(fleetv1beta1.ResourceScheduledConditionType),
+				Reason:             "ScheduleSucceeded",
+				Message:            fmt.Sprintf(resourcePlacementConditionScheduleSucceededMessageFormat, c.ClusterName, c.Reason),
+				ObservedGeneration: crp.Generation,
+			}
 			rp.ClusterName = c.ClusterName
 			if c.ClusterScore != nil {
 				scheduledCondition.Message = fmt.Sprintf(resourcePlacementConditionScheduleSucceededWithScoreMessageFormat, c.ClusterName, c.ClusterScore, c.Reason)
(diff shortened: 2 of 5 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -438,6 +438,12 @@ const (
 	// - "False" means some of them have failed.
 	// - "Unknown" means we haven't started the apply yet.
 	ResourcesAppliedConditionType ResourcePlacementConditionType = "ResourceApplied"
+
+	// PlacementScheduledConditionType indicates whether we have successfully scheduled the placement.
+	// Its condition status can be one of the following:
+	// - "True" means we have successfully scheduled the resources to fully satisfy the placement requirement.
+	// - "False" means we didn't fully satisfy the placement requirement. We will fill the Message field.
+	PlacementScheduledConditionType ResourcePlacementConditionType = "PlacementScheduled"
```
michaelawyu, line 446: Isn't there already a `ClusterResourcePlacementScheduledConditionType`?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -20,6 +22,17 @@ import (
 	"go.goms.io/fleet/pkg/utils/controller"
 )
 
+const (
+	// ApplyFailedReason is the reason string of placement condition when the selected resources fail to apply.
+	ApplyFailedReason = "ApplyFailed"
```
michaelawyu, line 27: Reason should be a camelCase string.

michaelawyu: Same for the other reasons down below.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -20,6 +22,17 @@ import (
...
+	// ApplyPendingReason is the reason string of placement condition when the selected resources are pending to apply.
+	ApplyPendingReason = "ApplyPending"
+	// ApplySucceededReason is the reason string of placement condition when the selected resources are applied successfully.
+	ApplySucceededReason = "ApplySucceeded"
+
+	resourcePlacementConditionMessageFormat = "%s is not seleted: %s"
```
michaelawyu, line 33: A typo: `selected`?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -542,3 +555,114 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
+		return status, nil
+	}
+
+	// TODO collect work status and build appliedCondition & resourcePlacementStatus
+	status.PlacementStatuses = buildResourcePlacementStatus(crp, latestSchedulingPolicySnapshot)
+	return status, nil
+}
+
+func buildScheduledCondition(crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot) metav1.Condition {
+	scheduledCondition := latestSchedulingPolicySnapshot.GetCondition(string(fleetv1beta1.PolicySnapshotScheduled))
```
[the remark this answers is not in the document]
michaelawyu: The condition here should explain the specifics of the scheduling result.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -542,3 +555,114 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
+
+	if scheduledCondition == nil ||
+		scheduledCondition.ObservedGeneration < latestSchedulingPolicySnapshot.Generation ||
```
michaelawyu, line 607: The policy snapshot generation does not change much I am afraid (we do not update its spec); this comparison here might not be necessary at all.

apis/placement/v1beta1/clusterresourceplacement_types.go
zhiying-lin: I also had the same confusion before and misunderstood the purpose of this fields.

```
	// PlacementStatuses contains a list of placement status on the clusters that are selected by PlacementPolicy.
	// Each selected cluster according to the latest resource placement is guaranteed to have a corresponding placementStatuses.
	// In the pickN case, there are N placement statuses where N = NumberOfClusters. Some of them may not have assigned
	// clusters when we cannot fill the required number of clusters.
	// +optional
	PlacementStatuses []ResourcePlacementStatus
```

We want to use this array to explain to user why the pickN cannot be satisfied and why the remaining replicas cannot be scheduled, so that the clusterName is optional.
when clusterName is empty, we will use the condition to explain why.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: i checked the kubernetes repo. They are using "CamelCase".

eg, https://github.com/openshift/kubernetes/blob/af29f64acbf9e6d33cde7f3194cf858f54058d3b/staging/src/k8s.io/api/extensions/v1beta1/types.go#L1237

another example, https://github.com/kubernetes/kubernetes/blob/e3bc35bc1bc7b399b564f2c4efc75eb9959e70f7/pkg/kubelet/kubelet_node_status.go#L341

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: yeah, it's not needed, just for defensive coding.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -542,3 +555,114 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
+		Type:               string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType),
+		Reason:             scheduledCondition.Reason,
+		Message:            scheduledCondition.Message,
+		ObservedGeneration: crp.Generation,
+	}
+}
+
+func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot) []fleetv1beta1.ResourcePlacementStatus {
+	res := make([]fleetv1beta1.ResourcePlacementStatus, 0, len(latestSchedulingPolicySnapshot.Status.ClusterDecisions))
+	for i, c := range latestSchedulingPolicySnapshot.Status.ClusterDecisions {
```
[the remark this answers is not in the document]
zhiying-lin: nope, explained in the api comment.

pkg/controllers/clusterresourceplacement/controller.go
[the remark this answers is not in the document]
zhiying-lin: hmm, they are all overview of the scheduling status. The specifies of the scheduling result will be populated in the placementStatuses.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -660,3 +676,122 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
 }
+
+// TODO: handle multiple resourceSnapshots
+func (r *Reconciler) buildPlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot, latestResourceSnapshot *fleetv1beta1.ClusterResourceSnapshot) (*fleetv1beta1.ClusterResourcePlacementStatus, error) {
+	status := &fleetv1beta1.ClusterResourcePlacementStatus{
+		SelectedResources: make([]fleetv1beta1.ResourceIdentifier, 0, len(latestResourceSnapshot.Spec.SelectedResources)),
+		PlacementStatuses: make([]fleetv1beta1.ResourcePlacementStatus, 0, 20), // pre-allocate with a reasonable capacity
+		Conditions:        []metav1.Condition{},
+	}
+	// TODO query the sub resourceSnapshots by using latestResourceSnapshot and then update the status.
```
ryanzhang-oss, line 687: work generator already does this, maybe we can refactor it to share the common logic?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -660,3 +676,122 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
+			Kind:      obj.GroupVersionKind().Kind,
+			Name:      obj.GetName(),
+			Namespace: obj.GetNamespace(),
+		}
+		status.SelectedResources = append(status.SelectedResources, ri)
+	}
+	scheduledCondition := buildScheduledCondition(crp, latestSchedulingPolicySnapshot)
+	meta.SetStatusCondition(&status.Conditions, scheduledCondition)
+
+	if scheduledCondition.Status == metav1.ConditionUnknown {
```
ryanzhang-oss, line 706: what if it's failed?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -660,3 +676,122 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
+	for i, c := range latestSchedulingPolicySnapshot.Status.ClusterDecisions {
+		scheduledCondition := metav1.Condition{
+			Status:             metav1.ConditionTrue,
+			Type:               string(fleetv1beta1.PlacementScheduledConditionType),
+			Reason:             "ScheduleSucceeded",
+			Message:            fmt.Sprintf(resourcePlacementConditionScheduleSucceededMessageFormat, c.ClusterName, c.Reason),
+			ObservedGeneration: crp.Generation,
+		}
+		var rp fleetv1beta1.ResourcePlacementStatus
+		if !latestSchedulingPolicySnapshot.Status.ClusterDecisions[i].Selected {
```
ryanzhang-oss, line 762: why is loop on both index and take an loop var? Here the index is used to get the element while  at line 773 the loop var is used.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -660,3 +676,122 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
+		Reason:             scheduledCondition.Reason,
+		Message:            scheduledCondition.Message,
+		ObservedGeneration: crp.Generation,
+	}
+}
+
+func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot) []fleetv1beta1.ResourcePlacementStatus {
+	res := make([]fleetv1beta1.ResourcePlacementStatus, 0, len(latestSchedulingPolicySnapshot.Status.ClusterDecisions))
+	for i, c := range latestSchedulingPolicySnapshot.Status.ClusterDecisions {
+		scheduledCondition := metav1.Condition{
```
ryanzhang-oss, line 754: shouldn't this only be created if the selected= true?

ryanzhang-oss: inside the else block below

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -660,3 +676,122 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
+			Status:             metav1.ConditionTrue,
+			Type:               string(fleetv1beta1.PlacementScheduledConditionType),
+			Reason:             "ScheduleSucceeded",
+			Message:            fmt.Sprintf(resourcePlacementConditionScheduleSucceededMessageFormat, c.ClusterName, c.Reason),
+			ObservedGeneration: crp.Generation,
+		}
+		var rp fleetv1beta1.ResourcePlacementStatus
```
ryanzhang-oss, line 761: we are not creating one RP per cluster decision, right?

ryanzhang-oss:
btw, I am not sure if we have recorded any events in the new CRP. We should.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: sure, will use it when handling multiple resourceSnapshots.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: added a comment in the code to explain why it's unknown only.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: we should create one RP for each selected cluster. For the unselected ones, we could have one RP for multiple unselected ones if they are failed because of the same reasons.

For now, i just created a separate RP for each unselected decision to begin with.

zhiying-lin:
> btw, I am not sure if we have recorded any events in the new CRP. We should.

not yet, will handle them in a separate PR

pkg/controllers/clusterresourceplacement/controller.go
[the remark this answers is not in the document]
ryanzhang-oss: It seems that the PolicySnapshotScheduled condition is the ClusterResourcePlacementScheduledConditionType condition on the CRP. Are we on the same page now?

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: the number of RP should be the same as the "numberOfClusters" in the pick N case. The current logic seems to create one per cluster?

zhiying-lin: oh, i thought even selectN is meet and we would still populate why others are not picked.

Will remove these clusters.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -438,6 +438,12 @@ const (
 	// - "False" means some of them have failed.
 	// - "Unknown" means we haven't started the apply yet.
 	ResourcesAppliedConditionType ResourcePlacementConditionType = "ResourceApplied"
+
+	// ResourceScheduledConditionType indicates whether we have successfully scheduled the selected resources.
+	// Its condition status can be one of the following:
+	// - "True" means we have successfully scheduled the resources to satisfy the placement requirement.
+	// - "False" means we didn't fully satisfy the placement requirement. We will fill the Message field.
+	ResourceScheduledConditionType ResourcePlacementConditionType = "ResourceScheduled"
```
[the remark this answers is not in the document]
zhiying-lin: it's not possible to be "unknown". we will only populate the ResourcePlacementStatus only when scheduler finishes the scheduling and the result is either "succeed" or "fail".

> I assume that this is the condition to indicate if the we have updated the binding to the latest resource snapshot, right?
> 
> In this case, maybe the description needs a bit more clarification.
i'm going to add another new condition type to reflect that. It will be in a separate PR when collecting work status

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -2214,3 +2214,896 @@ func TestHandleDelete(t *testing.T) {
...
+					Policy:     snapshotPolicy,
+					PolicyHash: policyHash,
+				},
+			},
+			latestResourceSnapshot: &fleetv1beta1.ClusterResourceSnapshot{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: fmt.Sprintf(fleetv1beta1.ResourceSnapshotNameFmt, testName, 0),
+					Labels: map[string]string{
+						fleetv1beta1.ResourceIndexLabel: "0",
+						fleetv1beta1.CRPTrackingLabel:   testName,
```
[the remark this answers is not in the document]
zhiying-lin: yeah, the caller will make sure that. the buildStatus func does not check here. So it does not matter whether it has or not.

ryanzhang-oss (approved):

PR #432, opened 2023-07-17
Rollout main logic

ryanzhang-oss:

the main rollout logic

Fixes #

I have:

### DIFF
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -1,0 +1,474 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+// Package rollout features a controller to do rollout.
+package rollout
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"time"
+
+	"golang.org/x/sync/errgroup"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/utils/controller"
+)
+
+// Reconciler recompute the cluster resource binding.
+type Reconciler struct {
+	client.Client
+
+	recorder record.EventRecorder
+}
+
+// Reconcile triggers a single binding reconcile round.
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	startTime := time.Now()
+	crpName := req.NamespacedName.Name
+	klog.V(2).InfoS("Start to rollout the bindings", "clusterResourcePlacement", crpName)
+
+	// add latency log
+	defer func() {
+		klog.V(2).InfoS("Rollout reconciliation loop ends", "clusterResourcePlacement", crpName, "latency", time.Since(startTime).Milliseconds())
+	}()
+
+	// Get the cluster resource placement
+	crp := fleetv1beta1.ClusterResourcePlacement{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: crpName}, &crp); err != nil {
+		if errors.IsNotFound(err) {
+			klog.V(4).InfoS("Ignoring NotFound clusterResourcePlacement", "clusterResourcePlacement", crpName)
+			return ctrl.Result{}, nil
+		}
+		klog.ErrorS(err, "Failed to get clusterResourcePlacement", "clusterResourcePlacement", crpName)
+		return ctrl.Result{}, controller.NewAPIServerError(true, err)
+	}
+	// check that the crp is not being deleted
+	if crp.DeletionTimestamp != nil {
+		klog.V(2).InfoS("Ignoring clusterResourcePlacement that is being deleted", "clusterResourcePlacement", crpName)
+		return ctrl.Result{}, nil
+	}
+
+	// check that it's actually rollingUpdate strategy
+	// TODO: support the rollout all at once type of RolloutStrategy
+	if crp.Spec.Strategy.Type != fleetv1beta1.RollingUpdateRolloutStrategyType {
+		klog.V(2).InfoS("Ignoring clusterResourcePlacement with non-rolling-update strategy", "clusterResourcePlacement", crpName)
+		return ctrl.Result{}, nil
+	}
+
+	// list all the bindings associated with the clusterResourcePlacement
+	bindingList := &fleetv1beta1.ClusterResourceBindingList{}
+	crpLabelMatcher := client.MatchingLabels{
+		fleetv1beta1.CRPTrackingLabel: crp.Name,
+	}
+	if err := r.Client.List(ctx, bindingList, crpLabelMatcher); err != nil {
+		klog.ErrorS(err, "Failed to list all the bindings associated with the clusterResourcePlacement",
+			"clusterResourcePlacement", crpName)
+		return ctrl.Result{}, controller.NewAPIServerError(true, err)
+	}
+	// take a deep copy of the bindings so that we can safely modify them
+	allBindings := make([]*fleetv1beta1.ClusterResourceBinding, 0, len(bindingList.Items))
+	for _, binding := range bindingList.Items {
+		allBindings = append(allBindings, binding.DeepCopy())
+	}
+
+	// find the latest resource snapshot
+	latestResourceSnapshotName, err := r.fetchLatestResourceSnapshot(ctx, crpName)
+	if err != nil {
+		klog.ErrorS(err, "Failed to find the latest resource snapshot for the clusterResourcePlacement",
+			"clusterResourcePlacement", crpName)
+		return ctrl.Result{}, err
+	}
+	klog.V(2).InfoS("Found the latest resource snapshot for the clusterResourcePlacement", "clusterResourcePlacement", crpName, "latestResourceSnapshotName", latestResourceSnapshotName)
+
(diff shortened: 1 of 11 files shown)


pkg/controllers/rollout/controller.go
```diff
@@ -0,0 +1,278 @@
...
+	// calculate the max number of bindings that can be added according to user specified maxUnavailable
+	maxSurgeNumber, _ := intstr.GetScaledValueFromIntOrPercent(crp.Spec.Strategy.RollingUpdate.MaxSurge, targetNumber, true)
+	maxReadyNumber := targetNumber + maxSurgeNumber
+	// This is the upper bound of the number of bindings that can be ready during the rolling update
+	// We count anything that still has work object on the hub cluster as can be ready since the member agent may have connection issue with the hub cluster
+	upperBoundReadyNumber := len(canBeReadyBindings)
+	for i := 0; i < maxReadyNumber-upperBoundReadyNumber; i++ {
+		// we don't differentiate if binding needs to be bound to new cluster or its resource snapshot needs to be updated
+		// TODO:
+		tobeUpgradedBinding = append(tobeUpgradedBinding, updateCandidates[i])
```
zhiying-lin, line 185: are we going to differentiate the new bound or snapshot updating in the first phase?

based on the maxSurge definition, we only need to append the new bound ones.

pkg/controllers/rollout/controller.go
```diff
@@ -0,0 +1,278 @@
...
+
+	// calculate the max number of bindings that can be unavailable according to user specified maxUnavailable
+	maxUnavailableNumber, _ := intstr.GetScaledValueFromIntOrPercent(crp.Spec.Strategy.RollingUpdate.MaxUnavailable, targetNumber, true)
+	minAvailableNumber := targetNumber - maxUnavailableNumber
+	// This is the lower bound of the number of bindings that can be available during the rolling update
+	// Since we can't predict the number of bindings that can be unavailable after they are applied, we don't take them into account
+	lowerBoundAvailable := len(readyBindings) - len(canBeUnavailableBindings)
+	for i := 0; i < lowerBoundAvailable-minAvailableNumber; i++ {
+		tobeRemovedBinding = append(tobeRemovedBinding, removeCandidates[i])
+	}
```
zhiying-lin, line 174: if we still have some capacity (lowerBoundAvaiable - minAvailable - len(toBeRevmovedBinding)), we can update the resourceSnapshot on the bounded binding.

pkg/controllers/clusterresourcesnapshot/watcher.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	// enqueue the CRP to the rollout controller queue
+	r.RolloutController.Enqueue(crp)
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Watcher) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&fleetv1beta1.ClusterResourceSnapshot{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
michaelawyu, line 92: Hi Ryan! This will lead the controller to ignore metadata changes; if the CRP controller adds a label/annotation, the event will not be picked up.

pkg/controllers/clusterresourcesnapshot/watcher.go
```diff
@@ -0,0 +1,105 @@
...
+		WithEventFilter(predicate.Funcs{
+			// skipping delete events as rollout controller doesn't handle that.
+			DeleteFunc: func(_ event.DeleteEvent) bool {
+				return false
+			},
+			// skipping update events as the specs are immutable.
+			UpdateFunc: func(evt event.UpdateEvent) bool {
```
michaelawyu, line 99: Hi Ryan! The spec is surely immutable, but the metadata is still subject to change; we probably shouldn't ignore these events.

pkg/controllers/clusterresourcesnapshot/watcher.go
michaelawyu: The resource snapshot spec is immutable (ideally), so this predicate will filter out every change I guess.

pkg/controllers/rollout/controller.go
```diff
@@ -0,0 +1,278 @@
...
+// Reconcile triggers a single binding reconcile round.
+func (r *Reconciler) Reconcile(ctx context.Context, key controller.QueueKey) (ctrl.Result, error) {
+	startTime := time.Now()
+	crpName, ok := key.(string)
+	if !ok {
+		err := fmt.Errorf("got a resource key %+v not of type namespaced key", key)
+		klog.ErrorS(err, "we have encountered a fatal error that can't be retried")
+		return ctrl.Result{}, controller.NewUnexpectedBehaviorError(err)
+	}
+	klog.V(2).InfoS("Start to rollout the bindings", "memberCluster", crpName)
```
michaelawyu, line 44: Hi Ryan! There is a type here; it should be `clusterResourcePlacement` rather than `memberCluster` I assume?

pkg/controllers/rollout/controller.go
```diff
@@ -0,0 +1,278 @@
...
+	removeCandidates := make([]*fleetv1beta1.ClusterResourceBinding, 0)
+	// the subset of removeCandidates that we pick to remove
+	tobeRemovedBinding := make([]*fleetv1beta1.ClusterResourceBinding, 0)
+
+	// calculate the cutoff time for a binding to be applied before so that it can be considered ready
+	var unavailablePeriod time.Duration
+	if crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds != nil {
+		unavailablePeriod = time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds)
+	} else {
+		unavailablePeriod = 60 * time.Second
```
michaelawyu, line 107: Hi Ryan! It might be better to make this default value a top-level constant IMO.

pkg/controllers/rollout/controller.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	readyTimeCutOff := time.Now().Add(-unavailablePeriod * time.Second)
```
michaelawyu, line 109: When using the default value, this would become 60 * time.Second * time.Second.

pkg/controllers/rollout/controller.go
```diff
@@ -0,0 +1,278 @@
...
+	// list all the bindings associated with the clusterResourcePlacement
+	bindingList := &fleetv1beta1.ClusterResourceBindingList{}
+	crpLabelMatcher := client.MatchingLabels{
+		fleetv1beta1.CRPTrackingLabel: crp.Name,
+	}
+	if err := r.Client.List(ctx, bindingList, crpLabelMatcher); err != nil {
+		klog.ErrorS(err, "Failed to list all the bindings associated with the clusterResourcePlacement",
+			"clusterResourcePlacement", klog.KObj(crp))
+		return nil, nil, controller.NewAPIServerError(false, err)
```
michaelawyu, line 118: Hi Ryan! Just to be certain, is this an uncached client?

pkg/controllers/rollout/controller.go
michaelawyu, line 185: An index overflow can also happen here I am afraid.

pkg/controllers/rollout/controller.go
```diff
@@ -0,0 +1,278 @@
...
+	maxReadyNumber := targetNumber + maxSurgeNumber
+	// This is the upper bound of the number of bindings that can be ready during the rolling update
+	// We count anything that still has work object on the hub cluster as can be ready since the member agent may have connection issue with the hub cluster
+	upperBoundReadyNumber := len(canBeReadyBindings)
+	for i := 0; i < maxReadyNumber-upperBoundReadyNumber; i++ {
+		// we don't differentiate if binding needs to be bound to new cluster or its resource snapshot needs to be updated
+		// TODO:
+		tobeUpgradedBinding = append(tobeUpgradedBinding, updateCandidates[i])
+	}
+	return tobeRemovedBinding, tobeUpgradedBinding, nil
```
michaelawyu, line 187: Hi Ryan! Is there any risk of the workflow getting stuck here?

michaelawyu: Say, for example, we have a CRP with the following setting:

* numOfClusters = 10
* maxSurge = 0
* maxUnavailable = 4 (minAvailable = 6)

And suppose currently in the system there are 5 scheduled bindings, 5 bound bindings, and 7 unscheduled bindings.

michaelawyu: Now, len(canBeReady) = 5 bound + 7 unscheduled = 12, which is > maxReady (10).  So no upgrade will happen.

And len(ready) = 5, which is < minAvailable, so no removal will happen either.

pkg/controllers/clusterresourcebinding/watcher.go
```diff
@@ -0,0 +1,79 @@
...
+		klog.ErrorS(err, "Invalid ClusterResourceBinding", "clusterResourceBinding", bindingRef)
+		return ctrl.Result{}, controller.NewUnexpectedBehaviorError(err)
+	}
+	// enqueue the CRP to the rollout controller queue
+	r.RolloutController.Enqueue(crp)
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Watcher) SetupWithManager(mgr ctrl.Manager) error {
```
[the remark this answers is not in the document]
ryanzhang-oss: yeah, I thought about that too. The reason I opted to use two watchers is there are some error handling logics. If we just use the event handler then the errors will not be retried and thus we could end up not handle the event.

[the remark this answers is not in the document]
ryanzhang-oss: the watcher reconciler will retry the error

pkg/controllers/clusterresourcesnapshot/watcher.go
ryanzhang-oss: yeah, is there any problem with filter out any metadata/status changes?

pkg/controllers/clusterresourcesnapshot/watcher.go
ryanzhang-oss: which metadata change should we handle?

pkg/controllers/rollout/controller.go
ryanzhang-oss: this is the CRD default value, will ETCD/API server honor the CRE schema?

pkg/controllers/rollout/controller.go
ryanzhang-oss: no, this is cached.

pkg/controllers/rollout/controller.go
```diff
@@ -0,0 +1,443 @@
...
+	// handle the bindings depends on its state
+	for i := 0; i < len(tobeUpgradedBinding); i++ {
+		binding := tobeUpgradedBinding[i]
+		switch binding.Spec.State {
+		// The only thing we can do on a bound binding is to update its resource resourceBinding
+		case fleetv1beta1.BindingStateBound:
+			binding.Spec.ResourceSnapshotName = latestResourceSnapshotName
+			binding.Status.LastResourceUpdateTime = metav1.Now()
+			errs.Go(func() error {
+				if err := r.Client.Update(cctx, binding); err != nil {
```
zhiying-lin, line 307: we only update the spec here and need another call to update the status.

PR #433, opened 2023-07-17
feat: add clusteraffinity plugin skeleton and some data structures

zhiying-lin:

Add the cluster affinity plugin.

Fixes #

I have:

The clusteraffinity plugin Prefilter/Filter/Prescore/Score and some of the unit tests will be in a separate PR.

### DIFF
--- a/pkg/scheduler/framework/plugins/clusteraffinity/scoring.go
+++ b/pkg/scheduler/framework/plugins/clusteraffinity/scoring.go
@@ -1,0 +1,31 @@
+package clusteraffinity
+
+import (
+	"context"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/scheduler/framework"
+)
+
+// PreScore allows the plugin to connect to the PreScore extension point in the scheduling
+// framework.
+func (p *Plugin) PreScore(
+	_ context.Context,
+	state framework.CycleStatePluginReadWriter,
+	_ *fleetv1beta1.ClusterSchedulingPolicySnapshot,
+) (status *framework.Status) {
+	// TODO not implemented.
+	_, _ = p.readPluginState(state)
+	return nil
+}
+
+// Score allows the plugin to connect to the Score extension point in the scheduling framework.
+func (p *Plugin) Score(
+	_ context.Context,
+	_ framework.CycleStatePluginReadWriter,
+	_ *fleetv1beta1.ClusterSchedulingPolicySnapshot,
+	_ *fleetv1beta1.MemberCluster,
+) (score *framework.ClusterScore, status *framework.Status) {
+	// TODO not implemented.
+	return nil, nil
+}
--- a/pkg/scheduler/framework/plugins/clusteraffinity/state.go
+++ b/pkg/scheduler/framework/plugins/clusteraffinity/state.go
@@ -1,0 +1,43 @@
+package clusteraffinity
+
+import (
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/scheduler/framework"
+)
+
+type pluginState struct {
+	// requiredAffinityTerms is a list of processed version of required cluster affinity terms.
+	requiredAffinityTerms []framework.AffinityTerm
+
+	// preferredAffinityTerms is a list of processed version of preferred cluster affinity terms.
+	preferredAffinityTerms *framework.PreferredAffinityTerms
+}
+
+// preparePluginState initializes the state for the plugin to use in the scheduling cycle.
+// TODO will call this func in the PreFilter stage.
+func preparePluginState(policy *fleetv1beta1.ClusterSchedulingPolicySnapshot) (*pluginState, error) {
+	if policy.Spec.Policy != nil ||
+		policy.Spec.Policy.Affinity == nil ||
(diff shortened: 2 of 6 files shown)


michaelawyu:
Hi Zhiying! The node affinity plugin technically speaking does not need to keep states. The plugin state is introduced for the topology spread constraints plugin as each run of the Filter/Score extension point must know a global state, i.e., the current spread across the fleet; this is not true for this plugin IMO.

pkg/scheduler/framework/plugins/clusteraffinity/state.go
```diff
@@ -0,0 +1,43 @@
...
+	// requiredAffinityTerms is a list of processed version of required cluster affinity terms.
+	requiredAffinityTerms []framework.AffinityTerm
+
+	// preferredAffinityTerms is a list of processed version of preferred cluster affinity terms.
+	preferredAffinityTerms *framework.PreferredAffinityTerms
+}
+
+// preparePluginState initializes the state for the plugin to use in the scheduling cycle.
+// TODO will call this func in the PreFilter stage.
+func preparePluginState(policy *fleetv1beta1.ClusterSchedulingPolicySnapshot) (*pluginState, error) {
```
michaelawyu, line 18: For example, this bookkeeping is not necessary as the extension points can read them directly; it does not save any cost.

pkg/scheduler/framework/types.go
```diff
@@ -0,0 +1,99 @@
+/*
```
michaelawyu, line 1: I am afraid that it might be the same for many utilities in this file as well. The extension point can calculate them directly rather than wrapping the same set of data in a different struct.

zhiying-lin:
discussed offline, prefer to convert the affinity terms in one place.
Moved the types.go to the clusteraffinity package.

pkg/scheduler/framework/plugins/clusteraffinity/state.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+)
+
+type pluginState struct {
+	// requiredAffinityTerms is a list of processed version of required cluster affinity terms.
+	requiredAffinityTerms []AffinityTerm
+
+	// preferredAffinityTerms is a list of processed version of preferred cluster affinity terms.
+	preferredAffinityTerms *PreferredAffinityTerms
```
michaelawyu, line 12: A nit here: for consistency here it should either all use aliases, or all use the plain types.

michaelawyu, line 12: A nit here: for consistency it should either all use aliases, or all use plain types.

pkg/scheduler/framework/plugins/clusteraffinity/types.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// preferredAffinityTerm is a "processed" representation of PreferredClusterSelector.
+type preferredAffinityTerm struct {
+	AffinityTerm
+	weight int32
+}
+
+// PreferredAffinityTerms is a "processed" representation of []PreferredClusterSelector.
+type PreferredAffinityTerms struct {
```
michaelawyu, line 32: A nit here: it can just be a type alias I assume?

pkg/scheduler/framework/plugins/clusteraffinity/scoring.go
```diff
@@ -0,0 +1,31 @@
...
+
+// PreScore allows the plugin to connect to the PreScore extension point in the scheduling
+// framework.
+func (p *Plugin) PreScore(
+	_ context.Context,
+	state framework.CycleStatePluginReadWriter,
+	_ *fleetv1beta1.ClusterSchedulingPolicySnapshot,
+) (status *framework.Status) {
+	// TODO not implemented.
+	_, _ = p.readPluginState(state)
```
michaelawyu, line 18: Just curious: the state in this PR is only read at this extension point; I assume it's for linting reasons?

michaelawyu (commented):
A few nits; others all LGTM.

pkg/scheduler/framework/plugins/clusteraffinity/scoring.go
zhiying-lin: yeah, it's only for the linting purpose

pkg/scheduler/framework/plugins/clusteraffinity/types.go
zhiying-lin: it has to be a struct, otherwise, cannot compute the score of these terms.

```
func (t *PreferredAffinityTerms) Score(cluster *fleetv1beta1.MemberCluster) int32 {
	var score int32
	for _, term := range t.terms {
		if term.AffinityTerm.Matches(cluster) {
			score += term.weight
		}
	}
	return score
}
```

michaelawyu: Why? You can still define methods for aliases

michaelawyu: https://go.dev/play/p/FUbp0RsFNn7

michaelawyu: This isn't a blocking thing in any way though, I do not have any strong opposition on this front.

michaelawyu (approved):
LGTM

PR #438, opened 2023-07-20
fix: improve work generator

ryanzhang-oss:

- Add parentResource label
- Handle resourceSnapshot not found
- Add more test

Fixes #

I have:

### DIFF
--- a/apis/placement/v1beta1/binding_types.go
+++ b/apis/placement/v1beta1/binding_types.go
@@ -85,11 +85,6 @@ type ResourceBindingStatus struct {
 	// Conditions is an array of current observed conditions for ClusterResourceBinding.
 	// +optional
 	Conditions []metav1.Condition `json:"conditions"`
-
-	// LastResourceUpdateTime is the last time the binding is updated to point to a new resource snapshot
-	// after it's created.
-	// +optional
-	LastResourceUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
 }
 
 // ResourceBindingConditionType identifies a specific condition of the ClusterResourceBinding.
--- a/apis/placement/v1beta1/binding_types.go
+++ b/apis/placement/v1beta1/binding_types.go
@@ -13,9 +13,6 @@ import (
 const (
 	// WorkFinalizer is used to make sure that the binding is not deleted until the work objects it generates are all deleted.
 	WorkFinalizer = fleetPrefix + "work-cleanup"
-
-	// ParentBindingLabel is the label that contains the name of the binding that generates the work.
-	ParentBindingLabel = fleetPrefix + "parent-resource-binding"
 )
 
 // +kubebuilder:object:root=true
@@ -88,6 +85,11 @@ type ResourceBindingStatus struct {
 	// Conditions is an array of current observed conditions for ClusterResourceBinding.
 	// +optional
 	Conditions []metav1.Condition `json:"conditions"`
+
+	// LastResourceUpdateTime is the last time the binding is updated to point to a new resource snapshot
+	// after it's created.
+	// +optional
+	LastResourceUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
 }
 
 // ResourceBindingConditionType identifies a specific condition of the ClusterResourceBinding.
@@ -100,14 +102,6 @@ const (
 	// - "False" means the corresponding work CR is not created yet.
 	// - "Unknown" means it is unknown.
 	ResourceBindingBound ResourceBindingConditionType = "Bound"
-
-	// ResourceBindingUpdated indicates whether (and when) a binding is updated to point
-	// to a new resource snapshot.
-	// Its condition status can be one of the following:
-	// - "True" means the binding has been updated.
-	// - "False" means the binding has not been updated yet.
-	// - "Unknown" means the update status is unknown.
-	ResourceBindingUpdated ResourceBindingConditionType = "Updated"
 
 	// ResourceBindingApplied indicates the applied condition of the given resources.
 	// Its condition status can be one of the following:
--- a/pkg/controllers/memberclusterplacement/membercluster_controller.go
+++ b/pkg/controllers/memberclusterplacement/membercluster_controller.go
@@ -38,7 +38,7 @@ func (r *Reconciler) Reconcile(_ context.Context, key controller.QueueKey) (ctrl
 	startTime := time.Now()
 	memberClusterName, ok := key.(string)
 	if !ok {
-		err := fmt.Errorf("got a resource key %+v not of type cluster wide key", key)
+		err := fmt.Errorf("got a resource key %+v not of type namespaced key", key)
 		klog.ErrorS(err, "we have encountered a fatal error that can't be retried")
 		return ctrl.Result{}, err
 	}
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -115,11 +115,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	if syncErr == nil {
 		return ctrl.Result{}, controller.NewUpdateIgnoreConflictError(updateErr)
 	}
-	// This is valid ONLY because we are 100% POSITIVE that the work generator uses the same client as the rollout controller and scheduler.
-	// Or we can miss the newly created master snapshot
 	if errors.Is(syncErr, errResourceSnapshotDeleted) {
-		// don't retry if the master resource snapshot is deleted
-		return ctrl.Result{}, nil
+		// this is expected to happen when the cache is not synced yet
+		return ctrl.Result{Requeue: true}, nil
 	}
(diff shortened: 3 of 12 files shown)


pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -68,45 +70,141 @@ var _ = Describe("Test clusterSchedulingPolicySnapshot Controller", func() {
...
+						Status:             metav1.ConditionFalse,
+						Reason:             syncWorkFailed,
+						ObservedGeneration: binding.GetGeneration(),
+					},
+				},
+			}
 			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)).Should(Succeed())
-			Expect(len(binding.Finalizers)).Should(Equal(0))
+			diff = cmp.Diff(wantMC, binding.Status, ignoreConditionOption)
+			Expect(diff).Should(BeEmpty(), fmt.Sprintf("binding(%s) mismatch (-want +got):\n%s", binding.Name, diff))
```
zhiying-lin, line 196: Could change to

Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)).Should(Succeed())
wantMC := fleetv1beta1.ResourceBindingStatus{...}
diff := cmp.Diff(wantMC, binding.Status, ignoreConditionOption)

move the wantMC closer to where it's used

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -68,45 +70,141 @@ var _ = Describe("Test clusterSchedulingPolicySnapshot Controller", func() {
...
+			Expect(binding.GetCondition(string(fleetv1beta1.ResourceBindingBound)).Message).Should(Equal(errResourceSnapshotDeleted.Error()))
+			// delete the binding
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)).Should(Succeed())
+			Expect(k8sClient.Delete(ctx, binding)).Should(Succeed())
+			By(fmt.Sprintf("resource binding  %s is deleted", binding.Name))
+			// check the binding is deleted
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)
+				return apierrors.IsNotFound(err)
+			}, timeout, interval).Should(BeTrue(), "Failed to get the expected work in hub cluster")
```
zhiying-lin, line 206:
```suggestion
			}, timeout, interval).Should(BeTrue(), "Expect the work to be deleted in the hub cluster")
```

zhiying-lin (commented):
hi ryan, the UT failed, you may need to fix them

apis/placement/v1beta1/binding_types.go
```diff
@@ -88,6 +85,11 @@ type ResourceBindingStatus struct {
 	// Conditions is an array of current observed conditions for ClusterResourceBinding.
 	// +optional
 	Conditions []metav1.Condition `json:"conditions"`
+
+	// LastResourceUpdateTime is the last time the binding is updated to point to a new resource snapshot
+	// after it's created.
+	// +optional
+	LastResourceUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
```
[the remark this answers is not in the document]
ryanzhang-oss: This is used as a HB. The last update time in condition doesn't change when the state doesn't change.

pkg/controllers/memberclusterplacement/membercluster_controller.go
```diff
@@ -38,7 +38,7 @@ func (r *Reconciler) Reconcile(_ context.Context, key controller.QueueKey) (ctrl
 	startTime := time.Now()
 	memberClusterName, ok := key.(string)
 	if !ok {
-		err := fmt.Errorf("got a resource key %+v not of type cluster wide key", key)
+		err := fmt.Errorf("got a resource key %+v not of type namespaced key", key)
```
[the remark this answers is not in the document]
ryanzhang-oss: there are two types of key, "cluster wide" or 'namespaced"

pkg/controllers/workgenerator/controller.go
```diff
@@ -77,8 +88,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	} else {
+		resourceBinding.SetConditions(metav1.Condition{
+			Status:             metav1.ConditionTrue,
+			Type:               string(fleetv1beta1.ResourceBindingBound),
+			Reason:             allWorkSyncedReason,
+			ObservedGeneration: resourceBinding.Generation,
+		})
+	}
+	// update the resource binding status
+	updateErr := r.Client.Status().Update(ctx, &resourceBinding)
```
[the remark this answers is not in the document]
ryanzhang-oss: if we don't update then the binding status then its status is still "bound". Is that correct?

pkg/controllers/workgenerator/controller.go
```diff
@@ -77,8 +88,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		klog.ErrorS(updateErr, "Failed to update the resourceBinding status", "resourceBinding", bindingRef)
+	}
+	if syncErr == nil {
+		return ctrl.Result{}, controller.NewUpdateIgnoreConflictError(updateErr)
+	}
+	if errors.Is(syncErr, errResourceSnapshotDeleted) {
+		// don't retry if the master resource snapshot is deleted
+		return ctrl.Result{}, nil
+	}
+	return ctrl.Result{}, syncErr
```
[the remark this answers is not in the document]
ryanzhang-oss: this is not right. We need to return syncErr if it's not nil.

pkg/controllers/workgenerator/controller_integration_test.go
ryanzhang-oss: the diff line used both `binding` and `wantMC`. Is there any reason why wantMC should be the closest to the cmp instead of binding? Is that it's more complicated?

pkg/controllers/workgenerator/controller.go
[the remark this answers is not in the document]
ryanzhang-oss: btw, I realized that  work generate may not be able to find the JUST created snapshot if the rollout/scheduler controller uses a different informer than the work generator. We don't have this problem now but this is kinda dangerous if we ever going to separate out the work geneartor.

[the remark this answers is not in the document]
ryanzhang-oss: maybe not worth the risk and just reconcile?

pkg/controllers/workgenerator/controller.go
```diff
@@ -77,8 +88,40 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			Status:             metav1.ConditionTrue,
+			Type:               string(fleetv1beta1.ResourceBindingBound),
+			Reason:             allWorkSyncedReason,
+			ObservedGeneration: resourceBinding.Generation,
+		})
+	}
+	// update the resource binding status
+	updateErr := r.Client.Status().Update(ctx, &resourceBinding)
+	if updateErr != nil {
+		klog.ErrorS(updateErr, "Failed to update the resourceBinding status", "resourceBinding", bindingRef)
```
michaelawyu, line 113: Should this return directly?

michaelawyu: * updateErr, no sync error -> return with no error
* no updateErr, no sync error -> return with no error
* updateErr, sync error, sync error is errResourceSnapshotDeleted -> return with no error
* no updateErr, sync error, sync error is errResourceSnapshotDeleted -> return with no error
*  no updateErr, sync error, sync error is not errResourceSnapshotDeleted -> return with error

ryanzhang-oss: > * updateErr, no sync error -> return with no error

Shouldn't this return an error?

> * no updateErr, no sync error -> return with no error
> * updateErr, sync error, sync error is errResourceSnapshotDeleted -> return with no error

same here

> * no updateErr, sync error, sync error is errResourceSnapshotDeleted -> return with no error

I realized that there is a race between the resourceSnapshot is created to the binding created pointing to it in the test environment. I think it won't happen in prod since scheduler/rollout controller share the same cache as the work generator.

> * no updateErr, sync error, sync error is not errResourceSnapshotDeleted -> return with error

ryanzhang-oss:
> hi ryan, the UT failed, you may need to fix them

Do you mean IT? All tests pass now.

zhiying-lin:
> > hi ryan, the UT failed, you may need to fix them
> 
> Do you mean IT? All tests pass now, I retried it 2 times.

oh, this comment was posted for the first review, 4 days ago before you fixed the IT.

pkg/controllers/workgenerator/controller.go
michaelawyu: Oh, sorry for the confusion; the comment was to document the observed behavior (before any new commits), not the desired one.

zhiying-lin (approved):

PR #448, opened 2023-07-25
feat: complete crp reconcile

zhiying-lin:
Complete the CRP reconcile.

I have:

Need to add the UT and IT in a separate PR

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -419,7 +419,7 @@ const (
 	// Its condition status can be one of the following:
 	// - "True" means all the selected resources are successfully applied to all the target clusters.
 	// - "False" means some of them have failed. We will place some of the detailed failure in the FailedResourcePlacement array.
-	// - "Unknown" means we haven't started the apply yet.
+	// - "Unknown" means we haven't started or finished the apply yet.
 	ClusterResourcePlacementAppliedConditionType ClusterResourcePlacementConditionType = "ClusterResourcePlacementApplied"
 )
 
@@ -428,19 +428,25 @@ const (
 type ResourcePlacementConditionType string
 
 const (
-	// ResourceWorkCreatedConditionType indicates whether we have created the corresponding work object under the
-	// per-cluster namespaces (i.e., fleet-member-<member-name>).
+	// ResourceWorkCreatedConditionType indicates whether we have created/updated the corresponding work objects under the
+	// per-cluster namespaces (i.e., fleet-member-<member-name>) which have the latest resources selected by the placement.
 	// Its condition status can be one of the following:
-	// - "True" means we have successfully created the corresponding work resources.
-	// - "False" means we have failed to creat the corresponding work resources. We will fill the Reason field.
-	// - "Unknown" means we don't have a scheduling decision yet.
+	// - "True" means we have successfully created the latest corresponding work resources or updated the existing work
+	// resources to the latest.
+	// - "False" means we have not created the latest corresponding work resources or updated the existing work resources
+	// to the latest.
+	// There are few possibilities:
+	// - In the processing state
+	// - Rollout controller has decided not to create the resources in this cluster for now to honor the rollout strategy
+	// configurations specified in the placement.
+	// - Work cannot be created because of the unknown failures.
 	ResourceWorkCreatedConditionType ResourcePlacementConditionType = "WorkCreated"
 
 	// ResourcesAppliedConditionType indicates whether the selected member cluster has applied the selected resources locally.
 	// Its condition status can be one of the following:
 	// - "True" means all the selected resources are successfully applied to the target cluster.
 	// - "False" means some of them have failed.
-	// - "Unknown" means we haven't started the apply yet.
+	// - "Unknown" means we haven't started or finished the apply yet.
 	ResourcesAppliedConditionType ResourcePlacementConditionType = "ResourceApplied"
 
 	// ResourceScheduledConditionType indicates whether we have successfully scheduled the selected resources.
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -12,17 +12,22 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	workv1alpha1 "sigs.k8s.io/work-api/pkg/apis/v1alpha1"
+
+	workapi "go.goms.io/fleet/pkg/controllers/work"
 
 	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	"go.goms.io/fleet/pkg/utils"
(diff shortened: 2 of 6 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -419,7 +419,7 @@ const (
 	// Its condition status can be one of the following:
 	// - "True" means all the selected resources are successfully applied to all the target clusters.
 	// - "False" means some of them have failed. We will place some of the detailed failure in the FailedResourcePlacement array.
-	// - "Unknown" means we haven't started the apply yet.
+	// - "Unknown" means we haven't started or finished the apply yet.
 	ClusterResourcePlacementAppliedConditionType ClusterResourcePlacementConditionType = "ClusterResourcePlacementApplied"
```
zhiying-lin, line 423: now i'm thinking the crp also needs the ClusterResourcePlacementWorkCreated condition

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -131,14 +178,28 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
+	crp.Status.SelectedResources = status.SelectedResources
+	crp.Status.PlacementStatuses = status.PlacementStatuses
+	crp.SetConditions(status.Conditions...)
+
+	if err := r.Client.Status().Update(ctx, crp); err != nil {
+		klog.ErrorS(err, "Failed to update the status of the clusterResourcePlacement", "clusterResourcePlacement", klog.KObj(crp))
+		return ctrl.Result{}, controller.NewUpdateIgnoreConflictError(err)
+	}
+	// We keep a slow reconcile loop here to periodically update the work status.
+	return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
```
zhiying-lin, line 202: 5 minutes may be too long. 2 mins?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -36,16 +41,58 @@ const (
...
 
-func (r *Reconciler) Reconcile(ctx context.Context, _ controller.QueueKey) (ctrl.Result, error) {
-	// TODO workaround to bypass lint check
-	return r.handleUpdate(ctx, nil)
+func (r *Reconciler) Reconcile(ctx context.Context, key controller.QueueKey) (ctrl.Result, error) {
+	name, ok := key.(string)
+	if !ok {
+		err := fmt.Errorf("get place key %+v not of type string", key)
+		klog.ErrorS(err, "We have encountered a fatal error that can't be retried, requeue after a day")
+		return ctrl.Result{RequeueAfter: time.Hour * 24}, controller.NewUnexpectedBehaviorError(err)
```
michaelawyu, line 63: This is a fatal error that even upon retries it would just fail again; do we need to bother with requeuing?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -678,7 +739,7 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
 }
 
 // TODO: handle multiple resourceSnapshots
-func (r *Reconciler) buildPlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot, latestResourceSnapshot *fleetv1beta1.ClusterResourceSnapshot) (*fleetv1beta1.ClusterResourcePlacementStatus, error) {
+func (r *Reconciler) buildPlacementStatus(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot, latestResourceSnapshot *fleetv1beta1.ClusterResourceSnapshot) (*fleetv1beta1.ClusterResourcePlacementStatus, error) {
 	status := &fleetv1beta1.ClusterResourcePlacementStatus{
```
michaelawyu, line 743: (The thread is about the section below, new line 750-764)

michaelawyu: The handleUpdate method already performs selectResources routine (and it has to do it every time); here it just deserializes what it just serializes, and this is expensive.

michaelawyu: Also, a nit about the `buildScheduledCondition` method: for the unknown type, it might be better for the message to be a little bit more user-facing, i.e., "scheduling is in progress", rather than "record the reason ..."

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -713,7 +774,7 @@ func (r *Reconciler) buildPlacementStatus(crp *fleetv1beta1.ClusterResourcePlace
 		appliedCondition := metav1.Condition{
 			Status:             metav1.ConditionUnknown,
 			Type:               string(fleetv1beta1.ClusterResourcePlacementAppliedConditionType),
-			Reason:             ApplyPendingReason,
+			Reason:             ApplyNotNeededReason,
```
michaelawyu, line 777: Emm, to me `pending` is good enough already.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -131,14 +178,28 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
 
-	// update the status based on the latestPolicySnapshot status
-	// update the status based on the work
-	return ctrl.Result{}, nil
+	status, err := r.buildPlacementStatus(ctx, crp, latestSchedulingPolicySnapshot, latestResourceSnapshot)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Here we don't reset the crp status so that LastTransitionTime will be updated only if the new status of the condition
```
michaelawyu, line 191: Just a nit: why not pass in the old status for updating?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -722,8 +783,12 @@ func (r *Reconciler) buildPlacementStatus(crp *fleetv1beta1.ClusterResourcePlace
 		return status, nil
 	}
 
-	// TODO collect works and build appliedCondition & resourcePlacementStatus
-	status.PlacementStatuses = buildResourcePlacementStatus(crp, latestSchedulingPolicySnapshot)
+	placementStatuses, appliedCondition, err := r.buildResourcePlacementStatusAndAppliedCondition(ctx, crp, latestSchedulingPolicySnapshot, latestResourceSnapshot)
```
michaelawyu, line 786: A nit: add a comment here to explain that after prev. check the status must be up-to-date.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -768,7 +833,7 @@ func classifyClusterDecisions(decisions []fleetv1beta1.ClusterDecision) (selecte
 	return selected, unselected
 }
 
-func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot) []fleetv1beta1.ResourcePlacementStatus {
+func (r *Reconciler) buildResourcePlacementStatusAndAppliedCondition(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement, latestSchedulingPolicySnapshot *fleetv1beta1.ClusterSchedulingPolicySnapshot, latestResourceSnapshot *fleetv1beta1.ClusterResourceSnapshot) ([]fleetv1beta1.ResourcePlacementStatus, metav1.Condition, error) {
 	res := make([]fleetv1beta1.ResourcePlacementStatus, 0, len(latestSchedulingPolicySnapshot.Status.ClusterDecisions))
 	decisions := latestSchedulingPolicySnapshot.Status.ClusterDecisions
 	selected, unselected := classifyClusterDecisions(decisions)
```
michaelawyu, line 839: This thread is about the section below (new line 841) and its related logic:

michaelawyu: For PickAll type, the scheduled condition (on policy snapshot) would not be false, as it is always fully scheduled; for the PickN type, the condition being false implies not being able to pick enough number of clusters. And in this case, we cannot pinpoint exactly which unselected clusters are the cause of the problem, so they have to be all included (or at least, included as much as possible) for the user to make an analysis.

michaelawyu:
Added some comments; this is a long PR, I will finish reviewing later today. Apologies for the delay.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: good catch! forget that we've already handled selectResources.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin:
```
         // PlacementStatuses contains a list of placement status on the clusters that are selected by PlacementPolicy.
	// Each selected cluster according to the latest resource placement is guaranteed to have a corresponding placementStatuses.
	// In the pickN case, there are N placement statuses where N = NumberOfClusters. Some of them may not have assigned
	// clusters when we cannot fill the required number of clusters.
	// +optional
	PlacementStatuses []ResourcePlacementStatus 
```

The implementation is following the placementStatuses definition. The expected size of placementStatuses is N = NumberOfClusters (will be 5 if using the example in the comment). The other two (scheduled 3) will be used to explain why. 

have a TODO in line 893. We will populate these two by summarizing the failure reasons.
For now, we just populate the sample failures.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -65,7 +112,7 @@ func (r *Reconciler) handleDelete(ctx context.Context, crp *fleetv1beta1.Cluster
 	controllerutil.RemoveFinalizer(crp, fleetv1beta1.ClusterResourcePlacementCleanupFinalizer)
 	if err := r.Client.Update(ctx, crp); err != nil {
 		klog.ErrorS(err, "Failed to remove crp finalizer", "clusterResourcePlacement", crpKObj)
-		return ctrl.Result{}, err
+		return ctrl.Result{}, controller.NewUpdateIgnoreConflictError(err)
```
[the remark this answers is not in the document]
zhiying-lin: defensive coding :)

pkg/controllers/clusterresourceplacement/controller.go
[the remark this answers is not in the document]
zhiying-lin: to me, we still need a way to tell cx even the CRP are not fully fulfilled, whether the resources are placed to the selected clusters.

We can make it clear in the public doc.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -819,7 +906,277 @@ func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, la
...
+			Type:   string(fleetv1beta1.ResourceWorkCreatedConditionType),
+			Reason: "UpdatingOrBlocked",
+			// Or it may fail to update because of the unknown failures.
+			Message:            "In the process of updating, or operation is blocked by the rollout strategy",
+			ObservedGeneration: crp.Generation,
+		}, nil
+	}
+
+	// if newWorkCounter > 0
+	return metav1.Condition{
```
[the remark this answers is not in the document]
zhiying-lin: hmm, if the work has been updated to the latest, it means the rollout has been done in the hub. i did not get why the rollout is blocked.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: @ryanzhang-oss what's your opinion about this? Do you think if we should distinguish the ApplyNotNeeded or ApplyPending cases?

i think it will be helpful.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: I'm fine with discarding the request here. We have similar cases when we handle the WithEventFilter. We just skip the request and log the error.

I will do that to keep it consistent.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -428,19 +428,25 @@ const (
 type ResourcePlacementConditionType string
 
 const (
-	// ResourceWorkCreatedConditionType indicates whether we have created the corresponding work object under the
-	// per-cluster namespaces (i.e., fleet-member-<member-name>).
+	// ResourceWorkCreatedConditionType indicates whether we have created/updated the corresponding work objects under the
```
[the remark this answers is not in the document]
zhiying-lin: will update to "WorkSynchronized"

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: I didn't fully get this

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -12,17 +12,22 @@ import (
...
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	workv1alpha1 "sigs.k8s.io/work-api/pkg/apis/v1alpha1"
+
+	workapi "go.goms.io/fleet/pkg/controllers/work"
```
ryanzhang-oss, line 30: empty line

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: I can see both sides of the story.  Pending could be interpreted as just an unknown. Since we already have an "unknown" status, I guess it doesn't hurt to provide a bit more information. The only downside it that we need to make sure we don't confuse the cx.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -819,7 +906,277 @@ func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, la
...
+				Condition: *appliedCond,
+			}
+			res = append(res, failedManifest)
+		}
+	}
+	return false, res
+}
+
+// parseResourceSnapshotIndexFromWork returns error when parsing the label which should never return error in production.
+func parseResourceSnapshotIndexFromWork(work *workv1alpha1.Work) (int, error) {
```
ryanzhang-oss, line 1173: probably is useful to other controllers

PR #449, opened 2023-07-26
feat: add the sameplacementaffinity plugin (filtering)

zhiying-lin:

- add sameplacementantiaffinity plugin and implement the filter extension point

Fixes #

I have:

Added unit tests

Scoring will be sent out in a separate PR.

### DIFF
--- a/pkg/scheduler/framework/plugins/sameplacementaffinity/filtering_test.go
+++ b/pkg/scheduler/framework/plugins/sameplacementaffinity/filtering_test.go
@@ -1,0 +1,113 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package sameplacementaffinity
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/scheduler/framework"
+)
+
+const (
+	clusterName = "member-1"
+)
+
+var (
+	cmpStatusOptions = cmp.Options{
+		cmpopts.IgnoreFields(framework.Status{}, "reasons", "err"),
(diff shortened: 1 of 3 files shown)


michaelawyu:
Hi Zhiying!

Can we separate the minor fixes into a separate PR?

michaelawyu:
Also, the sameplacementaffinity plugin needs to connect to the scoring stages.

michaelawyu:
This requires a change of CycleState that I haven't merged into the code base yet; I'll send the PR later today.

zhiying-lin:
> Can we separate the minor fixes into a separate PR?
sure, created https://github.com/Azure/fleet/pull/451

> Also, the sameplacementaffinity plugin needs to connect to the scoring stages.

I did not get this part. Why does the plugin need the scoring stage? is it because the scoring is required stage and has to be implemented?

michaelawyu:
Hi Zhiying! I replied that in chat, but the reason is that earlier we introduced a BoundOrScheduled score which helps the scheduler to prioritize clusters that already has an associated binding

michaelawyu:
Come to think of that it might be better to call it `PrevBoundOrScheduled` score :D

pkg/scheduler/framework/plugins/sameplacementaffinity/filtering_test.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func TestFilter(t *testing.T) {
+	tests := []struct {
+		name                     string
+		scheduledOrBoundBindings []*fleetv1beta1.ClusterResourceBinding
+		want                     *framework.Status
+	}{
+		{
+			name: "placement has already been scheduled",
```
michaelawyu, line 39: Just a nit, the name should be `placement has been unscheduled` I assume?

michaelawyu: (The matching binding is unscheduled)

michaelawyu (dismissed):
Just a question regarding tests, others LGTM.

pkg/scheduler/framework/plugins/sameplacementaffinity/filtering_test.go
zhiying-lin: The binding should be scheduled. Fixed.

PR #450, opened 2023-07-26
feat: v1beta1 IMC controller

Arvindthiru:

Fixes #

I have:

image
v1beta1 IMC with status 
image

- The assumption from the RP side is that the member agent is installed for a managed cluster that joins the fleet, so when the member agent is installed RP get the associated member cluster CR name and passes it as a ENV variable 
-> Which is use to construct the member cluster namespaces name that gets passed 
-> To the hubOptions which get used to create the hub config for the hub client used by the member agent 
-> Meaning the member agent runs under the assumption that it's targeting one Member cluster CR's namespace. 

- Link: https://github.com/Azure/fleet/blob/main/cmd/memberagent/main.go#L80. But in our case we have two versions of controllers for v1alpha1 & v1beta1 but the caveat is that they can only target member cluster CRs (v1alpha1, v1beta1 CRs) with the same name that's passed when the member agent was installed 

- Another concern is we use the same work controller for both versions of the IMC controllers

### DIFF
--- a/apis/types.go
+++ b/apis/types.go
@@ -9,6 +9,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	fleetv1alpha1 "go.goms.io/fleet/apis/v1alpha1"
 )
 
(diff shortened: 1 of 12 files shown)


apis/types.go
```diff
@@ -9,6 +9,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
```
ryanzhang-oss, line 12: not for this but it seems we might want to split this into their own dir?

ryanzhang-oss (approved):

PR #452, opened 2023-07-26
feat: scheduler source (1/): add cluster scheduling policy snapshot scheduler source controller

michaelawyu:

This PR adds a scheduler source controller for processing policy snapshot changes.

I have:

- [x] Integration tests

This PR overlaps with another scheduler source controller PR.


pkg/scheduler/sources/clusterschedulingpolicysnapshot/controller.go
```diff
@@ -0,0 +1,172 @@
...
+	if policySnapshot.DeletionTimestamp != nil {
+		// The policy snapshot has been deleted; ignore it.
+		return ctrl.Result{}, nil
+	}
+
+	// Verify if the policy snapshot is currently active.
+	isLatestVal, ok := policySnapshot.Labels[fleetv1beta1.IsLatestSnapshotLabel]
+	if !ok {
+		// The IsLatestSnapshot label is not present; normally this should never occur.
+		klog.ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("IsLatestSnapshotLabel is missing")),
```
zhiying-lin, line 64: change to warning?
cause it's possible when the pod may exit after removing the active label before creating a new snapshot. When the new pod is created, CRP controller could be recovered by itself and the label will be corrected.

For the NewUnexpectedBehaviorError, it means it cannot be recovered by itself and needs oncall person to be involved.

Should we remove this controller.NewUnexpectedBehaviorError wrapper?

pkg/utils/test_util.go
```diff
@@ -136,3 +140,66 @@ func (matcher AlreadyExistMatcher) FailureMessage(actual interface{}) (message s
 func (matcher AlreadyExistMatcher) NegatedFailureMessage(actual interface{}) (message string) {
 	return format.Message(actual, "not to be already exist")
 }
+
+// SchedulerWorkqueueKeyCollector helps collect keys from a scheduler work queue for testing
```
zhiying-lin, line 144: could you please move this to test/utils?

Would like to deprecate this file and we should not mix testing utils and non testing utils in a package.

pkg/scheduler/sources/clusterschedulingpolicysnapshot/controller.go
michaelawyu: The API mentioned that this label must be set for all policy snapshots; plus, for the deactivation part, this label should be set to `false` rather than removed I assume?

pkg/scheduler/sources/clusterschedulingpolicysnapshot/controller.go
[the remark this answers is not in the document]
michaelawyu: Done.

pkg/utils/test_util.go
michaelawyu: Done.

pkg/scheduler/sources/clusterschedulingpolicysnapshot/controller.go
zhiying-lin: you're right, the label should be always present but the value could be false.

test/utils/keycollector/keycollector.go
```diff
@@ -0,0 +1,76 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package keycollector
```
zhiying-lin, line 6: missing a package comment

zhiying-lin (dismissed):
one minor comment, others LGTM

test/utils/keycollector/keycollector.go
michaelawyu: Done.

zhiying-lin (approved):

PR #458, opened 2023-07-27
fix: remove the auto generated selectors for job

### LINKED ISSUE #456: [BUG] Jobs can't be propagate to member clusters
### **Describe the bug**
During my testing, I encountered an error when Fleet Manager tried to distribute Jobs to Member Clusters as below:
```
message: 'Failed to apply manifest: Job.batch 'hello-world-job' is invalid:
          [spec.template.metadata.labels[controller-uid]: Invalid value: map[string]string{'controller-uid':'91cb485c-9966-401a-b68e-2a1d3ffc1c98',
          'job-name':'hello-world-job'}: must be ''f81acbaf-b289-465d-923f-f022f6cf77bf'',
          spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{'controller-uid':'91cb485c-9966-401a-b68e-2a1d3ffc1c98'},
          MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: `selector` not auto-generated]'
        reason: appliedManifestFailed
        status: 'False'
```

### **Environment**
Please provide the following: 
- Hub cluster details
- Member cluster details

### **To Reproduce**
```
apiVersion: batch/v1
kind: Job
metadata:
  name: hello-world-job
spec:
  completions: 20
  parallelism: 5
  template:
    spec:
      containers:
      - name: hello-world
        image: busybox
        command: ["sh", "-c", "echo 'Hello, World!'"]
      restartPolicy: Never
```

### **Expected behavior**
Should see the job in the member clusters

ryanzhang-oss:

K8s api auto generated some [selector](https://github.com/kubernetes/kubernetes/blob/d4fde1e92a83cb533ae63b3abe9d49f08efb7a2f/pkg/registry/batch/job/strategy.go#L219)  for job based on the jobs' uid before it persist that will not work on the member cluster since its uid will change.

Fixes #456

I have:

UT

### DIFF
--- a/pkg/controllers/clusterresourceplacement/resource_selector.go
+++ b/pkg/controllers/clusterresourceplacement/resource_selector.go
@@ -355,12 +355,15 @@ func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
 			return nil, fmt.Errorf("failed to get the ports field in Serivce object, name =%s: %w", object.GetName(), err)
 		}
 	} else if object.GetKind() == "Job" && object.GetAPIVersion() == batchv1.SchemeGroupVersion.String() {
-		if manualSelector, exist, _ := unstructured.NestedString(object.Object, "spec", "manualSelector"); !exist || manualSelector == "true" {
+		if manualSelector, exist, _ := unstructured.NestedBool(object.Object, "spec", "manualSelector"); !exist || !manualSelector {
 			// remove the selector field added by the api-server if the job is not created with manual selector
 			// https://github.com/kubernetes/kubernetes/blob/d4fde1e92a83cb533ae63b3abe9d49f08efb7a2f/pkg/registry/batch/job/strategy.go#L219
-			unstructured.RemoveNestedField(object.Object, "spec", "selector")
+			// k8s used to add an old label called "controller-uid" but use a new label called "batch.kubernetes.io/controller-uid" after 1.26
+			unstructured.RemoveNestedField(object.Object, "spec", "selector", "matchLabels", "controller-uid")
+			unstructured.RemoveNestedField(object.Object, "spec", "selector", "matchLabels", "batch.kubernetes.io/controller-uid")
 			unstructured.RemoveNestedField(object.Object, "spec", "template", "metadata", "creationTimestamp")
-			unstructured.RemoveNestedField(object.Object, "spec", "template", "metadata", "labels")
+			unstructured.RemoveNestedField(object.Object, "spec", "template", "metadata", "labels", "controller-uid")
+			unstructured.RemoveNestedField(object.Object, "spec", "template", "metadata", "labels", "batch.kubernetes.io/controller-uid")
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -353,6 +354,14 @@ func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to get the ports field in Serivce object, name =%s: %w", object.GetName(), err)
 		}
+	} else if object.GetKind() == "Job" && object.GetAPIVersion() == batchv1.SchemeGroupVersion.String() {
+		if manualSelector, exist, _ := unstructured.NestedString(object.Object, "spec", "manualSelector"); !exist || manualSelector == "true" {
```
michaelawyu, line 358: Hi Ryan! This should be `!exist || manualSelector == "false"`, right? If the `manualSelector` is `true`, the customer has its own `.spec.selector` specified and we shouldn't clear the value.

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -353,6 +354,14 @@ func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
...
+			// remove the selector field added by the api-server if the job is not created with manual selector
+			// https://github.com/kubernetes/kubernetes/blob/d4fde1e92a83cb533ae63b3abe9d49f08efb7a2f/pkg/registry/batch/job/strategy.go#L219
+			unstructured.RemoveNestedField(object.Object, "spec", "selector")
+			unstructured.RemoveNestedField(object.Object, "spec", "template", "metadata", "creationTimestamp")
+			unstructured.RemoveNestedField(object.Object, "spec", "template", "metadata", "labels")
```
michaelawyu, line 363: This clears out all the labels on the pod template, not just the UID label set by the system, which could have some side effects on the user-side, e.g. the user may still want all the child pods to have some labels that are not used for job selectors -> though I am not 100% sure how many users would have such labels

zhiying-lin: echo chen's comment. Can we only clear the the controller-uid  when its value = the job uuid? same for the  job-name.

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -353,6 +354,17 @@ func generateRawContent(object *unstructured.Unstructured) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to get the ports field in Serivce object, name =%s: %w", object.GetName(), err)
 		}
+	} else if object.GetKind() == "Job" && object.GetAPIVersion() == batchv1.SchemeGroupVersion.String() {
+		if manualSelector, exist, _ := unstructured.NestedBool(object.Object, "spec", "manualSelector"); !exist || !manualSelector {
+			// remove the selector field added by the api-server if the job is not created with manual selector
```
zhiying-lin, line 359:
```suggestion
			// remove the selector field and labels added by the api-server if the job is not created with manual selector
			// whose value conflict with the ones created by the member cluster api server
```

zhiying-lin (approved):
one minor comment, others LTGM

PR #459, opened 2023-07-28
feat: add sameplacementaffinity plugin (scoring)

zhiying-lin:

- add scoring of the sameplacementaffinity plugin

Fixes #

I have:

Added UT

### DIFF
--- a/pkg/scheduler/framework/plugins/sameplacementaffinity/scoring_test.go
+++ b/pkg/scheduler/framework/plugins/sameplacementaffinity/scoring_test.go
@@ -1,0 +1,78 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package sameplacementaffinity
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/framework/plugins/sameplacementaffinity/scoring_test.go
```diff
@@ -0,0 +1,78 @@
...
+					Spec: fleetv1beta1.ResourceBindingSpec{
+						TargetCluster: "another-cluster",
+						State:         fleetv1beta1.BindingStateBound,
+					},
+				},
+			},
+			want: &framework.ClusterScore{ObsoletePlacementAffinityScore: 0},
+		},
+		{
+			name: "nil obsolete binding",
```
michaelawyu, line 56: A nit: it should be `no obsolete binding`?

michaelawyu (approved):
LGTM

pkg/scheduler/framework/plugins/sameplacementaffinity/scoring_test.go
zhiying-lin: the array is nil and prefer to use "nil obsolete binding" to indicate that.

PR #462, opened 2023-07-28
feat: Add a system environment variable "HUB_KUBE_HEADER" to allow sending custom header to hub's API Server

mingqishao:
The hub cluster need additional http header for authentication or authorization. This PR to add a new system environment variable "HUB_KUBE_HEADER" to allow sending custom header to hub's API Server

Fixes #

I have:

### DIFF
--- a/cmd/memberagent/main.go
+++ b/cmd/memberagent/main.go
@@ -7,12 +7,17 @@ package main
 
 //goland:noinspection ALL
 import (
+	"bufio"
 	"context"
 	"encoding/base64"
 	"errors"
 	"flag"
 	"fmt"
+	"io"
+	"net/http"
+	"net/textproto"
 	"os"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
@@ -31,6 +36,7 @@ import (
 
 	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	fleetv1alpha1 "go.goms.io/fleet/apis/v1alpha1"
+	"go.goms.io/fleet/pkg/client"
 	imcv1alpha1 "go.goms.io/fleet/pkg/controllers/internalmembercluster/v1alpha1"
 	imcv1beta1 "go.goms.io/fleet/pkg/controllers/internalmembercluster/v1beta1"
 	workapi "go.goms.io/fleet/pkg/controllers/work"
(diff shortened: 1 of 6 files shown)


cmd/memberagent/main.go
```diff
@@ -192,6 +198,18 @@ func buildHubConfig(hubURL string, useCertificateAuth bool, tlsClientInsecure bo
 			hubConfig.TLSClientConfig.CAData = caData
 		}
 	}
+
+	if header, ok := os.LookupEnv("HUB_KUBE_HEADER"); ok {
+		r := textproto.NewReader(bufio.NewReader(strings.NewReader(header)))
+		h, err := r.ReadMIMEHeader()
+		if err != nil && errors.Is(err, io.EOF) {
```
michaelawyu, line 205: Hi Mitch! Just in case other types of errors happen here, would it be better if we just check against `err`?

michaelawyu:
Other than the one comment things LGTM; though it might be better to place this under `pkg/util` I assume?

pkg/client/round_trippper.go
```diff
@@ -0,0 +1,32 @@
+package client
```
ryanzhang-oss, line 1: we have multiple util dir, can you move this one file there or inside one of those? I don't foresee this dir to have many other files so it's kinda weird to have a top dir under pkg just having one funciton.

pkg/client/round_trippper.go
```diff
@@ -0,0 +1,32 @@
...
+		header:                header,
+		delegatedRoundTripper: rt,
+	}
+}
+
+var _ http.RoundTripper = &customHeadersRoundTripper{}
+
+func (c *customHeadersRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	for key, values := range c.header {
+		if req.Header.Get(key) != "" {
```
ryanzhang-oss, line 24: when will this happen in our case? Is that the env has changed?

mingqishao: This will happen when the custom header is trying to override a other header, for example `HOST`, or `Content-Type`

pkg/client/round_trippper.go
mingqishao: I moved them to `pkg/utils/httpclient` in [this commit](https://github.com/Azure/fleet/pull/462/commits/d3cfe7eb3acc5746b6385fb31a6c09525482a693).

cmd/memberagent/main.go
```diff
@@ -192,6 +198,18 @@ func buildHubConfig(hubURL string, useCertificateAuth bool, tlsClientInsecure bo
 			hubConfig.TLSClientConfig.CAData = caData
 		}
 	}
+
+	if header, ok := os.LookupEnv("HUB_KUBE_HEADER"); ok {
```
ryanzhang-oss, line 202: can you add a comment on why we have this?

ryanzhang-oss (approved):

cmd/memberagent/main.go
mingqishao: I added comments to explain why we need HUB_KUBE_HEADER header.

ryanzhang-oss (approved):

PR #463, opened 2023-08-01
feat: flag for enableDenyWebhook

Arvindthiru:

Fixes #

I have:


charts/hub-agent/templates/deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
           args:
             - --leader-elect=true
             - --enable-webhook={{ .Values.enableWebhook }}
+            - --enable-deny-webhook={{ .Values.enableDenyWebhook }}
```
ryanzhang-oss, line 25: we can probably use a better name? How about something like safeguard or guardRail webhook?

ryanzhang-oss (approved):

PR #472, opened 2023-08-07
feat: complete crp status implementation

zhiying-lin:

- move work status utils into crp package
- finish implementing building crp status
- pass in the selectedResources directly when building crp status

I have:

- will move existing crp status utils into placement_status.go in a separate PR
- will add more unit tests in a separate PR


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -668,54 +653,42 @@ func parseResourceGroupHashFromAnnotation(s *fleetv1beta1.ClusterResourceSnapsho
...
+				ObservedGeneration: crp.Generation,
+			},
 		}
-		meta.SetStatusCondition(&status.Conditions, appliedCondition)
-		// skip populating detailed resourcePlacementStatus
-		return status, nil
+		crp.SetConditions(conditions...)
+		// skip populating detailed resourcePlacementStatus & work related conditions
+		// reset other status fields
+		crp.Status.PlacementStatuses = []fleetv1beta1.ResourcePlacementStatus{}
```
ryanzhang-oss, line 687: empty status?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -759,21 +732,35 @@ func classifyClusterDecisions(decisions []fleetv1beta1.ClusterDecision) (selecte
...
 
 	// In the pickN case, if the placement cannot be satisfied. For example, pickN deployment requires 5 clusters and
 	// scheduler schedules the resources on 3 clusters. We'll populate why the other two cannot be scheduled.
 	// Here it is calculating how many unscheduled resources there are.
 	unscheduledClusterCount := 0
-	if crp.Spec.Policy != nil &&
-		crp.Spec.Policy.PlacementType == fleetv1beta1.PickNPlacementType &&
-		crp.Spec.Policy.NumberOfClusters != nil {
-		unscheduledClusterCount = int(*crp.Spec.Policy.NumberOfClusters) - len(selected)
+	if crp.Spec.Policy != nil {
```
ryanzhang-oss, line 744: I think the policy can be nil. what is the `unscheduledClusterCount` in that case?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -0,0 +1,319 @@
...
+		klog.ErrorS(err, "Failed to parse the resource snapshot index label from latest clusterResourceSnapshot", "clusterResourcePlacement", crpKObj, "clusterResourceSnapshot", klog.KObj(latestResourceSnapshot))
+		return controller.NewUnexpectedBehaviorError(err)
+	}
+	// Used to build the work synchronized condition
+	oldWorkCounter := 0 // The work is pointing to the old resourceSnapshot.
+	newWorkCounter := 0 // The work is pointing to the latest resourceSnapshot.
+	// Used to build the work applied condition
+	pendingWorkCounter := 0 // The work has not been applied yet.
+
+	failedResourcePlacements := make([]fleetv1beta1.FailedResourcePlacement, 0, 10) // preallocate the memory
```
ryanzhang-oss, line 163: where do we get the magic number 10 from?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -0,0 +1,319 @@
...
+		isPending, failedManifests := buildFailedResourcePlacements(&workList.Items[i])
+		if isPending {
+			pendingWorkCounter++
+		}
+		if len(failedManifests) != 0 && len(failedResourcePlacements) <= maxFailedResourcePlacementLimit {
+			failedResourcePlacements = append(failedResourcePlacements, failedManifests...)
+		}
+	}
+
+	if len(failedResourcePlacements) > maxFailedResourcePlacementLimit {
```
ryanzhang-oss, line 197: can this happen when we have line 192 check already?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -0,0 +1,319 @@
...
+	workSynchronizedCondition, err := buildWorkSynchronizedCondition(crp, desiredWorkCounter, newWorkCounter, oldWorkCounter)
+	if err != nil {
+		return err
+	}
+	meta.SetStatusCondition(&status.Conditions, workSynchronizedCondition)
+	if newWorkCounter+oldWorkCounter == 0 {
+		// skip setting the applied work condition
+		return nil
+	}
+	workAppliedCondition := buildWorkAppliedCondition(crp, pendingWorkCounter > 0, len(failedResourcePlacements) > 0)
```
ryanzhang-oss, line 221: can  newNworkCount < desiredWorkCount but pendingWorkCount == 0 and no failedPlacement? It seems that if the sync condition is not true then we can always set appliedCondition as "unknown" or "false" depends on you interpret "unknown"

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -787,10 +774,28 @@ func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, la
...
-
-		// TODO populate the work status
 		meta.SetStatusCondition(&rp.Conditions, scheduledCondition)
-		res = append(res, rp)
+		if err := r.setWorkStatusForResourcePlacementStatus(ctx, crp, latestResourceSnapshot, &rp); err != nil {
+			return err
+		}
+		placementStatuses = append(placementStatuses, rp)
+
+		syncCondition := meta.FindStatusCondition(rp.Conditions, string(fleetv1beta1.ResourceWorkSynchronizedConditionType))
```
ryanzhang-oss, line 783: iirc, this is reading from the RP status which was set in function call in line 778? If so, why not return it from the function call directly?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -787,10 +774,28 @@ func buildResourcePlacementStatus(crp *fleetv1beta1.ClusterResourcePlacement, la
...
+		if syncCondition == nil || syncCondition.Status != metav1.ConditionTrue {
+			syncPendingCount++
+		} else {
+			syncSucceededCount++
+		}
+		appliedCondition := meta.FindStatusCondition(rp.Conditions, string(fleetv1beta1.ResourcesAppliedConditionType))
```
ryanzhang-oss, line 789: same here

pkg/controllers/clusterresourceplacement/placement_status.go
zhiying-lin: can newNworkCount < desiredWorkCount but pendingWorkCount == 0 and no failedPlacement?

it could, eg, desiredWorkCount = 5, we successfully create 3 new works.

I updated the status implementation here and build the applied status only on the newWork.
The appliedCondition will be unknown if the syncCondition is not true.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -439,26 +448,33 @@ const (
...
+	// Its condition status can be one of the following:
+	// - "True" means we have successfully created the latest corresponding work(s) or updated the existing work(s) to
+	// the latest.
+	// - "False" means we have not created the latest corresponding work(s) or updated the existing work(s) to the latest
+	// yet.
+	// There are few possibilities:
+	// - In the processing state
+	// - Rollout controller has decided not to create or update the resources in this cluster for now to honor the
+	// rollout strategy configurations specified in the placement.
+	// - Work cannot be created/updated because of the unknown failures.
```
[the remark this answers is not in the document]
ryanzhang-oss: I think the comment might be clearer to say the work is not created/updated for any reasons or something like that.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -0,0 +1,312 @@
...
+
+	synchronizedPendingReason   = "SynchronizedPending"
+	synchronizedSucceededReason = "SynchronizedSucceeded"
+
+	// ApplyFailedReason is the reason string of placement condition when the selected resources fail to apply.
+	ApplyFailedReason = "ApplyFailed"
+	// ApplyPendingReason is the reason string of placement condition when the selected resources are pending to apply.
+	ApplyPendingReason = "ApplyPending"
+	// ApplySucceededReason is the reason string of placement condition when the selected resources are applied successfully.
+	ApplySucceededReason = "ApplySucceeded"
```
ryanzhang-oss, line 43: I wonder why these 3 reasons are exported while others don't.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -0,0 +1,312 @@
...
+)
+
+var (
+	// We only include 100 failed resource placements even if there are more than 100.
+	maxFailedResourcePlacementLimit = 100
+)
+
+const (
+	// ClusterResourcePlacementStatus condition reasons
+	schedulingUnknownReason = "Scheduling"
```
ryanzhang-oss, line 33: SchedulePending?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -0,0 +1,312 @@
...
+
+	synchronizedPendingReason   = "SynchronizedPending"
```
ryanzhang-oss, line 35: SynchronizePending?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -0,0 +1,312 @@
...
+	resourceApplyFailedReason = "ApplyFailed"
+	// resourceApplyPendingReason is the reason string of placement condition when the selected resources are pending to apply.
+	resourceApplyPendingReason = "ApplyPending"
+	// resourceApplySucceededReason is the reason string of placement condition when the selected resources are applied successfully.
+	resourceApplySucceededReason = "ApplySucceeded"
+
+	// workSynchronizePendingReason is the reason string of placement condition when the work(s) are pending to synchronize.
+	workSynchronizePendingReason = "WorkSynchronizePending"
+	// workSynchronizedSucceededReason is the reason string of placement condition when the work(s) are synchronized successfully.
+	workSynchronizedSucceededReason = "WorkSynchronizedSucceeded"
```
ryanzhang-oss, line 56: WorkSynchronizeSucceeded

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: what's the semantic of nil policy? I thought it meant "applying to all clusters". What if the selected doesn't include all?

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: yeah, I am not too sure about the empty placement status in this case. For example, say the new policy will schedule the resources to one or two more clusters. Some of the placement statuses actually are still valid.

I do understand that users need to figure out if any of those status are valid.

pkg/controllers/clusterresourceplacement/placement_status.go
zhiying-lin: They're used by the integration tests. We could expose others if needed.

ryanzhang-oss (approved):

PR #475, opened 2023-08-10
feat: add default plugin list

zhiying-lin:

Create a default profile for the scheduler

Fixes #

I have:

### DIFF
--- a/pkg/scheduler/profile/profile.go
+++ b/pkg/scheduler/profile/profile.go
@@ -1,0 +1,38 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+// Package profile holds the definition of a scheduling Profile.
+package profile
+
+import (
+	"go.goms.io/fleet/pkg/scheduler/framework"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/profile/profile.go
```diff
@@ -0,0 +1,38 @@
...
+	"go.goms.io/fleet/pkg/scheduler/framework/plugins/topologyspreadconstraints"
+)
+
+const (
+	// defaultProfileName is the default name for the scheduling profile.
+	defaultProfileName = "DefaultProfile"
+)
+
+// NewDefaultProfile creates a default scheduling profile.
+func NewDefaultProfile() *framework.Profile {
```
ryanzhang-oss, line 23: where is this used?

ryanzhang-oss (approved):

pkg/scheduler/profile/profile.go
zhiying-lin: it will be used in the main func where it creates the scheduler, profile and framework.

PR #477, opened 2023-08-11
feat: add v1beta1 flag to support v1beta1 controllers

zhiying-lin:

- create v1beta1 controllers

Fixes #

I have:

May need to add flag for the webhook too. It depends on the test result and could be added in a separate PR.

### DIFF
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -25,6 +25,7 @@ import (
 	"go.goms.io/fleet/cmd/hubagent/options"
 	"go.goms.io/fleet/cmd/hubagent/workload"
 	mcv1alpha1 "go.goms.io/fleet/pkg/controllers/membercluster/v1alpha1"
+	mcv1beta1 "go.goms.io/fleet/pkg/controllers/membercluster/v1beta1"
 	fleetmetrics "go.goms.io/fleet/pkg/metrics"
 	"go.goms.io/fleet/pkg/webhook"
 	// +kubebuilder:scaffold:imports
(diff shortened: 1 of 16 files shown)


cmd/memberagent/main.go
```diff
@@ -53,7 +54,9 @@ var (
 	metricsAddr          = flag.String("metrics-bind-address", ":8090", "The address the metric endpoint binds to.")
 	enableLeaderElection = flag.Bool("leader-elect", false,
 		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
-	leaderElectionNamespace = flag.String("leader-election-namespace", "kube-system", "The namespace in which the leader election resource will be created.")
+	leaderElectionNamespace     = flag.String("leader-election-namespace", "kube-system", "The namespace in which the leader election resource will be created.")
+	enablePlacementV1Alpha1APIs = flag.Bool("enable-placement-v1alpha1-apis", true, "If set, the agents will watch for the placement v1alpha1 APIs.")
```
ryanzhang-oss, line 58: MC and IMC are cluster api, not placement?

cmd/memberagent/main.go
```diff
@@ -67,10 +70,32 @@ func init() {
...
+		handleExitFunc()
+		os.Exit(1)
+	}
+)
+
 func main() {
 	flag.Parse()
 	utilrand.Seed(time.Now().UnixNano())
-	defer klog.Flush()
+	defer handleExitFunc()
```
ryanzhang-oss, line 87: what's the difference here?

cmd/memberagent/main.go
```diff
@@ -86,7 +111,7 @@ func main() {
 	mcName := os.Getenv("MEMBER_CLUSTER_NAME")
 	if mcName == "" {
 		klog.ErrorS(errors.New("member cluster name cannot be empty"), "error has occurred retrieving MEMBER_CLUSTER_NAME")
-		os.Exit(1)
+		exitWithErrorFunc()
```
ryanzhang-oss, line 114: we already have the defer klog.Flush, why add another one in the call?

ryanzhang-oss: If this is needed, why not change it in the two error handlers too?

cmd/memberagent/main.go
zhiying-lin: right now there's no diff. in case, there are other funcs we won't miss before we exit.

cmd/memberagent/main.go
zhiying-lin: > we already have the defer klog.Flush, why add another one in the call?

defer happens when we return the main func.
Here it exits directly and the defer func won't be called.

cmd/memberagent/main.go
```diff
@@ -67,26 +70,48 @@ func init() {
 	metrics.Registry.MustRegister(fleetmetrics.JoinResultMetrics, fleetmetrics.LeaveResultMetrics, fleetmetrics.WorkApplyTime)
 }
 
+var (
+	handleExitFunc = func() {
```
michaelawyu, line 74: `klog` has a built-in `FlushAndExit` method.

cmd/hubagent/main.go
```diff
@@ -93,13 +94,25 @@ func main() {
...
 
 	klog.V(2).InfoS("starting hubagent")
-
-	if err = (&mcv1alpha1.Reconciler{
-		Client:                  mgr.GetClient(),
-		NetworkingAgentsEnabled: opts.NetworkingAgentsEnabled,
-	}).SetupWithManager(mgr); err != nil {
-		klog.ErrorS(err, "unable to create v1alpha1 controller", "controller", "MemberCluster")
-		exitWithErrorFunc()
+	if opts.EnableV1Alpha1APIs {
```
michaelawyu, line 97: Just curiois, this is intentionally made to be not mutually exclusive I assume?

michaelawyu: 'cause the code in other files from this PR is modified to support both controllers.

test/integration/suite_test.go
```diff
@@ -105,6 +105,7 @@ var _ = BeforeSuite(func() {
 	By("Setup custom controllers")
 	opts := options.NewOptions()
 	opts.LeaderElection.LeaderElect = false
+	opts.EnableV1Alpha1APIs = true
```
ryanzhang-oss, line 108: the default is true so this is technically not needed?

zhiying-lin: default was false before, now it's changed to true.

michaelawyu (dismissed):
LGTM ;)

ryanzhang-oss (approved):

PR #482, opened 2023-08-17
feat: add events for the crp and update crp status

zhiying-lin:

- add events for the crp 
- update crp status
- add logs

Fixes #

I have:

UT

need to add more integration tests

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -192,10 +192,10 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		klog.V(2).InfoS("Placement has been synchronized", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 		r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementSyncSuccess", "Successfully synchronized the placement")
 	}
-	if !isCRPApplied(oldCRP) && isCRPApplied(crp) {
-		klog.V(2).InfoS("Placement has been applied", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
-		r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementApplySuccess", "Successfully applied the placement")
-	}
+
+	// There is no need to check if the CRP is applied or not.
+	// If the applied condition is true, it means the scheduling is done & works have been synchronized.
+	// So that the rollout should be completed, and it will never reach to this line.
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -133,35 +137,68 @@ func (r *Reconciler) deleteClusterResourceSnapshots(ctx context.Context, crp *fl
...
+	if !isCRPScheduled(oldCRP) && isCRPScheduled(crp) {
+		klog.V(2).InfoS("Placement has been scheduled", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
+		r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementScheduleSuccess", "Successfully scheduled the placement")
+	}
+	if !isCRPSynchronized(oldCRP) && isCRPSynchronized(crp) {
+		klog.V(2).InfoS("Placement has been synchronized", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
+		r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementSyncSuccess", "Successfully synchronized the placement")
+	}
+	if !isCRPApplied(oldCRP) && isCRPApplied(crp) {
+		klog.V(2).InfoS("Placement has been applied", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
```
ryanzhang-oss, line 196: can  !isCRPApplied(oldCRP) && isCRPApplied(crp) be "true" but isRolloutCompleted(crp)  false?

which means isCRPApplied(crp)  is true but somehow isCRPScheduled(crp) or  isCRPSynchronized(crp) is false. That's kinda counter intuitive to me.

zhiying-lin: based on the current implementation, it's impossible. will remove it and add a comment instead.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -133,35 +137,68 @@ func (r *Reconciler) deleteClusterResourceSnapshots(ctx context.Context, crp *fl
...
-	// update the status based on the latestPolicySnapshot status
-	// update the status based on the work
-	return ctrl.Result{}, nil
+	if err := r.Client.Status().Update(ctx, crp); err != nil {
+		klog.ErrorS(err, "Failed to update the status", "clusterResourcePlacement", crpKObj)
+		return ctrl.Result{}, err
+	}
+	klog.V(2).InfoS("Updated the clusterResourcePlacement status", "clusterResourcePlacement", crpKObj)
+
+	if isRolloutCompleted(crp) {
```
ryanzhang-oss, line 177: I think we probably also need some tests on the event get triggered. Can leave this as UT task for later.

ryanzhang-oss (approved):

PR #483, opened 2023-08-17
chore: bump controller runtime to 1.14.6

ryanzhang-oss:
I just realized that our controller runtime is pretty old. I upgraded to a newer version but stop short of the 1.15.0 which has many breaking changes! The new one does not require "Fo"r in the build which fixed a hack we had on the rollout controller. The side effect of it is to force us to upgrade crossplane runtime and bring all the k8s.io lib to 1.26 which is the latest we support.

A good, surprising benefit is that the informer finally has a remove handler function that we can now handle resource type deletion.

Fixes #

I have:

No logic changes so just compile.

Just compiler fixes


pkg/controllers/rollout/controller.go
```diff
@@ -364,25 +364,7 @@ func (r *Reconciler) updateBindings(ctx context.Context, latestResourceSnapshotN
...
-				},
-				DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
-					return false
-				},
-				GenericFunc: func(genericEvent event.GenericEvent) bool {
-					return false
-				},
-			},
-		)).
+	return ctrl.NewControllerManagedBy(mgr).Named("rollout_controller").
```
zhiying-lin, line 367: wow, we don't need this any more. that's awesome!

zhiying-lin (approved):

pkg/controllers/rollout/controller.go
ryanzhang-oss: yeah, that's the reason I wanted to pick up this version. I forgot where I saw a bug/issue that they want to remove the requirement for "For" so I was like yeah, I want this.

PR #485, opened 2023-08-21
feat: add validResourceSelectors condition type

zhiying-lin:

- add a new condition type for crp
- validate the selectors first before creating any snapshots to avoid the unnecessary scheduling 

Fixes #

I have:

will add integration tests


pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -30,6 +30,9 @@ var (
 
 const (
 	// ClusterResourcePlacementStatus condition reasons
+	validResourceSelectorsReason   = "ResourceEligible"
+	invalidResourceSelectorsReason = "ResourceIneligible"
```
michaelawyu, line 34: This is used to for checking if the resource selectors are good; personally I feel that ineligible might not be most accurate, e.g. failing due to invalid GVRs rather than selecting non-cluster-scoped resources.

michaelawyu: A simple `invalidResourceSelector` woud do I guess?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -149,11 +150,33 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
+			Message:            fmt.Sprintf("Resources are not eligible: %v", err),
+			ObservedGeneration: crp.Generation,
+		}
+		crp.SetConditions(validResourceSelectorCondition)
+		if updateErr := r.Client.Status().Update(ctx, crp); updateErr != nil {
+			klog.ErrorS(updateErr, "Failed to update the status", "clusterResourcePlacement", crpKObj)
+			return ctrl.Result{}, updateErr
+		}
+		// We requeue here to check if the runtime resources are fixed.
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, err
```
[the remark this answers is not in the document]
michaelawyu: I ask this as from the workflow we use for gathering resources, most of the errors are either transient (API server errors), or not recoverable (invalid/not supported GVR, namespace scoped resources, reserved objects, etc.).

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -418,6 +418,11 @@ type FailedResourcePlacement struct {
 type ClusterResourcePlacementConditionType string
 
 const (
+	// ClusterResourcePlacementValidResourceSelectorsConditionType indicates whether cluster resource selectors have been
+	// recognized as valid.
+	// This will be false if the resource is not a cluster scope resource, or it's a reserved namespace or GVR is invalid.
+	ClusterResourcePlacementValidResourceSelectorsConditionType ClusterResourcePlacementConditionType = "ValidResourceSelectors"
```
ryanzhang-oss, line 424: we will have a webhook to validate CRP, why do we need a condition for this?

zhiying-lin: webhook may catch most of errors but it still could happen that, webhook validates the selector and then CRD is deleted during the runtime and gvk becomes invalid when crp controller validates.

PR #486, opened 2023-08-21
fix: run scheduler in a separate goroutine

michaelawyu:

This PR features a simple fix that runs the scheduler in a separate goroutine; otherwise the main function will be blocked.

I have:

N/A

### DIFF
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -9,6 +9,7 @@ import (
 	"flag"
 	"os"
 	"strings"
+	"sync"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -42,6 +43,8 @@ var (
 		handleExitFunc()
 		os.Exit(1)
 	}
+
+	wg sync.WaitGroup
 )
 
 const (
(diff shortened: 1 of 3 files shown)


cmd/hubagent/workload/setup.go
```diff
@@ -188,7 +188,10 @@ func SetupControllers(ctx context.Context, mgr ctrl.Manager, config *rest.Config
 		defaultSchedulingQueue := queue.NewSimpleClusterResourcePlacementSchedulingQueue()
 		defaultScheduler := scheduler.NewScheduler("DefaultScheduler", defaultFramework, defaultSchedulingQueue, mgr)
 		klog.Info("Starting the scheduler")
-		defaultScheduler.Run(ctx)
+		// Scheduler must run in a separate goroutine as Run() is a blocking call.
+		go func() {
```
ryanzhang-oss, line 192: when will this go routine dangle on exit? Do we not want to monitor/wait for it?

michaelawyu: Hi Ryan! The scheduler is set to exit on context cancellation (same for the controller manager).

michaelawyu: I will add a waitgroup to make things clearer for both.

zhiying-lin (approved):

cmd/hubagent/main.go
```diff
@@ -42,6 +43,8 @@ var (
 		handleExitFunc()
 		os.Exit(1)
 	}
+
+	wg sync.WaitGroup
```
zhiying-lin, line 47: declare the wg in the main func instead? otherwise, the whole main package can update the wg.

zhiying-lin (approved):

PR #487, opened 2023-08-21
feat: add API version toggles in hub agent helm charts

michaelawyu:

This PR adds toggles for API versions in the hub agent helm chart to allow usage of v1beta1 APIs on the upstream end.

I have:

N/A

### DIFF
--- a/charts/hub-agent/values.yaml
+++ b/charts/hub-agent/values.yaml
@@ -33,3 +33,6 @@ affinity: {}
 
 secret:
   name: hub-kubeconfig-secret
+
+enableV1Alpha1APIs: true
+enableV1Beta1APIs: false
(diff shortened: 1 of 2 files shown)


charts/hub-agent/values.yaml
```diff
@@ -33,3 +33,6 @@ affinity: {}
 
 secret:
   name: hub-kubeconfig-secret
+
+enableV1Alpha1APIs: true
+enableV1Beta1APIs: false
```
ryanzhang-oss, line 38: Those are the default already, right? Do we still need this?

michaelawyu: Hi Ryan! This toggle is used for the (new) E2E tests, which installs agents using these helm charts and needs the v1beta1 APIs to be enabled.

zhiying-lin (approved):

charts/hub-agent/values.yaml
ryanzhang-oss: I merged too soon. Why is the enableV1Beta1API false then?

PR #488, opened 2023-08-21
fix: use correct packages in newly added unit tests

michaelawyu:

This PR fixes a problem in PR #480, where outdated checks fail to detect an inconsistency between the API package in use by the PR and that in the latest main branch.

I have:

N/A


michaelawyu:
Merging this PR without approval as it is related to #480, which handles a commit that should have been included there (#480 has been approved). The general idea is that #480 was sent a while ago and did not reflect the latest API change in the main branch; if there's any concern, please let me know.

PR #493, opened 2023-08-22
fix: start up the CRP controller in main.go

michaelawyu:

This PR fixes an issue where the CRP controller is not started up in main.go.

I have:

N/A

This PR includes some contents from #486.

### DIFF
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -120,6 +120,30 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 	if opts.EnableV1Beta1APIs {
 		klog.Info("Setting up clusterResourcePlacement v1beta1 controller")
 		clusterResourcePlacementControllerV1Beta1 = controller.NewController(crpControllerV1Beta1Name, controller.NamespaceKeyFunc, crpc.Reconcile, rateLimiter)
+
+		// Start up the controller.
+		//
+		// Note that, with v1alpha1 controllers, all custom ones are started up alongside the
+		// resource change detector, when its Start() method is invoked by the controller
+		// manager. This is no longer the case with v1beta1 controllers, specifically, the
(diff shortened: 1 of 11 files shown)


cmd/hubagent/workload/setup.go
```diff
@@ -119,6 +120,30 @@ func SetupControllers(ctx context.Context, mgr ctrl.Manager, config *rest.Config
...
+
+		// Start up the controller.
+		//
+		// Note that, with v1alpha1 controllers, all custom ones are started up alongside the
+		// resource change detector, when its Start() method is invoked by the controller
+		// manager. This is no longer the case with v1beta1 controllers, specifically, the
+		// v1beta1 CRP controller, as it depends on its own watchers. As a result, here
+		// this controller is spun up separately.
+		//
+		// The controller must run in a separate goroutine as Run() is a blocking call.
```
ryanzhang-oss, line 132: This reconciler is added to the mgr at line 150 (old line). Is that not enough?

cmd/hubagent/workload/setup.go
```diff
@@ -119,6 +120,30 @@ func SetupControllers(ctx context.Context, mgr ctrl.Manager, config *rest.Config
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			err := clusterResourcePlacementControllerV1Beta1.Run(ctx, opts.ConcurrentClusterPlacementSyncs)
```
zhiying-lin, line 137: can we run the controller in the changeDetector start() func, similar to v1alpha1 controller

michaelawyu: Done, as we previously discussed.

cmd/hubagent/workload/setup.go
michaelawyu: Hi Ryan! The one at line 150 is the watcher for the v1beta1 CRP controller (it pushes keys to the workqueue); the v1beta1 CRP controller is not started at this moment (its Run() is not called).

zhiying-lin (approved):

PR #496, opened 2023-08-23
interface: add work apis

zhiying-lin:

- copied the work-apis and made few changes
- changed the crd category

Fixes #

I have:

- rename the "ResourceIdentifier" as "WorkResourceIdentifier" so that it won't conflict with our own definition.
- fix the typo in "AppliedtWorkStatus"
- update the category of the work & appliedwork to "fleet-placement"


apis/cluster/v1beta1/internalmembercluster_types.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // +kubebuilder:object:root=true
-// +kubebuilder:resource:scope=Namespaced,categories={fleet},shortName=imc
+// +kubebuilder:resource:scope=Namespaced,categories={fleet-cluster},shortName=imc
```
michaelawyu, line 14: It might be better to specify two categories here: `fleet`, and `fleet-cluster`.

apis/placement/v1beta1/appliedwork_types.go
```diff
@@ -0,0 +1,91 @@
+/*
```
michaelawyu, line 1: A nit: would it be better if we have a comment here documenting the origin of this spec?

michaelawyu:
Just for clarification, the change of reference will be sent in another PR, right?

apis/placement/v1beta1/appliedwork_types.go
ryanzhang-oss: we probably need to include the original copy right too

apis/cluster/v1beta1/internalmembercluster_types.go
ryanzhang-oss: can one use selector to select CRDs based on its category?

zhiying-lin: They can use category to select CRs

zhiying-lin:
> Just for clarification, the change of reference will be sent in another PR, right?

Correct, will send out few PRs

michaelawyu (approved):
LGTM ;)

PR #490, opened 2023-08-22
test: add basic integration test for crp

zhiying-lin:
### Description of your changes

- add basic integration tests

Fixes #

I have:

- [X] Run `make reviewable` to ensure this PR is ready for review.

### How has this code been tested

IT

PR #498, opened 2023-08-23
fix: report correct condition when resources are successfully applied

michaelawyu:

This PR fixes an issue where correct `Applied` condition is not added when a CRP has been processed as expected.

I have:

N/A


pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -100,7 +100,7 @@ func buildClusterResourcePlacementApplyCondition(crp *fleetv1beta1.ClusterResour
 		}
 	}
 
-	if !isSync || pendingCount > 0 {
```
zhiying-lin, line 103: This pendingCount means the work status itself is in the pending and isSync means whether the work has been updated to the latest resource. They're different.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -930,7 +930,7 @@ func (r *Reconciler) setResourcePlacementStatusAndResourceConditions(ctx context
 	}
 	crp.Status.PlacementStatuses = placementStatuses
 	crp.SetConditions(buildClusterResourcePlacementSyncCondition(crp, syncPendingCount, syncSucceededCount))
-	crp.SetConditions(buildClusterResourcePlacementApplyCondition(crp, syncPendingCount > 0, appliedPendingCount, appliedSucceededCount, appliedFailedCount))
+	crp.SetConditions(buildClusterResourcePlacementApplyCondition(crp, syncPendingCount == 0, appliedPendingCount, appliedSucceededCount, appliedFailedCount))
```
zhiying-lin, line 933: it's fixed in #490

pkg/controllers/clusterresourceplacement/placement_status.go
michaelawyu: But with isSync <-> (pendCount == 0) there's no need to do the comparison here I am afraid.

michaelawyu: i.e., if isSync is false, syncPendingCount must be > 0, since isSync is determined by pendingCount == 0 and pendingCount cannot be negative.

michaelawyu: Discussed with Zhiying offline; it was a missight on my end, the variable is about applied parts.

zhiying-lin (approved):

PR #500, opened 2023-08-24
feat: Webhook for reserved namespace resources part-1

Arvindthiru:

Fixes #

I have:

This PR provides a generic way to for the fleet guard rail to handle all name spaced resources in a group. This is part 1 a follow up PR for the remaining api groups

- autoscaling/v2 - horizontalpodautoscalers 
- cluster.kubernetes-fleet.io/v1beta1 - imc
- coordination.k8s.io/v1 - leases
- discovery.k8s.io/v1 - endpointslices
- events.k8s.io/v1 - events (already being selected)
- networking.k8s.io/v1 - Ingress class
- policy/v1 - PodDisruptionBudgets
- storage.k8s.io/v1 - csistoragecapacities

Groups covered in E2E for resources:

- **apps/v1**
  - Deployment, DaemonSet, ReplicaSet
- **core/v1**
  - Pod, Service, ConfigMap, Secret
- **batch/v1**
  - CronJob, Job
- **rbac/v1**
  - Role, RoleBinding

### DIFF
--- a/pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
+++ b/pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
@@ -8,14 +8,10 @@ import (
 	"strings"
 
 	admissionv1 "k8s.io/api/admission/v1"
-	appsv1 "k8s.io/api/apps/v1"
-	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -36,13 +32,8 @@ const (
 var (
 	crdGVK       = metav1.GroupVersionKind{Group: v1.SchemeGroupVersion.Group, Version: v1.SchemeGroupVersion.Version, Kind: "CustomResourceDefinition"}
 	mcGVK        = metav1.GroupVersionKind{Group: fleetv1alpha1.GroupVersion.Group, Version: fleetv1alpha1.GroupVersion.Version, Kind: "MemberCluster"}
+	imcGVK       = metav1.GroupVersionKind{Group: fleetv1alpha1.GroupVersion.Group, Version: fleetv1alpha1.GroupVersion.Version, Kind: "InternalMemberCluster"}
 	namespaceGVK = metav1.GroupVersionKind{Group: corev1.SchemeGroupVersion.Group, Version: corev1.SchemeGroupVersion.Version, Kind: "Namespace"}
-
-	appsV1GV        = schema.GroupVersion{Group: appsv1.SchemeGroupVersion.Group, Version: appsv1.SchemeGroupVersion.Version}
-	rbacV1GV        = schema.GroupVersion{Group: rbacv1.SchemeGroupVersion.Group, Version: rbacv1.SchemeGroupVersion.Version}
-	coreV1GV        = schema.GroupVersion{Group: corev1.SchemeGroupVersion.Group, Version: corev1.SchemeGroupVersion.Version}
-	batchV1GV       = schema.GroupVersion{Group: batchv1.SchemeGroupVersion.Group, Version: batchv1.SchemeGroupVersion.Version}
-	fleetV1Alpha1GV = schema.GroupVersion{Group: fleetv1alpha1.GroupVersion.Group, Version: fleetv1alpha1.GroupVersion.Version}
 )
 
 // Add registers the webhook for K8s built-in object types.
@@ -68,8 +59,6 @@ func (v *fleetResourceValidator) Handle(ctx context.Context, req admission.Reque
 	var response admission.Response
 	if req.Operation == admissionv1.Create || req.Operation == admissionv1.Update || req.Operation == admissionv1.Delete {
 		switch {
-		case req.Namespace != "":
-			response = v.handleNamespacedResources(ctx, req)
 		case req.Kind == crdGVK:
 			klog.V(2).InfoS("handling CRD resource", "GVK", crdGVK, "namespacedName", namespacedName, "operation", req.Operation)
 			response = v.handleCRD(req)
@@ -79,6 +68,12 @@ func (v *fleetResourceValidator) Handle(ctx context.Context, req admission.Reque
 		case req.Kind == namespaceGVK:
 			klog.V(2).InfoS("handling namespace resource", "GVK", namespaceGVK, "namespacedName", namespacedName, "operation", req.Operation)
 			response = v.handleNamespace(req)
+		case req.Kind == imcGVK:
+			klog.V(2).InfoS("handling internal member cluster resource", "GVK", imcGVK, "namespacedName", namespacedName, "operation", req.Operation)
+			response = v.handleInternalMemberCluster(ctx, req)
+		case req.Namespace != "":
+			klog.V(2).InfoS(fmt.Sprintf("handling %s resource", req.Kind.Kind), "GVK", req.Kind, "namespacedName", namespacedName, "operation", req.Operation)
+			response = v.handleResource(req)
 		default:
 			klog.V(2).InfoS("resource is not monitored by fleet resource validator webhook", "GVK", req.Kind.String(), "namespacedName", namespacedName, "operation", req.Operation)
 			response = admission.Allowed(fmt.Sprintf("user: %s in groups: %v is allowed to modify resource with GVK: %s", req.UserInfo.Username, req.UserInfo.Groups, req.Kind.String()))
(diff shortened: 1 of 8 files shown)


pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -159,6 +188,69 @@ func (v *fleetResourceValidator) handleNamespace(req admission.Request) admissio
...
+// handleService allows/denies the request to modify service after validation.
+func (v *fleetResourceValidator) handleService(req admission.Request) admission.Response {
+	var service corev1.Service
+	if err := v.decodeRequestObject(req, &service); err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	return validation.ValidateUserForResource(service.Kind, types.NamespacedName{Name: service.Name, Namespace: service.Namespace}, v.whiteListedUsers, req.UserInfo)
+}
+
+// handleService allows/denies the request to modify secret after validation.
```
michaelawyu, line 209: Hi Arvind! It should be `handleSecret` I assume?

pkg/webhook/validation/uservalidation.go
```diff
@@ -48,10 +49,20 @@ func ValidateUserForFleetCRD(group string, namespacedName types.NamespacedName,
 // ValidateUserForResource checks to see if user is allowed to modify argued resource.
 func ValidateUserForResource(resKind string, namespacedName types.NamespacedName, whiteListedUsers []string, userInfo authenticationv1.UserInfo) admission.Response {
```
michaelawyu, line 50: Hi Arvind! I couldn't help noticing that the logic of two functions are essentially the same; is this expected?

Arvindthiru: I wanted to add the node group check only for resources related to pod that's the only difference

michaelawyu: I see. The thing is, `system:nodes` is a group primarily reserved for kubelet and is considered to be owned by the system. If we would like to simplify things we can just greenlight this group completely.

michaelawyu: Besides, I have some doubt that kubelet will not touch APIs like deployment, daemon sets; they are very high-level and not of the concern for node-related components.

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -61,21 +69,42 @@ func (v *fleetResourceValidator) Handle(ctx context.Context, req admission.Reque
 		case crdGVK:
 			klog.V(2).InfoS("handling CRD resource", "GVK", crdGVK, "namespacedName", namespacedName, "operation", req.Operation)
 			response = v.handleCRD(req)
-		case mcGVK:
-			klog.V(2).InfoS("handling Member cluster resource", "GVK", mcGVK, "namespacedName", namespacedName, "operation", req.Operation)
-			response = v.handleMemberCluster(req)
+		case configMapGVK:
```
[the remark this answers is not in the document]
michaelawyu: There are ~100 objects available right now 😣

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -55,30 +61,39 @@ type fleetResourceValidator struct {
...
-			response = v.handleRoleBinding(req)
-		case namespaceGVK:
-			klog.V(2).InfoS("handling namespace resource", "GVK", namespaceGVK, "namespacedName", namespacedName, "operation", req.Operation)
-			response = v.handleNamespace(req)
-		default:
-			klog.V(2).InfoS("resource is not monitored by fleet resource validator webhook", "GVK", req.Kind.String(), "namespacedName", namespacedName, "operation", req.Operation)
-			response = admission.Allowed(fmt.Sprintf("user: %s in groups: %v is allowed to modify resource with GVK: %s", req.UserInfo.Username, req.UserInfo.Groups, req.Kind.String()))
+		msgFormat := fmt.Sprintf("handling %s resource", req.Kind.Kind)
+		// req.Name and req.Namespace for namespace objects are populated and the same for kind:Namespace objects, so setting it to empty.
+		if req.Kind.Kind == "Namespace" {
```
michaelawyu, line 69: Hi Arvind! Just to be sure that I understand the situation correctly: we do not want to allow any resource under reserved namespaces, right?

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -55,30 +61,39 @@ type fleetResourceValidator struct {
...
+			req.Namespace = ""
```
michaelawyu, line 70: Hi Arvind! It's just a readability concern, I feel that better clarity can be achieved by refactoring this part a little bit. The same logic can be achieved by:

michaelawyu:
```
switch {
    case ... // Special GVKs, e.g., CRD, MemberCluster, Namespace => need special/additional handling
    case req.Namespace == "" // We know for sure that at this point it's guaranteed not to be a Namespace resource
                                                // If the req.Namespace field is empty, it's a cluster scoped resource, we do not need
                                                // to process it (if I understand it correctly).
    default (or another case):    // Namespace scoped resources, need to make sure that it's not in the reserved
                                                // namespace, unless the request is from a whitelisted source
}
```

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
michaelawyu: The thing is, we can use generic matcher if that's the case.

test/e2e/webhook_test.go
```diff
@@ -793,15 +824,46 @@ var _ = Describe("Fleet's CR Resource Handler webhook tests", Ordered, func() {
...
+				},
+				Rules: []rbacv1.PolicyRule{
+					{
+						Verbs:     []string{"*"},
+						APIGroups: []string{"*"},
+						Resources: []string{"*"},
+					},
+				},
+			}
+			Eventually(func() error {
```
michaelawyu, line 849: Hi Arvind! A simple Expect() clause should suffice here I suppose.

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -31,15 +34,18 @@ const (
...
-	imcGVK         = metav1.GroupVersionKind{Group: fleetv1alpha1.GroupVersion.Group, Version: fleetv1alpha1.GroupVersion.Version, Kind: "InternalMemberCluster"}
-	roleGVK        = metav1.GroupVersionKind{Group: rbacv1.SchemeGroupVersion.Group, Version: rbacv1.SchemeGroupVersion.Version, Kind: "Role"}
-	roleBindingGVK = metav1.GroupVersionKind{Group: rbacv1.SchemeGroupVersion.Group, Version: rbacv1.SchemeGroupVersion.Version, Kind: "RoleBinding"}
-	namespaceGVK   = metav1.GroupVersionKind{Group: corev1.SchemeGroupVersion.Group, Version: corev1.SchemeGroupVersion.Version, Kind: "Namespace"}
+	crdGVK       = metav1.GroupVersionKind{Group: v1.SchemeGroupVersion.Group, Version: v1.SchemeGroupVersion.Version, Kind: "CustomResourceDefinition"}
+	mcGVK        = metav1.GroupVersionKind{Group: fleetv1alpha1.GroupVersion.Group, Version: fleetv1alpha1.GroupVersion.Version, Kind: "MemberCluster"}
+	namespaceGVK = metav1.GroupVersionKind{Group: corev1.SchemeGroupVersion.Group, Version: corev1.SchemeGroupVersion.Version, Kind: "Namespace"}
+
+	appsV1GV        = schema.GroupVersion{Group: appsv1.SchemeGroupVersion.Group, Version: appsv1.SchemeGroupVersion.Version}
```
michaelawyu, line 41: Hi Arvind! We need to target all API groups if I understand it correctly.

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -159,6 +145,29 @@ func (v *fleetResourceValidator) handleNamespace(req admission.Request) admissio
...
+func (v *fleetResourceValidator) handleNamespacedResources(ctx context.Context, req admission.Request) admission.Response {
+	resourceMsgFormat := fmt.Sprintf("handling %s resource", req.Kind.Kind)
+	reqGroupVersion := schema.GroupVersion{Group: req.Resource.Group, Version: req.Resource.Version}
+	namespacedName := types.NamespacedName{Name: req.Name, Namespace: req.Namespace}
+	var response admission.Response
+	if reqGroupVersion == appsV1GV || reqGroupVersion == coreV1GV || reqGroupVersion == rbacV1GV || reqGroupVersion == batchV1GV {
+		klog.V(2).InfoS(resourceMsgFormat, "GVK", req.Kind, "namespacedName", namespacedName, "operation", req.Operation)
+		response = v.handleResource(req)
+	} else if reqGroupVersion == fleetV1Alpha1GV {
+		if req.Kind.Kind == "InternalMemberCluster" {
```
michaelawyu, line 163: Hi Arvind! Is there any particular reason why `InternalMemberCluster` is singled out here? We also have a number of other v1alpha1 resources that needs guarding I believe.

pkg/webhook/webhook.go
```diff
@@ -494,6 +516,15 @@ func bindWebhookConfigToFleetSystem(ctx context.Context, k8Client client.Client,
 	return nil
 }
 
+func getGeneralRulesForNamespacedResources(cudOperations []admv1.OperationType, scope admv1.ScopeType) []admv1.RuleWithOperations {
```
michaelawyu, line 519: Hi Arvind! A nit here: with generic matchers, is this method still needed? Also, since the method is already for namespaced resources, there's no need to pass in scope I assume?

michaelawyu:
Hi Arvind! Would you mind giving a try running the test suite in parallel (`ginkgo -p`)? I am afraid that many of the test cases would fail because of the setup would be in conflict when using multiple nodes. For now it's not much of an issue since we haven't enabled the node setup, but in the long run it could be damaging.

michaelawyu:
For now let's leave it as is as we never enabled it. We could do more tweaking when the situation comes.

michaelawyu (approved):

PR #503, opened 2023-08-24
fix: delete associated works when bindings are deleted

michaelawyu:

This PR fixes an issue where associated works are not deleted when the scheduler removes bindings.

I have:

N/A

### DIFF
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -160,6 +160,16 @@ func (r *Reconciler) handleDelete(ctx context.Context, resourceBinding *fleetv1b
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	// delete all the listed works
+	for workName := range works {
+		work := works[workName]
+
+		if err := r.Client.Delete(ctx, work); err != nil {
+			return ctrl.Result{}, controller.NewAPIServerError(false, err)
+		}
+	}
+
 	// remove the work finalizer on the binding if all the work objects are deleted
 	if len(works) == 0 {
 		controllerutil.RemoveFinalizer(resourceBinding, fleetv1beta1.WorkFinalizer)
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -162,10 +162,13 @@ func (r *Reconciler) handleDelete(ctx context.Context, resourceBinding *fleetv1b
 	}
 
 	// delete all the listed works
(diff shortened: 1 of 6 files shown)


michaelawyu:
Closing this PR for now as the root cause seems to be elsewhere.

michaelawyu:
This PR is sent as, when writing the E2E tests, it occurs that the system will not delete works derived from deleted bindings; logically it should be handled by the work generator (which creates the works). Though, when refreshing the tests it seems that the controller expects some kind of GC controller to do the work instead, which does not seem to exist at this moment. If the GC controller does exist, please let me know, and I would fix the issue differently.

pkg/controllers/workgenerator/controller.go
```diff
@@ -160,6 +160,16 @@ func (r *Reconciler) handleDelete(ctx context.Context, resourceBinding *fleetv1b
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	// delete all the listed works
+	for workName := range works {
+		work := works[workName]
+
+		if err := r.Client.Delete(ctx, work); err != nil {
```
zhiying-lin, line 168: ignore the notfound error?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -677,20 +677,30 @@ var _ = Describe("Test clusterSchedulingPolicySnapshot Controller", func() {
...
-					return k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)
-				}, duration, interval).Should(Succeed(), "the binding should not be removed until all the work is removed")
-				// delete the work (as the testEnv doesn't run GC controller)
-				Expect(k8sClient.Delete(ctx, &work)).Should(Succeed())
-				By(fmt.Sprintf("work  %s is deleted", work.Name))
-				// Check the binding is not still removed
-				Consistently(func() error {
-					return k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)
-				}, duration, interval).Should(Succeed(), "the binding should not be removed until all the work is removed")
-				// delete the work2 (as the testEnv doesn't run GC controller)
```
zhiying-lin, line 691: @ryanzhang-oss what's the GC controller? i cannot recall.

pkg/controllers/workgenerator/controller.go
michaelawyu: Done.

zhiying-lin (approved):

PR #505, opened 2023-08-25
feat: declare v1beta1 work related const in the api folder

zhiying-lin:

- declare the work related const in the api folder, so that
    - v1beta1 controllers won't need to import work controller package any more.
    - the annotations can use shared fleet prefix.

Fixes #

I have:

### DIFF
--- a/apis/placement/v1beta1/work_types.go
+++ b/apis/placement/v1beta1/work_types.go
@@ -26,6 +26,20 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// The following definitions are declared in the manager.go file.
+const (
+	// ManifestHashAnnotation is the annotation that indicates whether the spec of the object has been changed or not.
+	ManifestHashAnnotation = fleetPrefix + "spec-hash"
+
+	// LastAppliedConfigAnnotation is to record the last applied configuration on the object.
+	LastAppliedConfigAnnotation = fleetPrefix + "last-applied-configuration"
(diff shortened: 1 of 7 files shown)


apis/placement/v1beta1/work_types.go
```diff
@@ -26,6 +26,20 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// The following definitions are declared in the manager.go file.
+const (
+	// ManifestHashAnnotation is the annotation that indicates whether the spec of the object has been changed or not.
+	ManifestHashAnnotation = fleetPrefix + "spec-hash"
```
michaelawyu, line 32: These two fields are added in this PR but not referenced at all?

michaelawyu: They are in use by the work controller I assume?

zhiying-lin: Will be used in the v1beta1 work controller in a separate PR.

michaelawyu (approved):
LGTM.

apis/placement/v1beta1/work_types.go
```diff
@@ -26,6 +26,20 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// The following definitions are declared in the manager.go file.
```
ryanzhang-oss, line 29: what does this comment mean?

PR #509, opened 2023-08-30
fix: address a minor issue in hub agent chart + enable API version params in member agent chart

michaelawyu:

This PR fixes a minor issue in hub agent chart + enables API version params in member agent chart.

I have:

N/A

### DIFF
--- a/charts/hub-agent/templates/deployment.yaml
+++ b/charts/hub-agent/templates/deployment.yaml
@@ -27,12 +27,8 @@ spec:
             - --webhook-client-connection-type={{.Values.webhookClientConnectionType}}
             - --v={{ .Values.logVerbosity }}
             - -add_dir_header
-            {{- if .Values.enableV1Alpha1APIs }}
-            - --enable-v1alpha1-apis
-            {{- end }}
-            {{- if .Values.enableV1Beta1APIs }}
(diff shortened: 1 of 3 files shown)


charts/hub-agent/templates/deployment.yaml
```diff
@@ -27,12 +27,8 @@ spec:
             - --webhook-client-connection-type={{.Values.webhookClientConnectionType}}
             - --v={{ .Values.logVerbosity }}
             - -add_dir_header
-            {{- if .Values.enableV1Alpha1APIs }}
-            - --enable-v1alpha1-apis
-            {{- end }}
-            {{- if .Values.enableV1Beta1APIs }}
```
zhiying-lin, line 33: hi Chen, just wonder why the original one is not correct, thanks

zhiying-lin (approved):

charts/hub-agent/templates/deployment.yaml
michaelawyu: Hi Zhiying! The args have default values (`true` for v1alpha1 and `false` for v1beta1); in the old setup there is no way to disable v1alpha1 and enable only v1beta1, even if we set `enableV1Alpha1APIs` to `false`.

PR #519, opened 2023-09-06
feat: add formalized log in order to setup the dashboard

zhiying-lin:
So that we can create the monitoring based on these logs

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterresourceplacement/resource_selector.go
+++ b/pkg/controllers/clusterresourceplacement/resource_selector.go
@@ -124,7 +124,7 @@ func (r *Reconciler) fetchClusterScopedResources(selector fleetv1beta1.ClusterRe
 	}
 	restMapping, err := r.RestMapper.RESTMapping(gk, selector.Version)
 	if err != nil {
-		return nil, controller.NewUserError(fmt.Errorf("failed to get GVR of the selector: %w", err))
+		return nil, controller.NewUserError(fmt.Errorf("invalid placement %s, failed to get GVR of the selector: %w", placeName, err))
 	}
 	gvr := restMapping.Resource
 	gvk := schema.GroupVersionKind{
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -124,7 +124,7 @@ func (r *Reconciler) fetchClusterScopedResources(selector fleetv1beta1.ClusterRe
 	}
 	restMapping, err := r.RestMapper.RESTMapping(gk, selector.Version)
 	if err != nil {
-		return nil, controller.NewUserError(fmt.Errorf("failed to get GVR of the selector: %w", err))
+		return nil, controller.NewUserError(fmt.Errorf("invalid placement %s, failed to get GVR of the selector: %w", placeName, err))
```
ryanzhang-oss, line 127: is the "format" defined anywhere?

ryanzhang-oss: is that every placement error starts with "invalid placement"?

zhiying-lin: today it's not. assuming each error should contain the placement name. Are you suggesting to have this format so that we can query the crp error log?

ryanzhang-oss: oh, does the "formalized" here mean having the placement name? Are those the only errors that don't have the placement name?

Not necessarily for this PR. Do we want to define a format/regex for our error message so that we can search later?

ryanzhang-oss (approved):

pkg/controllers/clusterresourceplacement/resource_selector.go
zhiying-lin: >does the "formalized" here mean having the placement name?

The PR is to define a fixed prefix for all of the controller errors in the controller package.

>Are those the only errors that don't have the placement name?
i have not reviewed all of the code. Most of the new code should have the CRP name in the error string, not sure if old code does or not. Will do later.

>Do we want to define a format/regex for our error message so that we can search later?

controller common error logs the fixed the prefix, which should help us to query.

PR #520, opened 2023-09-08
fix: add scope in the log when fail to get token

zhiying-lin:

Need to log the scope for debug purpose

Fixes #

I have:


ryanzhang-oss (approved):
the token code might need to be rewritten in the future.

PR #529, opened 2023-09-15
fix: address an issue where rollout controller will panic when processing CRPs of the PickFixed placement type

michaelawyu:

This PR fixes an issue where rollout controller will panic when processing CRPs of the PickFixed placement type, as such CRPs do not have numOfClusters field set.

I have:

- [x] Unit tests
- [x] Integration tests
- [x] E2E tests (upstream)


ryanzhang-oss (approved):
There might be other places that may have this issue since the fixed type is added after many of the code is already written.

PR #534, opened 2023-09-20
fix: set the resource change processing controller to handle v1beta1 CRPs correctly + ignore member cluster updates

michaelawyu:

This PR fixes an issue where the resource change processing controller only processes v1alpha1 CRPs, which leads to unreasonably long delays for fleets with v1beta1 API to apply resource changes.

I have:

- [x] Unit tests
- [x] E2E tests (upstream)

Expedient fixes, which should get refactored in the future when v1alpha1 API becomes obsolete.

This also edits the E2E tests we have right now a little bit to verify the effectiveness.

### DIFF
--- a/pkg/controllers/resourcechange/resourcechange_controller.go
+++ b/pkg/controllers/resourcechange/resourcechange_controller.go
@@ -136,6 +136,7 @@ func (r *Reconciler) findPlacementsSelectedDeletedResV1Alpha1(res keys.ClusterWi
 
 	if len(matchedCrps) == 0 {
 		klog.V(2).InfoS("change in deleted object does not affect any v1alpha1 placement", "obj", res)
+		return
 	}
 
 	for _, crp := range matchedCrps {
@@ -171,6 +172,7 @@ func (r *Reconciler) findPlacementsSelectedDeletedResV1Beta1(res keys.ClusterWid
 
 	if len(matchedCrps) == 0 {
 		klog.V(2).InfoS("change in deleted object does not affect any v1beta1 placement", "obj", res)
+		return
 	}
 
 	for _, crp := range matchedCrps {
--- a/pkg/controllers/resourcechange/resourcechange_controller.go
+++ b/pkg/controllers/resourcechange/resourcechange_controller.go
@@ -23,6 +23,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	fleetv1alpha1 "go.goms.io/fleet/apis/v1alpha1"
 	"go.goms.io/fleet/pkg/utils"
 	"go.goms.io/fleet/pkg/utils/controller"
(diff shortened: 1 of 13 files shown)


pkg/controllers/resourcechange/resourcechange_controller.go
```diff
@@ -90,42 +91,37 @@ func (r *Reconciler) Reconcile(_ context.Context, key controller.QueueKey) (ctrl
...
-			r.PlacementControllerV1Alpha1.Enqueue(crp)
-		}
-		if r.PlacementControllerV1Beta1 != nil {
-			klog.V(2).InfoS("Change in object triggered v1beta1 placement reconcile", "obj", clusterWideKey, "crp", crp)
-			r.PlacementControllerV1Beta1.Enqueue(crp)
-		}
-	}
 
-	return ctrl.Result{}, nil
+	return r.triggerAffectedPlacementsForUpdatedClusterRes(clusterWideKey, clusterObj.(*unstructured.Unstructured))
```
ryanzhang-oss, line 95: unfortunately looks like there are a lot of very similar code except type change.

test/e2e/placement_test.go
```diff
@@ -34,14 +38,43 @@ var _ = Describe("placing resources using a CRP with no placement policy specifi
...
+
+		configMap.Data = map[string]string{
+			"data": "updated",
+		}
+		Expect(hubClient.Update(ctx, configMap)).To(Succeed(), "Failed to update config map")
+	})
+
+	It("should place the resources on all member clusters", func() {
+		for idx := range allMemberClusters {
+			workResourcesPlacedActual := workNamespaceAndConfigMapPlacedOnClusterActual(allMemberClusters[idx])
```
ryanzhang-oss, line 77: does this check if the config data is changed?

pkg/controllers/resourcechange/resourcechange_controller.go
```diff
@@ -137,21 +133,50 @@ func (r *Reconciler) findPlacementsSelectedDeletedRes(res keys.ClusterWideKey, c
 			}
 		}
 	}
+
 	if len(matchedCrps) == 0 {
-		klog.V(2).InfoS("change in deleted object does not affect any placement", "obj", res)
-		return ctrl.Result{}, nil
+		klog.V(2).InfoS("change in deleted object does not affect any v1alpha1 placement", "obj", res)
```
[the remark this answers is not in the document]
michaelawyu: Done.

test/e2e/placement_test.go
michaelawyu: Hi Ryan! Yeah, this one retrieves the data from the hub cluster as the baseline, so it already includes the change.

pkg/controllers/resourcechange/resourcechange_controller.go
michaelawyu: Originally the query returns the CRPs needed to be reconciled, and the outer level enqueues them. However, the fix requires us to have a branching point (v1alpha1 vs v1beta1 -> we also have factor in the cases where both versions are enabled, though I guess that's not a supported scenario any more), and it might be a little bit cumbersome to distinguish between the different versions of CRPs on the outer level, hence the type change (i.e., it returns the result directly).

ryanzhang-oss (approved):

PR #535, opened 2023-09-21
fix: watch for observed CRP annotation change in scheduler watcher

michaelawyu:

This PR fixes an issue where the scheduler watcher ignores observed CRP annotation changes.

I have:

- [x] Integration tests
- [x] E2E tests (upstream)

### DIFF
--- a/pkg/scheduler/watchers/clusterschedulingpolicysnapshot/controller.go
+++ b/pkg/scheduler/watchers/clusterschedulingpolicysnapshot/controller.go
@@ -124,7 +124,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 			}
 
 			// Policy snapshot spec is immutable; however, the scheduler will have to respond
-			// to changes in the annotations.
+			// to changes in the numberOfCluster & CRPGeneration annotations.
 			oldAnnotations := e.ObjectOld.GetAnnotations()
 			newAnnotations := e.ObjectNew.GetAnnotations()
 
--- a/pkg/scheduler/watchers/clusterschedulingpolicysnapshot/controller.go
+++ b/pkg/scheduler/watchers/clusterschedulingpolicysnapshot/controller.go
@@ -124,13 +124,19 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 			}
 
 			// Policy snapshot spec is immutable; however, the scheduler will have to respond
-			// to changes in the number of clusters annoation.
+			// to changes in the annoations.
 			oldAnnotations := e.ObjectOld.GetAnnotations()
 			newAnnotations := e.ObjectNew.GetAnnotations()
 
 			oldNumOfClusters := oldAnnotations[fleetv1beta1.NumberOfClustersAnnotation]
 			newNumOfClusters := newAnnotations[fleetv1beta1.NumberOfClustersAnnotation]
 			if oldNumOfClusters != newNumOfClusters {
+				return true
+			}
+
+			oldObservedCRPGeneration := oldAnnotations[fleetv1beta1.CRPGenerationAnnotation]
+			newObservedCRPGeneration := newAnnotations[fleetv1beta1.CRPGenerationAnnotation]
+			if oldObservedCRPGeneration != newObservedCRPGeneration {
 				return true
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/scheduler/watchers/clusterschedulingpolicysnapshot/controller.go
```diff
@@ -124,7 +124,7 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 			}
 
 			// Policy snapshot spec is immutable; however, the scheduler will have to respond
-			// to changes in the number of clusters annoation.
+			// to changes in the annotations.
```
ryanzhang-oss, line 127: nit: not all annotation change matters

pkg/scheduler/watchers/clusterschedulingpolicysnapshot/controller.go
```diff
@@ -134,6 +134,14 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 				return true
 			}
 
+			// The scheduler needs to update the policy snapshot based on the latest CRP generation, when resource selector
+			// has changed and there are no policy changes.
```
ryanzhang-oss, line 138: how will the scheduler know that it should do another scheduling in this case? Wouldn't it just go ahead and reschedule everything?

zhiying-lin: as the policy has not changed, 
- for the fixed scheduling type, the scheduling result will be the same. 
- for the select N, the scheduler will try to fullfill the "number of cluster", otherwise, it will take no action.

will test it to prove my assumption.

ryanzhang-oss: why not update the policysnapshot status from this controller directly?

ryanzhang-oss: what about select ALL?

zhiying-lin: no difference between select all & select fixed when handling this. If there are no member cluster changes, the result are always the same.

`why not update the policysnapshot status from this controller directly?` 
we could, but it's more like a hacky way. Also, today we have a lot of events could trigger the scheduler. And scheduler should guarantee that it should not reschedule everything.

zhiying-lin: I verified in my local and it's working as expected.

The test step is
1) create CRP using this yaml file to select all the clusters.

```
apiVersion: placement.kubernetes-fleet.io/v1beta1
kind: ClusterResourcePlacement
metadata:
  creationTimestamp: "2023-09-21T08:24:30Z"
  finalizers:
  - custom-deletion-blocker-finalizer
  - kubernetes-fleet.io/crp-cleanup
  - kubernetes-fleet.io/scheduler-cleanup
  generation: 1
  name: crp-1
  resourceVersion: "861"
  uid: 5c00403d-7852-4ec4-8da8-6edf7f6c6023
spec:
  policy:
    affinity:
      clusterAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          clusterSelectorTerms:
          - labelSelector:
              matchLabels:
                region: us-west
    numberOfClusters: 3
    placementType: PickAll
  resourceSelectors:
  - group: ""
    kind: Namespace
    name: work-1
    version: v1
  revisionHistoryLimit: 10
  strategy:
    type: RollingUpdate
```
2) remove the label from member-cluster-3, so that it won't be eligible any more.
3) update the resource selector

The scheduler sticked to the original decisions. The resources are placed on member-cluster-3.

zhiying-lin: read the code, https://github.com/Azure/fleet/blob/main/pkg/scheduler/framework/frameworkutils.go#L105 crossReferencePickedCustersAndObsoleteBindings will not delete any clusters when it becomes ineligible,  if the policy is still active.

ryanzhang-oss (approved):

zhiying-lin:
`so when the oldErr != nil, we already know newErr ==nil and newIsLatest is valid`

not really, if oldErr != nil && newErr == nil, the newIsLatest could be set as false. so we still need to return "newIsLatest". No difference compared with current code.

PR #537, opened 2023-09-22
test: add more tests about selecting resources

zhiying-lin:

- When matching cluster-scoped resources are updated (appearance/updated/disappearance)
     - should place and remove resources from clusters accordingly
     - should update CRP status accordingly
- When namespace-scoped resources in a matching namespace are updated (appearance/updated/disappearance)
     - should place and remove resources from clusters accordingly
     - should update CRP status accordingly

Fixes #

I have:


test/e2e/setup_test.go
```diff
@@ -88,6 +89,13 @@ var (
 	ignoreConditionLTTReasonAndMessageFields = cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime", "Reason", "Message")
 	ignoreAgentStatusHeartbeatField          = cmpopts.IgnoreFields(clusterv1beta1.AgentStatus{}, "LastReceivedHeartbeat")
 	ignoreNamespaceStatusField               = cmpopts.IgnoreFields(corev1.Namespace{}, "Status")
+
+	crpStatusCmpOptions = cmp.Options{
+		cmpopts.SortSlices(lessFuncCondition),
```
ryanzhang-oss, line 94: where were these function used before? wouldn't golang complain about it if there is not used before?

test/e2e/placement_selecting_resources_test.go
```diff
@@ -138,3 +142,453 @@ var _ = Describe("creating CRP and selecting resources by label", Ordered, func(
 		Eventually(finalizerRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove controller finalizers from CRP %s", crpName)
 	})
 })
+
+var _ = Describe("validating CRP when cluster-scoped resources become selected after the updates", Ordered, func() {
```
ryanzhang-oss, line 146: There are a lot of similarities between those cases. Would it make sense to share some code between them?

ryanzhang-oss: i.e. put the delete to afterAll and share one afterAll.

ryanzhang-oss: Also, did we place any cluster scoped resource (not namespace)?

ryanzhang-oss, line 146: will we see some failed cases (like failed to appy CR)?

test/e2e/setup_test.go
zhiying-lin: they're used in https://github.com/Azure/fleet/blob/main/test/e2e/actuals_test.go#L105.

The IDE will give the hints.

test/e2e/placement_selecting_resources_test.go
zhiying-lin: will do in next PR.

zhiying-lin: I tried this structure,

var _ = Describe(..., func (){
         BeforeAll()
         AfterAll()
         var _ = Describe("creating CRP and selecting resources by label", Ordered, func() {
                 It()
                It()
         })
})

However, ginkgo does not allow to put another layer on the top of the "ordered" spec. :( Have already created the cleanupxx related func to share the code.

zhiying-lin: ` did we place any cluster scoped resource (not namespace)?`

not yet, could add another test case in the next PR

ryanzhang-oss (approved):

PR #538, opened 2023-09-23
fix: consolidate unscheduled bindings.

ryanzhang-oss:
Currently, the unscheduled bindings are completely ignored by the framework. Unfortunately, this can lead to multiple bindings that point to the same cluster. I think it's better for the framework to deduplicate the bindings instead of asking rollout controller to handle it as rollout controller should not create or delete any bindings. The rollout controller still needs to handle the case that the previous binding is deleting while the cluster is selected again. 

Fixes #

I have:


pkg/scheduler/framework/frameworkutils.go
```diff
@@ -203,6 +198,33 @@ func crossReferencePickedCustersAndObsoleteBindings(
 	return toCreate, toDelete, toPatch, nil
 }
 
+func patchBinding(binding *placementv1beta1.ClusterResourceBinding, scored *ScoredCluster, policy *placementv1beta1.ClusterSchedulingPolicySnapshot) *bindingWithPatch {
+	// Update the binding so that it is associated with the latest score.
+	updated := binding.DeepCopy()
+	affinityScore := int32(scored.Score.AffinityScore)
+	topologySpreadScore := int32(scored.Score.TopologySpreadScore)
+	// set the state as it might be unscheduled before
+	updated.Spec.State = placementv1beta1.BindingStateScheduled
```
michaelawyu, line 207: Hi Ryan! This is not necessarily the case; obsolete binding can also be in the Bound state.

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -135,30 +139,21 @@ func crossReferencePickedCustersAndObsoleteBindings(
...
-		topologySpreadScore := int32(scored.Score.TopologySpreadScore)
-		updated.Spec.ClusterDecision = placementv1beta1.ClusterDecision{
-			ClusterName: scored.Cluster.Name,
-			Selected:    true,
-			ClusterScore: &placementv1beta1.ClusterScore{
-				AffinityScore:       &affinityScore,
-				TopologySpreadScore: &topologySpreadScore,
-			},
-			Reason: pickedByPolicyReason,
+	for _, binding := range unscheduled {
```
michaelawyu, line 146: For this one, originally we agreed that unscheduled is a terminal state for bindings; we do not bring it back to life as it risks contention.

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -48,14 +50,15 @@ func classifyBindings(policy *placementv1beta1.ClusterSchedulingPolicySnapshot,
 		targetCluster, isTargetClusterPresent := clusterMap[binding.Spec.TargetCluster]
 
 		switch {
-		case binding.DeletionTimestamp != nil:
+		case !binding.DeletionTimestamp.IsZero():
```
michaelawyu, line 53: Hi Ryan! This will capture all bindings that haven't been deleted I believe?

pkg/scheduler/framework/frameworkutils.go
michaelawyu: One of the tricky points is that, a binding can reach the unscheduled states from two other states: scheduled, and bound. If we revive an unscheduled binding, we cannot reliably tell which state to revive it back to at this point.

michaelawyu: It may involve some minor side effects, mostly related to rolling updates (if applicable).

michaelawyu:
Hi Ryan! For this PR, concerning the comments above, I am wondering if it is better to, instead of bringing unscheduled bindings back, put some clusters on hold until its associated unscheduled binding is away -> this could make sure that at any time one cluster can at at most 1 binding associated, at the cost of increased delay at certain circumstances.

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -135,30 +139,32 @@ func crossReferencePickedCustersAndObsoleteBindings(
...
+		// The binding's target cluster is picked again in the current run; yet the binding
+		// is originally de-selected by the previous scheduling round.
+		// Add the binding to the toPatch list so that we won't create more and more bindings.
+		// We need to recover the previous state before the binding is marked as unscheduled.
+		var desiredState placementv1beta1.BindingState
+		// we recorded the previous state in the binding's annotation
+		if previousState, exist := binding.GetAnnotations()[placementv1beta1.PreviousBindingStateAnnotation]; exist {
+			desiredState = placementv1beta1.BindingState(previousState)
+			// remove the annotation just to avoid confusion.
+			binding.SetAnnotations(nil)
```
michaelawyu, line 163: Hi Ryan! Just to be on the safe side, maybe it's better to just unset the one annotation we use for keeping the old state?

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -135,30 +139,32 @@ func crossReferencePickedCustersAndObsoleteBindings(
 
 		// The binding's target cluster is picked again in the current run; yet the binding
 		// is originally created/updated in accordance with an out-of-date scheduling policy.
+		// Add the binding to the toPatch list. We will simply keep the binding's state as
+		// it could be "scheduled" or "bound".
+		toPatch = append(toPatch, patchBinding(binding, binding.Spec.State, scored, policy))
```
michaelawyu, line 144: Hi Ryan! A concern is that, since we are using patch rather than update, this might overwrite a binding from bound back to scheduled under unfortunate circumstances (scheduler sees scheduled, but rollout has marked it as bound)

michaelawyu (dismissed):
LGTM ;)

michaelawyu:
Hi Ryan! There seems to be some linting errors though

pkg/scheduler/framework/framework.go
```diff
@@ -640,21 +664,19 @@ func (f *framework) createBindings(ctx context.Context, toCreate []*placementv1b
 		// TO-DO (chenyu1): Add some jitters here to avoid swarming the API when there is a large number of
 		// bindings to create.
 		if err := f.client.Create(ctx, binding); err != nil {
-			return controller.NewAPIServerError(false, fmt.Errorf("failed to create binding %s: %w", binding.Name, err))
+			return controller.NewCreateIgnoreAlreadyExistError(fmt.Errorf("failed to create binding %s: %w", binding.Name, err))
```
michaelawyu, line 667: Hi Ryan! Just a nit: though it may not happen in most of the cases, I am afraid that AlreadyExistsError cannot be ignored for this one, as the name is generated randomly each time

michaelawyu (approved):
LGTM ;)

PR #539, opened 2023-09-25
test: add more scheduler integration tests (2/)

michaelawyu:

This PR features some more integration tests for the scheduler.

It also includes some minor fixes for the prev. PR that was not rebased into the branch in time.

I have:

N/A


test/scheduler/actuals_test.go
```diff
@@ -371,3 +385,52 @@ func pickFixedPolicySnapshotStatusUpdatedActual(valid, invalidOrNotFound []strin
...
 	}
 }
+
+func pickAllPolicySnapshotStatusUpdatedActual(scored, filtered []string, policySnapshotName string) func() error {
+	return func() error {
+		policySnapshot := &placementv1beta1.ClusterSchedulingPolicySnapshot{}
+		if err := hubClient.Get(ctx, types.NamespacedName{Name: policySnapshotName}, policySnapshot); err != nil {
+			return err
+		}
+
```
zhiying-lin, line 395: compare the full struct of the policy status?

`If your function returns a struct, don't write test code that performs an individual comparison for each field of the struct. Instead, construct the struct that you're expecting your function to return, and compare in one shot using [diffs](https://github.com/golang/go/wiki/TestComments#print-diffs) or [deep comparisons](https://github.com/golang/go/wiki/TestComments#equality-comparison-and-diffs). The same rule applies to arrays and maps.
`
https://github.com/golang/go/wiki/TestComments#compare-full-structures

test/scheduler/utils_test.go
```diff
@@ -248,11 +273,31 @@ func ensureCRPDeletion(crpName string) {
...
 }
+
+func ensureProvisionalClusterDeletion(clusterName string) {
+	// Retrieve the provisional cluster.
+	memberCluster := &clusterv1beta1.MemberCluster{}
+	Expect(hubClient.Get(ctx, types.NamespacedName{Name: clusterName}, memberCluster)).To(Succeed(), "Failed to get member cluster")
+
+	// Remove all finalizers from the provisional cluster.
+	memberCluster.Finalizers = []string{}
+	Expect(hubClient.Update(ctx, memberCluster)).To(Succeed(), "Failed to update member cluster")
```
zhiying-lin, line 292: just wonder who will add the finalizer on membercluster

test/scheduler/pickall_integration_test.go
```diff
@@ -0,0 +1,216 @@
...
+						Reason:             dummyReason,
+					},
+				},
+				LastReceivedHeartbeat: metav1.NewTime(time.Now()),
+			},
+		}
+		Expect(hubClient.Status().Update(ctx, memberCluster)).To(Succeed(), "Failed to update member cluster status")
+	})
+
+	It("should not remove binding for the cluster that just becomes unhealthy", func() {
```
zhiying-lin, line 179: will you cover the 1) membercluster label is changed 2) membercluster is left in the next PR?

test/scheduler/pickall_integration_test.go
```diff
@@ -0,0 +1,216 @@
...
+	. "github.com/onsi/gomega"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
+	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+)
+
+var _ = Describe("scheduling CRPs with no scheduling policy specified", Ordered, func() {
```
ryanzhang-oss, line 26: This is still Ordered which is making it very difficult to reason for each test. Please make those independent. We don't need parallel but definitely not Ordered.

michaelawyu: Hi Ryan! This just means that all the test cases in this block run sequentially in **this** group; I marked this to avoid having to write a huge IT block

test/scheduler/pickall_integration_test.go
michaelawyu: As mentioned in the PR description, it's only part of the test.

test/scheduler/actuals_test.go
michaelawyu: Again and again, any Go guideline is just a general guideline, not a must. I don't particularly have a strong opinion on this one, but the check here is done separately to a) avoid having a large bloc of data and b) avoid stuffing the Diff() call with tons of options, which is even more difficult to reason about.

zhiying-lin: you can always construct the options using,

snapshotStatusCmpOptions = cmp.Options{
		ignoreClusterDecisionReasonField,
                cmpopts.SortSlices(lessFuncClusterDecision),
               ...
	}

and build the wantStatus like

wantStatus := SchedulingPolicySnapshotStatus{
         	ObservedCRPGeneration: wantCRPGeneration
                Conditions: []metav1.Condition{
                          wantScheduledCondition,
                }
               ClusterDecisions: wantClusterDecisions,
}

Then compare the status in one Diff call, the code will be cleaner than comparing the field one by one.
The amount of the data is the same as the current one.

michaelawyu: Stuffing all the options into one struct/array does not make them disappear. And building this want struct with blocks IMO does not help much with readability.

test/scheduler/utils_test.go
michaelawyu: No longer applies after the refactor.

test/scheduler/pickall_integration_test.go
michaelawyu: Refactored per offline discussion.

ryanzhang-oss (approved):

PR #541, opened 2023-09-26
test: add more e2e test cases

zhiying-lin:

add test cases
- validating CRP when resource selector is not valid
- validating CRP when selecting a reserved resource
- validating CRP when failed to apply resources
- validating CRP when placing cluster scope resource (other than namespace)

Fixes #

I have:


test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +594,433 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
 		Eventually(finalizerRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove controller finalizers from CRP %s", crpName)
 	})
 })
+
+// TODO should be blocked by the validation webhook
```
michaelawyu, line 598: Just for clarification, does the validation webhook check for this?

michaelawyu: 'cause for this specific test case, i.e., non-existent kind, it would require the webhook to know which APIs are registered. It also excludes the experience where a CRP is created before the API becomes available (a very rare case though).

test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +594,433 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
+		Expect(hubClient.Create(ctx, crp)).To(Succeed(), "Failed to create CRP %s", crpName)
+	})
+
+	AfterAll(func() {
+		By(fmt.Sprintf("deleting placement %s", crpName))
+		cleanupCRP(crpName)
+	})
+
+	It("should update CRP status as expected", func() {
+		crpStatusUpdatedActual := func() error {
```
michaelawyu, line 710: A nit: the logic here seems to be reused in multiple places.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +594,433 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
+var _ = Describe("validating CRP when failed to apply resources", Ordered, func() {
+	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+
+	BeforeAll(func() {
+		By("creating work resources on hub cluster")
+		createWorkResources()
+
+		By("creating work namespace on member cluster")
+		ns := workNamespace()
+		memberCluster := allMemberClusters[0]
```
michaelawyu, line 758: A nit: there is a `MemberCluster1` variable available.

test/e2e/utils_test.go
```diff
@@ -155,22 +159,26 @@ func checkIfRemovedWorkResourcesFromAllMemberClusters() {
 // cleanupCRP deletes the CRP and waits until the resources are not found.
 func cleanupCRP(name string) {
 	// TODO(Arvindthiru): There is a conflict which requires the Eventually block, not sure of series of operations that leads to it yet.
-	Eventually(func(g Gomega) error {
+	Eventually(func() error {
```
michaelawyu, line 162: Note that this is actually the supported pattern in Ginkgo.

zhiying-lin: oh, the fix is to remove the g.Expect(err) in the Eventually, which looks not correct.

test/e2e/placement_selecting_resources_test.go
zhiying-lin: i'm thinking of blocking them at the very beginning as it's not that obvious for the customer to notice the failure during the runtime.

@ryanzhang-oss what do you think?

test/e2e/placement_selecting_resources_test.go
zhiying-lin: This func is used in two places. Will refactor later if it's used in the third place.

test/e2e/utils_test.go
michaelawyu:
```
Eventually can also accept functions that take a Gomega parameter. These functions are then passed a local Gomega that can be used to make assertions inside the function as it is polled. Eventually will retry the function if an assertion fails. This would allow us to replace:
```

https://onsi.github.io/ginkgo/

zhiying-lin: i see, good to know this pattern.

I think we need to either use the "local gomega + return void" or "no local gomega + return error".

The current code is mixed.
The example is using local gomega to replace the "return error".

test/e2e/placement_selecting_resources_test.go
ryanzhang-oss: IMO, the validation webhook, in general, should not go so deep into CRP controller's logic.  Currently, it checks if the resource selector is valid, but it doesn't check if the selector selects anything. The line is a bit grey though.  Currently, it doesn't check if the resource selector's GVK is valid for v1beta1. We check it for v1alpha1.

ryanzhang-oss: However, the webhook wouldn't check if the selector select anything as resources can be applied later.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +594,430 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
+
+	It("should update CRP status as expected", func() {
+		crpStatusUpdatedActual := func() error {
+			crp := &placementv1beta1.ClusterResourcePlacement{}
+			if err := hubClient.Get(ctx, types.NamespacedName{Name: crpName}, crp); err != nil {
+				return err
+			}
+
+			wantStatus := placementv1beta1.ClusterResourcePlacementStatus{
+				Conditions: []metav1.Condition{
```
ryanzhang-oss, line 639: what is the reason for the false condition? I think that's the important part of the UX.  User will be less confusing if we state clearly that the Kind of the selector is invalid.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +594,430 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
```
ryanzhang-oss, line 717: same question about the message/reason.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +594,430 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
+					},
+					{
+						Type:               string(placementv1beta1.ClusterResourcePlacementAppliedConditionType),
+						Status:             metav1.ConditionFalse,
+						ObservedGeneration: crp.Generation,
+					},
+				},
+				PlacementStatuses: []placementv1beta1.ResourcePlacementStatus{
+					{
+						ClusterName: memberCluster1Name,
```
ryanzhang-oss, line 817: I don't get why only this cluster apply failed from the diff only (didn't check it out in an IDE).

test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +594,430 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
+						FailedResourcePlacements: []placementv1beta1.FailedResourcePlacement{
+							{
+								ResourceIdentifier: placementv1beta1.ResourceIdentifier{
+									Kind:    "Namespace",
+									Name:    workNamespaceName,
+									Version: "v1",
+								},
+								Condition: metav1.Condition{
```
ryanzhang-oss, line 825: again, what's the reason?

test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +594,430 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
+						Status:             metav1.ConditionFalse,
+						ObservedGeneration: crp.Generation,
+					},
+				},
+				PlacementStatuses: []placementv1beta1.ResourcePlacementStatus{
+					{
+						ClusterName: memberCluster1Name,
+						FailedResourcePlacements: []placementv1beta1.FailedResourcePlacement{
+							{
+								ResourceIdentifier: placementv1beta1.ResourceIdentifier{
```
ryanzhang-oss, line 820: so the configMap is applied successfully but the namespace is not? The only thing I can think of is that the namespace was already there on this cluster. Is that really an error?

test/e2e/placement_selecting_resources_test.go
zhiying-lin: There are two cases:
1. the GVK is invalid.
2. the GVK is valid while the resource does not exist.

Current behavior:
1. For the case1, CRP controller sets the scheduled condition as false to return the error.
2. For the case2, CRP controller won't return the error and just select nothing.

For the case2, webhook should not block it, either.
For the case1, @ryanzhang-oss are you suggesting keeping it consistent with v1alpha1 (block by webhook)?

btw, will add a test for case2.

ryanzhang-oss: Yes, for case 1, v1beta1 webhook will try to block it (it doesn't have the absolute truth though).

test/e2e/placement_selecting_resources_test.go
ryanzhang-oss: I guess the reason is that namespace is already there?  not related to this PR but I think we should give UX an option to "adopt" namespace.

ryanzhang-oss (approved):

PR #542, opened 2023-09-26
feat: add wait for deleting binding in rollout

ryanzhang-oss:
Add a check in the rollout controller to see if there are any deleting binding co-exist with another binding on the same cluster. For now, we just wait for the deleting binding to disappear by itself since the resource deletion on the member cluster is supposed to be fast and this case is rare.

Fixes #

I have:

UT/IT


pkg/controllers/rollout/controller.go
```diff
@@ -160,6 +169,54 @@ func (r *Reconciler) fetchLatestResourceSnapshot(ctx context.Context, crpName st
...
+			}
+		}
+	}
+	// check if there are any cluster that has a binding that is both being deleted and scheduled
+	for cluster, binding := range bindingMap {
+		// check if there is a  deleting binding on the same cluster
+		if deletingBinding[cluster] {
+			klog.V(2).InfoS("Find a binding assigned to a cluster with another deleting binding", "clusterResourcePlacement", crpObj, "binding", binding)
+			if binding.Spec.State == fleetv1beta1.BindingStateBound {
+				// the rollout controller won't move a binding from unscheduled to bound if there is a deleting binding on the same cluster.
```
michaelawyu, line 199: Hi Ryan! A nit: the comment should be `... from scheduled to bound ...` I assume?

pkg/controllers/rollout/controller.go
```diff
@@ -160,6 +169,54 @@ func (r *Reconciler) fetchLatestResourceSnapshot(ctx context.Context, crpName st
...
+			}
+			if binding.Spec.State == fleetv1beta1.BindingStateUnscheduled {
+				// this is a very rare case that the resource was in the middle of being removed from a member cluster after it is unselected.
+				// then the cluster get selected and unselected in two scheduling before the member agent is able to clean up all the resources.
+				if binding.GetAnnotations()[fleetv1beta1.PreviousBindingStateAnnotation] == string(fleetv1beta1.BindingStateBound) {
+					// its previous state can not bound as rollout won't roll a binding with a deleting binding pointing to the same cluster.
+					return false, controller.NewUnexpectedBehaviorError(fmt.Errorf(
+						"find a cluster `%s` that has a unscheduled binding `%+s` with previous state is `bound` and a deleting binding point to it", binding.Spec.TargetCluster, binding.Name))
+				}
+				return true, nil
```
michaelawyu, line 211: Hi Ryan! In this case, if the rollout controller is set to wait, the unscheduled binding will never be deleted, am I understanding it correctly? The scheduler will not delete bindings

ryanzhang-oss: yes, the unscheduled binding will only be deleted after the deleting binding disappears. However, there is an optimization that we can apply now that we know the previous state of an unscheduled binding to delete an unscheduled binding directly if it was never bound. I will leave this for next PR.

ryanzhang-oss: How can the flow stuck though? The deleting binding will disappear on its own.

michaelawyu: Oh, right. I missed a part. It won't get stuck.

michaelawyu (dismissed):
LGTM ;)

michaelawyu (approved):
LGTM ;)

PR #547, opened 2023-09-28
test: add e2e test to validate the revision history

zhiying-lin:
Test
- When resource selectors are updated 
- When configured with a revision history limit

Fixes #

I have:

### DIFF
--- a/test/e2e/placement_selecting_resources_test.go
+++ b/test/e2e/placement_selecting_resources_test.go
@@ -17,7 +17,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/pointer"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	"go.goms.io/fleet/pkg/controllers/clusterresourceplacement"
@@ -1035,7 +1034,7 @@ var _ = Describe("validating CRP when placing cluster scope resource (other than
 	})
 })
 
-var _ = Describe("validating CRP revision history when updating resource selector", Ordered, func() {
+var _ = Describe("validating CRP revision history allowing single revision when updating resource selector", Ordered, func() {
 	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
 
 	BeforeAll(func() {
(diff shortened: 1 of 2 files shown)


test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +593,105 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
+
+		snapshotList := &placementv1beta1.ClusterSchedulingPolicySnapshotList{}
+		Expect(hubClient.List(ctx, snapshotList, matchingLabels)).Should(Succeed(), "Failed to list the policy revisions")
+		Expect(len(snapshotList.Items)).Should(Equal(1), "clusterSchedulingPolicySnapshotList got %v, want 1", len(snapshotList.Items))
+
+		resourceSnapshotList := &placementv1beta1.ClusterResourceSnapshotList{}
+		Expect(hubClient.List(ctx, resourceSnapshotList, matchingLabels)).Should(Succeed(), "Failed to list the resource revisions")
+		Expect(len(snapshotList.Items)).Should(Equal(1), "clusterResourceSnapshotList got %v, want 1", len(snapshotList.Items))
+	})
+
```
ryanzhang-oss, line 680: why not add the configMap and check that we will have 2 resource snapshot and 1 policy snapshot? Is this covered somewhere else?

zhiying-lin: sure, will add one.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -592,3 +592,190 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
...
+	It("should remove placed resources from all member clusters", checkIfRemovedWorkResourcesFromAllMemberClusters)
+
+	It("should remove controller finalizers from CRP", func() {
+		finalizerRemovedActual := allFinalizersExceptForCustomDeletionBlockerRemovedFromCRPActual()
+		Eventually(finalizerRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove controller finalizers from CRP %s", crpName)
+	})
+})
+
+var _ = Describe("validating CRP revision history allowing multiple revisions when updating resource selector", Ordered, func() {
```
ryanzhang-oss, line 690: I didn't get why this creates two resourceSnapshot while the one above creates one.  Is there any reason why this can not just be merged with the one above? They are like 95% the same.

zhiying-lin: The above CRP sets the "RevisionHistoryLimit: pointer.Int32(1),".

They could be merged, but the whole process will be long as the CRP will be updated multiple times to update the revision limit & create resourcesnapshosts.

I feel the current approach is much cleaner.

ryanzhang-oss (dismissed):

PR #551, opened 2023-10-04
feat: v1beta1 CRP Validation for PickFixed Placement Policy

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -8,6 +8,7 @@ package validator
 
 import (
 	"fmt"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiErrors "k8s.io/apimachinery/pkg/util/errors"
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -72,6 +72,7 @@ func ValidateClusterResourcePlacement(clusterResourcePlacement *placementv1beta1
 	if len(clusterResourcePlacement.Name) > validation.DNS1035LabelMaxLength {
 		allErr = append(allErr, fmt.Errorf("the name field cannot have length exceeding %d", validation.DNS1035LabelMaxLength))
 	}
+
 	for _, selector := range clusterResourcePlacement.Spec.ResourceSelectors {
 		//TODO: make sure the selector's gvk is valid
 		if selector.LabelSelector != nil {
@@ -84,8 +85,10 @@ func ValidateClusterResourcePlacement(clusterResourcePlacement *placementv1beta1
 		}
 	}
 
-	if err := validatePlacementPolicy(clusterResourcePlacement.Spec.Policy); err != nil {
-		allErr = append(allErr, fmt.Errorf("the placement policy field is invalid: %w", err))
+	if clusterResourcePlacement.Spec.Policy != nil {
+		if err := validatePlacementPolicy(clusterResourcePlacement.Spec.Policy); err != nil {
+			allErr = append(allErr, fmt.Errorf("the placement policy field is invalid: %w", err))
+		}
 	}
 
 	if clusterResourcePlacement.Spec.Policy != nil && clusterResourcePlacement.Spec.Policy.Affinity != nil &&
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -8,6 +8,7 @@ package validator
 
 import (
 	"fmt"
+	"reflect"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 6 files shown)


pkg/webhook/clusterresourceplacement/v1beta1_clusterresourceplacement_validating_webhook.go
```diff
@@ -37,6 +37,7 @@ func Add(mgr manager.Manager, _ []string) error {
 func (v *clusterResourcePlacementValidator) Handle(_ context.Context, req admission.Request) admission.Response {
 	var crp placementv1beta1.ClusterResourcePlacement
 	if req.Operation == admissionv1.Create || req.Operation == admissionv1.Update {
+		klog.V(2).InfoS("handling CRP", "operation", req.Operation, "user", req.UserInfo.Username, "group", req.UserInfo.Groups, "namespacedName", types.NamespacedName{Name: req.Name})
```
ryanzhang-oss, line 40: do we need user name here?

test/e2e/utils_test.go
```diff
@@ -161,12 +161,12 @@ func cleanupCRP(name string) {
...
 		}
-		Expect(err).Should(Succeed(), "Failed to get CRP %s", name)
+		g.Expect(err).Should(Succeed(), "Failed to get CRP %s", name)
 
 		// Delete the CRP (again, if applicable).
 		//
-		// This helps the AfterAll node to run successfully even if the steps above fail early.
-		Expect(hubClient.Delete(ctx, crp)).To(Succeed(), "Failed to delete CRP %s", name)
+		// This helps the After All node to run successfully even if the steps above fail early.
+		g.Expect(hubClient.Delete(ctx, crp)).To(Succeed(), "Failed to delete CRP %s", name)
```
ryanzhang-oss, line 169: What is the difference between this and before?

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -94,6 +99,56 @@ func ValidateClusterResourcePlacement(clusterResourcePlacement *placementv1beta1
...
+		}
+	}
+
+	return apiErrors.NewAggregate(allErr)
+}
+
+func validatePolicyForPickFixedPlacementType(policy *placementv1beta1.PlacementPolicy) error {
+	allErr := make([]error, 0)
+	if len(policy.ClusterNames) == 0 {
+		allErr = append(allErr, fmt.Errorf("cluster names cannot be empty for policy type %s", policy.PlacementType))
```
ryanzhang-oss, line 125: nit: this has to be placementv1beta1.PickFixedPlacementType

pkg/utils/validator/clusterresourceplacement_test.go
```diff
@@ -397,6 +398,150 @@ func Test_validateClusterResourcePlacement(t *testing.T) {
...
+					},
+					Strategy: placementv1beta1.RolloutStrategy{
+						Type: placementv1beta1.RollingUpdateRolloutStrategyType,
+						RollingUpdate: &placementv1beta1.RollingUpdateConfig{
+							MaxUnavailable: &intstr.IntOrString{
+								Type:   0,
+								IntVal: 10,
+							},
+						},
+					},
```
ryanzhang-oss, line 426: do we need to set this? We are not testing rollout strategy

test/e2e/utils_test.go
Arvindthiru: We can use expect statements with Eventually blocks but only if we use the gomega object which intercepts the failures correctly and fails if any one of the Expect statements within Eventually block fails https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode

pkg/utils/validator/clusterresourceplacement_test.go
Arvindthiru: The validation for rollout strategy didn't have a empty struct check, just added it now so no need to pass this now. Just wanted to ensure the error returned had to do with invalid placement policy

Arvindthiru:
After discussing offline: Looks like if the cluster names is empty we just skip schduling so removed check https://github.com/Azure/fleet/blob/main/pkg/scheduler/framework/framework.go#L1328-L1333

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -91,9 +92,10 @@ func ValidateClusterResourcePlacement(clusterResourcePlacement *placementv1beta1
 			allErr = append(allErr, fmt.Errorf("the clusterAffinity field is invalid: %w", err))
 		}
 	}
-
-	if err := validateRolloutStrategy(clusterResourcePlacement.Spec.Strategy); err != nil {
-		allErr = append(allErr, fmt.Errorf("the rollout Strategy field  is invalid: %w", err))
+	if !reflect.DeepEqual(clusterResourcePlacement.Spec.Strategy, placementv1beta1.RolloutStrategy{}) {
```
ryanzhang-oss, line 95: why compare it with an empty struct?

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -121,17 +123,14 @@ func validatePlacementPolicy(policy *placementv1beta1.PlacementPolicy) error {
 
 func validatePolicyForPickFixedPlacementType(policy *placementv1beta1.PlacementPolicy) error {
 	allErr := make([]error, 0)
-	if len(policy.ClusterNames) == 0 {
-		allErr = append(allErr, fmt.Errorf("cluster names cannot be empty for policy type %s", policy.PlacementType))
-	}
```
ryanzhang-oss: why remove this? I see the scheduling code that treat this as not picking anything. I think we should still block it.

pkg/utils/validator/clusterresourceplacement.go
Arvindthiru: I'm comparing it to an empty object because rollout strategy is not a pointer, not this check if being called before calling validateRolloutStrategy

ryanzhang-oss (dismissed):

ryanzhang-oss (approved):

PR #554, opened 2023-10-07
feat: add the support for configmap wrapper

ryanzhang-oss:

add the support for configmap wrapper
-- we create a separate work object for the wrapper
-- CRP populate the envelope identifier when a wrapped object is not applied successfully

Fixes #

I have:

UT/IT


apis/placement/v1beta1/commons.go
```diff
@@ -52,13 +61,23 @@ const (
...
+
+	// EnvelopeNamespaceLabel is the label that contains the namespace of the envelope object that the work is generated from.
+	EnvelopeNamespaceLabel = fleetPrefix + "envelope-namespace"
+
+	// EnvelopeNameLabel is the label that contains the name of the envelope object that the work is generated from.
+	EnvelopeNameLabel = fleetPrefix + "envelope-name"
 
 	// PreviousBindingStateAnnotation is the annotation that records the previous state of a binding.
 	// This is used to remember if an "unscheduled" binding was moved from a "bound" state or a "scheduled" state.
 	PreviousBindingStateAnnotation = fleetPrefix + "PreviousBindingState"
```
michaelawyu, line 82: Hi Ryan! Just a nit: do we want to change it to kebab case as well?

pkg/controllers/workgenerator/controller.go
```diff
@@ -224,52 +231,98 @@ func (r *Reconciler) listAllWorksAssociated(ctx context.Context, resourceBinding
...
+				simpleManifests = append(simpleManifests, fleetv1beta1.Manifest(selectedResource))
+			}
 		}
-		if updated {
-			updateAny = true
+		if len(simpleManifests) != 0 {
+			// generate a work object for the manifests if there are still any non enveloped resources
+			newWork = append(newWork, generateSnapshotWorkObj(workNamePrefix, resourceBinding, snapshot, simpleManifests))
+		} else {
+			klog.V(2).InfoS("Skip generating work for the snapshot since there is no none-enveloped resource in the snapshot", "snapshot", klog.KObj(snapshot))
```
michaelawyu, line 282: Hi Ryan! A nit here: this message seems to be only suggesting that there're only unwrapped resources; the skip part reads a little bit confusing to me personally.

pkg/controllers/workgenerator/controller.go
```diff
@@ -328,25 +381,47 @@ func (r *Reconciler) fetchAllResourceSnapshots(ctx context.Context, resourceBind
...
+func (r *Reconciler) getConfigMapEnvelopWorkObj(ctx context.Context, workNamePrefix string, resourceBinding *fleetv1beta1.ClusterResourceBinding,
+	resourceSnapshot *fleetv1beta1.ClusterResourceSnapshot, envelopeObj *unstructured.Unstructured) (*fleetv1beta1.Work, error) {
+	// we group all the resources in one configMap to one work
+	manifest, err := extractResFromConfigMap(envelopeObj)
+	if err != nil {
+		klog.ErrorS(err, "configMap has invalid content", "snapshot", klog.KObj(resourceSnapshot),
+			"resourceBinding", klog.KObj(resourceBinding), "configMapWrapper", klog.KObj(envelopeObj))
+		return nil, controller.NewUserError(err)
+	}
+	klog.V(2).InfoS("Successfully extract the enveloped resources from the configMap", "numOfResources", len(manifest), "configMapWrapper", klog.KObj(envelopeObj))
```
michaelawyu, line 396: Hi Ryan! Just a nit: would it be easier for debugging if we add the link to resource binding/resource snapshot here as well?

pkg/controllers/workgenerator/controller.go
```diff
@@ -328,25 +381,47 @@ func (r *Reconciler) fetchAllResourceSnapshots(ctx context.Context, resourceBind
...
+	// Try to see if we already have a work represent the same enveloped object for this CRP in the same cluster
+	envelopWorkLabelMatcher := client.MatchingLabels{
```
michaelawyu, line 398: Hi Ryan! Is there any particular reason why the ParentResourceSnapshotIndex label is omitted for this?

pkg/controllers/workgenerator/manifests/resourceQuota.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
michaelawyu, line 1: Hi Ryan! Just a nit: this folder seems to have names of all three cases: snake, kebab, and camel.

pkg/controllers/workgenerator/manifests/test-envelop-configmap.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - apiGroups:
+        - ""
+        apiVersions:
+        - v1
+        operations:
+        - CREATE
+        - UPDATE
+        resources:
+        - pods
+      sideEffects: None
```
michaelawyu, line 51: Hi Ryan! It seems that this files needs a blank line in the end.

pkg/controllers/workgenerator/manifests/test-envelop-configmap2.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - apiGroups:
+        - ""
+        apiVersions:
+        - v1
+        operations:
+        - CREATE
+        - UPDATE
+        resources:
+        - pods
+      sideEffects: None
```
michaelawyu, line 39: Hi Ryan! It seems that this file needs a blank line in the end.

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: yes, this is actually critical. The work is associated with the same CRP/binding and enveloped configmap all the time. However, it can move from one snapshot to another. Sometimes the content of the configmap doesn't change between snapshots but it can also change. Thus, the ParentResourceSnapshotIndex cannot be part of the label matching. You can see that we modify the ParentResourceSnapshotIndex  and the manifest in the new work.

pkg/controllers/workgenerator/manifests/resourceQuota.yaml
ryanzhang-oss: the files start with "test-" are actually applied to the k8s while other 2 files are just used to compare in the test.

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -954,16 +962,219 @@ func TestSetPlacementStatus(t *testing.T) {
...
+							ObservedGeneration: 1,
+							Status:             metav1.ConditionTrue,
+						},
+					},
+				},
+			},
+			wantIsPending: true,
+			wantRes:       nil,
+		},
+		"pending if applied failed with multiple object": {
```
michaelawyu, line 1020: Hi Ryan! Just a nit, but this should be something like "report failure if applied failed with multiple" object? The case below also specifies "pending" but is actually of a failure state, if I understand it correctly.

pkg/controllers/workgenerator/controller.go
```diff
@@ -224,52 +233,102 @@ func (r *Reconciler) listAllWorksAssociated(ctx context.Context, resourceBinding
...
 		}
-		if updated {
-			updateAny = true
+		if len(simpleManifests) == 0 {
+			klog.V(2).InfoS("the snapshot contains enveloped resource only", "snapshot", klog.KObj(snapshot))
+		}
+		// generate a work object for the manifests even if there is nothing to place
+		// to allow CRP to collect the status of the placement
+		// TODO (RZ): revisit to see if we need this hack
+		work := generateSnapshotWorkObj(workNamePrefix, resourceBinding, snapshot, simpleManifests)
```
michaelawyu, line 287: Hi Ryan! Just for clarification, when the CRP has only enveloped resources, there would still be one additional work that's empty, right?

ryanzhang-oss: good point, I realized that there is a problem with new work objects. This PR alone is not enough to enable wrapper.

michaelawyu: Understood 👌

michaelawyu (approved):
LGTM ;)

PR #556, opened 2023-10-07
fix: report cluster name as expected in CRP status when CRP controller reconciles unselected clusters

michaelawyu:

This PR fixes an issue where cluster name is left empty in the status when clusters are unselected.

I have:

- [x] Unit tests
- [x] Integration tests
- [x] E2E tests

N/A

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -915,6 +915,7 @@ func (r *Reconciler) setResourcePlacementStatusAndResourceConditions(ctx context
 		// TODO: we could improve the message by summarizing the failure reasons from all of the unselected clusters.
 		// For now, it starts from adding some sample failures of unselected clusters.
 		var rp fleetv1beta1.ResourcePlacementStatus
+		rp.ClusterName = unselected[i].ClusterName
 		scheduledCondition := metav1.Condition{
 			Status:             metav1.ConditionFalse,
 			Type:               string(fleetv1beta1.ResourceScheduledConditionType),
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -915,6 +915,7 @@ func (r *Reconciler) setResourcePlacementStatusAndResourceConditions(ctx context
 		// TODO: we could improve the message by summarizing the failure reasons from all of the unselected clusters.
 		// For now, it starts from adding some sample failures of unselected clusters.
 		var rp fleetv1beta1.ResourcePlacementStatus
+		rp.ClusterName = unselected[i].ClusterName
```
zhiying-lin, line 918: The cluster name is missing by design and set the clusterName only if the cluster is selected. In the future, we'll provide a summary for few unselected clusters in one condition, so that clusterName cannot be set for a single cluster. trying to keep it consistent with the future behavior.

michaelawyu: This is a very confusing behavior IMO. Semantics aside, now say if you have a PickN CRP of 6 target clusters with 3 picked and 3 cannot be picked, there would be 3 no-name decisions there but with very specific reasons. If we'd like to be consistent with future behavior, one decision would suffice.

zhiying-lin: This follows the "PlacementStatuses" definition.
https://github.com/Azure/fleet/blob/main/apis/placement/v1beta1/clusterresourceplacement_types.go#L343

// In the pickN case, there are N placement statuses where N = NumberOfClusters; Or in the pickFixed case, there are
// N placement statuses where N = ClusterNames.
// In these cases, some of them may not have assigned clusters when we cannot fill the required number of clusters.

The API require us to populate 6 PlacementStatuses for pickN or pickFixed. :(

PR #562, opened 2023-10-10
fix: Print GVK instead of Kind for guard rail logs

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/webhook/validation/uservalidation.go
+++ b/pkg/webhook/validation/uservalidation.go
@@ -26,9 +26,9 @@ const (
 	kubeControllerManagerUser = "system:kube-controller-manager"
 	serviceAccountFmt         = "system:serviceaccount:fleet-system:%s"
 
-	resourceAllowedFormat      = "user: %s in groups: %v is allowed to %s resource %s/%s: %+v"
-	resourceDeniedFormat       = "user: %s in groups: %v is not allowed to %s resource %s/%s: %+v"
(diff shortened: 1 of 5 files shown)


pkg/webhook/validation/uservalidation.go
```diff
@@ -26,9 +26,9 @@ const (
 	kubeControllerManagerUser = "system:kube-controller-manager"
 	serviceAccountFmt         = "system:serviceaccount:fleet-system:%s"
 
-	resourceAllowedFormat      = "user: %s in groups: %v is allowed to %s resource %s/%s: %+v"
-	resourceDeniedFormat       = "user: %s in groups: %v is not allowed to %s resource %s/%s: %+v"
-	resourceAllowedGetMCFailed = "user: %s in groups: %v is allowed to %s resource %s/%s: %+v because we failed to get MC"
+	resourceAllowedFormat      = "user: %s in groups: %v is allowed to %s resource %+v/%s: %+v"
```
ryanzhang-oss, line 29: what does "%+v/%s" look like in the log?

Arvindthiru: user: test-user in groups: [system:masters] is allowed to CREATE resource meta.k8s.io/v1, Kind=Role/: test-namespace/test-role

ryanzhang-oss (approved):

PR #569, opened 2023-10-19
fix: update policy snapshot status in cases where only the observed CRP generation changes

michaelawyu:

This PR fixes an issue where policy snapshot status will not be updated when the scheduling policy itself does not change, yet the CRP generation is incremented (i.e., resource only change).

It also added a unit test that is not checked in in prev. PRs.

I have:

- [x] Unit tests
- [x] Integration tests
- [x] E2E tests (upstream)

### DIFF
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -696,6 +696,13 @@ func (f *framework) updatePolicySnapshotStatusFromBindings(
 ) error {
 	policyRef := klog.KObj(policy)
 
+	// Retrieve the corresponding CRP generation.
+	observedCRPGeneration, err := annotations.ExtractObservedCRPGenerationFromPolicySnapshot(policy)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/scheduler/framework/framework.go
```diff
@@ -704,18 +711,16 @@ func (f *framework) updatePolicySnapshotStatusFromBindings(
 	// Compare the new decisions + condition with the old ones.
 	currentDecisions := policy.Status.ClusterDecisions
 	currentCondition := meta.FindStatusCondition(policy.Status.Conditions, string(placementv1beta1.PolicySnapshotScheduled))
-	if equalDecisions(currentDecisions, newDecisions) && condition.EqualCondition(currentCondition, &newCondition) {
+	if observedCRPGeneration == policy.Status.ObservedCRPGeneration &&
```
ryanzhang-oss, line 714: when will we hit the case that the decision/condition doesn't change but the CRP's generation has changed (selector changed)

ryanzhang-oss (approved):

PR #571, opened 2023-10-19
fix: explicitly set the selectedResources as an empty array

zhiying-lin:

According to the json mashal documentation, ```The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any empty array, slice, map, or string.```
https://pkg.go.dev/encoding/json#Marshal

To explicitly set the selectedResources as an empty array, the field should be required. Otherwise, it will be omitted.

Fixes #

I have:

It's covered by e2e tests

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -337,8 +337,7 @@ type RollingUpdateConfig struct {
 // ClusterResourcePlacementStatus defines the observed state of the ClusterResourcePlacement object.
 type ClusterResourcePlacementStatus struct {
 	// SelectedResources contains a list of resources selected by ResourceSelectors.
-	// +optional
-	SelectedResources []ResourceIdentifier `json:"selectedResources,omitempty"`
+	SelectedResources []ResourceIdentifier `json:"selectedResources"`
 
 	// PlacementStatuses contains a list of placement status on the clusters that are selected by PlacementPolicy.
 	// Each selected cluster according to the latest resource placement is guaranteed to have a corresponding placementStatuses.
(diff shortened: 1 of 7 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -337,8 +337,7 @@ type RollingUpdateConfig struct {
 // ClusterResourcePlacementStatus defines the observed state of the ClusterResourcePlacement object.
 type ClusterResourcePlacementStatus struct {
 	// SelectedResources contains a list of resources selected by ResourceSelectors.
-	// +optional
-	SelectedResources []ResourceIdentifier `json:"selectedResources,omitempty"`
+	SelectedResources []ResourceIdentifier `json:"selectedResources"`
```
ryanzhang-oss, line 340: Is it safe to make this change if we already have v1beta1 fleet?

zhiying-lin: we have to rollout the agent release first, and CRP controller will have 5 minutes reconcile to fulfil the selectedResources fields for the existing CRPs.

Then rollout the manifest changes to mark the field as required.

ryanzhang-oss: the CRP controller is filling the array now already but the empty array got ignore. I am not sure if those are persisted in the ETCD and just not displayed or got omitted before persisting. If those are not persisted then the existing CRP will all be invalid after the CRD change.

zhiying-lin: You're right. It will be a breaking change. Will create the CRD PR today.

zhiying-lin:
Checked the kubernetes api repo and it assumes the omitted field = empty array, eg, https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5535C1-L5538C106

images & VolumesInUse  won't be explicitly set if no images in the node or volumes are in use.

PR #578, opened 2023-10-23
fix: fix the case that rollout stuck when there are failed to apply binding

ryanzhang-oss:

Rollout can get stuck when there are failed to apply bindings since there are already max unavailable bindings. The fix is to update the failed bindings first since it won't make more bindings unavailable since they are already unavailable anyway.

Fixes #

I have:

- [x ] Run `make reviewable` to ensure this PR is ready for review.

UT/IT/E2E

### DIFF
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -354,7 +354,7 @@ func pickBindingsToRoll(allBindings []*fleetv1beta1.ClusterResourceBinding, late
 		"lowerBoundAvailableBindings", lowerBoundAvailableNumber, "maxNumberOfBindingsToRemove", maxNumberToRemove)
 
 	// we can still update the bindings that are failed to apply already regardless of the maxNumberToRemove
-	for i := 0; i < maxUnavailableNumber && i < len(applyFailedUpdateCandidates); i++ {
+	for i := 0; i < len(applyFailedUpdateCandidates); i++ {
 		toBeUpdatedBinding = append(toBeUpdatedBinding, applyFailedUpdateCandidates[i])
 	}
 	if maxNumberToRemove > 0 {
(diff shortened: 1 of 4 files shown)


pkg/controllers/rollout/controller.go
```diff
@@ -329,6 +352,11 @@ func pickBindingsToRoll(allBindings []*fleetv1beta1.ClusterResourceBinding, late
 	klog.V(2).InfoS("Calculated the max number of bindings to remove", "clusterResourcePlacement", klog.KObj(crp),
 		"maxUnavailableNumber", maxUnavailableNumber, "minAvailableNumber", minAvailableNumber,
 		"lowerBoundAvailableBindings", lowerBoundAvailableNumber, "maxNumberOfBindingsToRemove", maxNumberToRemove)
+
+	// we can still update the bindings that are failed to apply already regardless of the maxNumberToRemove
+	for i := 0; i < maxUnavailableNumber && i < len(applyFailedUpdateCandidates); i++ {
```
zhiying-lin, line 357: we don't care about the maxUnavailableNumber, either? as they're already failed binding.

michaelawyu, line 357: Hi Ryan! Apologies if I misunderstand anything, but would this still get stuck if, say, all bindings are bound but failed to get applied (-> maxNumberToRemove <= 0) + maxUnavailable set to 0 with a non-zero maxSurge?

michaelawyu, line 357: Hi Ryan! Apologies if I misunderstand anything, but would this still get stuck if, say, all bindings are bound but failed to get applied (-> maxNumberToRemove <= 0) + maxUnavailable set to 0 with a non-zero maxSurge?

ryanzhang-oss: yes, but we also don't want to blast it too fast. This is kind of a middle ground.

ryanzhang-oss: you mean when the "maxUnavailableNumber" == 0? Hmm.. that could be an issue.

zhiying-lin (approved):

PR #579, opened 2023-10-24
fix: fix the CRP condition

zhiying-lin:

- build CRP ResourcePlacementStatus using existing conditions
- build CRP sync condition by using clusterResouceBinding bound condition instead of works
- change the CRP reconcile time when we are waiting for resource applied to depend on the apply unavailable time.

Fixes #

I have:

Unit tests

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -141,12 +141,11 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 	crpKObj := klog.KObj(crp)
 	oldCRP := crp.DeepCopy()
 	if crp.Spec.RevisionHistoryLimit != nil {
-		revisionLimit = *crp.Spec.RevisionHistoryLimit
 		if revisionLimit <= 0 {
 			err := fmt.Errorf("invalid clusterResourcePlacement %s: invalid revisionHistoryLimit %d", crp.Name, revisionLimit)
 			klog.ErrorS(controller.NewUnexpectedBehaviorError(err), "Invalid revisionHistoryLimit value and using default value instead", "clusterResourcePlacement", crpKObj)
-			// use the default value instead
-			revisionLimit = fleetv1beta1.RevisionHistoryLimitDefaultValue
+		} else {
+			revisionLimit = *crp.Spec.RevisionHistoryLimit
 		}
 	}
 
@@ -205,6 +204,8 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementSyncSuccess", "Successfully synchronized the placement")
 	}
 
+	// There is no need to check if the CRP is applied or not.
+	// If the applied condition is true, it means the scheduling is done & works have been synchronized which means the rollout is completed.
 	if isRolloutCompleted(crp) {
 		if !isRolloutCompleted(oldCRP) {
 			klog.V(2).InfoS("Placement rollout has finished", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
@@ -215,12 +216,10 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
-	// There is no need to check if the CRP is applied or not.
-	// If the applied condition is true, it means the scheduling is done & works have been synchronized.
-	// So that the rollout should be completed, and it will never reach to this line.
-
 	klog.V(2).InfoS("Placement rollout has not finished yet and requeue the request", "clusterResourcePlacement", crpKObj, "status", crp.Status)
-	return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+	// we need to requeue the request to update the status of the resources.
+	// TODO: adept the requeue time based on the rollout status.
+	return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
 }
 
 func (r *Reconciler) getOrCreateClusterSchedulingPolicySnapshot(ctx context.Context, crp *fleetv1beta1.ClusterResourcePlacement, revisionHistoryLimit int) (*fleetv1beta1.ClusterSchedulingPolicySnapshot, error) {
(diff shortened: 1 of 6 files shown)


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -852,7 +852,56 @@ func classifyClusterDecisions(decisions []fleetv1beta1.ClusterDecision) (selecte
...
+	// filter out the latest resource bindings
+	for i := range bindings {
+		if !bindings[i].DeletionTimestamp.IsZero() {
+			klog.V(2).InfoS("Filtering out the deleting clusterResourceBinding", "clusterResourceBinding", klog.KObj(&bindings[i]))
+			continue
+		}
+
+		if len(bindings[i].Spec.TargetCluster) == 0 {
+			err := fmt.Errorf("targetCluster is empty on clusterResourceBinding %s", bindings[i].Name)
+			klog.ErrorS(controller.NewExpectedBehaviorError(err), "Found an invalid clusterResourceBinding and skipping it when building placement status", "clusterResourceBinding", klog.KObj(&bindings[i]), "clusterResourcePlacement", crpKObj)
```
michaelawyu, line 890: Should this be an unexpected behavior?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -203,52 +198,47 @@ func (r *Reconciler) setWorkStatusForResourcePlacementStatus(ctx context.Context
...
-	if desiredWorkCounter == newWorkCounter && oldWorkCounter == 0 {
-		// We have created all the works according to the latest resource snapshot.
-		return true, metav1.Condition{
-			Status:             metav1.ConditionTrue,
-			Type:               string(fleetv1beta1.ResourceWorkSynchronizedConditionType),
-			Reason:             WorkSynchronizeSucceededReason,
-			Message:            "Successfully Synchronized work(s) for placement",
-			ObservedGeneration: crp.Generation,
-		}, nil
+func buildWorkSynchronizedCondition(crp *fleetv1beta1.ClusterResourcePlacement, binding *fleetv1beta1.ClusterResourceBinding) (bool, metav1.Condition) {
```
michaelawyu, line 213: Ah, just for clarification, this will set the work synchronized condition as soon as the binding has an up-to-date, true bound condition?

michaelawyu:
This is a large PR; I am only half-way through -> will review the rest later.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -180,13 +181,7 @@ func (r *Reconciler) setWorkStatusForResourcePlacementStatus(ctx context.Context
...
 			klog.ErrorS(err, "Invalid work", "clusterResourcePlacement", crpKObj, "clusterResourceSnapshot", klog.KObj(latestResourceSnapshot), "work", workKObj)
 			return nil, nil, controller.NewUnexpectedBehaviorError(err)
-		} else if indexFromWork < resourceIndex {
-			// The work is pointing to the old resourceSnapshot.
-			// it means the rollout controller has not updated the binding yet or work generator has not handled this work yet.
-			oldWorkCounter++
-		} else { // indexFromWork = resourceIndex
-			// The work is pointing to the latest resourceSnapshot.
-			newWorkCounter++
+		} else if indexFromWork == resourceIndex {
```
michaelawyu, line 184: Also for clarification, we no longer handle the < case?

pkg/controllers/clusterresourceplacement/placement_status.go
michaelawyu: i.e., it's possible that the work synchronized condition becomes true even if the works still point to an old resource snapshot

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -203,52 +198,47 @@ func (r *Reconciler) setWorkStatusForResourcePlacementStatus(ctx context.Context
...
-		// We have created all the works according to the latest resource snapshot.
-		return true, metav1.Condition{
-			Status:             metav1.ConditionTrue,
-			Type:               string(fleetv1beta1.ResourceWorkSynchronizedConditionType),
-			Reason:             WorkSynchronizeSucceededReason,
-			Message:            "Successfully Synchronized work(s) for placement",
-			ObservedGeneration: crp.Generation,
-		}, nil
+func buildWorkSynchronizedCondition(crp *fleetv1beta1.ClusterResourcePlacement, binding *fleetv1beta1.ClusterResourceBinding) (bool, metav1.Condition) {
+	if binding != nil {
```
michaelawyu, line 214: A nit: could we add a comment here to suggest that the work generator will only add the bound condition if the all works associated with a binding has been synchronzied?

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -909,49 +968,1282 @@ func TestSetPlacementStatus(t *testing.T) {
...
+						},
+						{
+							ClusterName: cluster2Name,
+							Selected:    true,
+							Reason:      "success",
+						},
+					},
+				},
+			},
+			latestResourceSnapshot: &fleetv1beta1.ClusterResourceSnapshot{
```
michaelawyu, line 1670: A nit: this object seems to be used repeatedly in the file.

michaelawyu (commented):
A nit: some of the objects in the test file are repeatedly used. Not particularly urgent though

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -42,11 +62,15 @@ func TestSetPlacementStatus(t *testing.T) {
...
-		latestResourceSnapshot *fleetv1beta1.ClusterResourceSnapshot
-		wantStatus             *fleetv1beta1.ClusterResourcePlacementStatus
+		name                          string
+		crpStatus                     fleetv1beta1.ClusterResourcePlacementStatus
+		policy                        *fleetv1beta1.PlacementPolicy
+		latestPolicySnapshot          *fleetv1beta1.ClusterSchedulingPolicySnapshot
+		latestResourceSnapshot        *fleetv1beta1.ClusterResourceSnapshot
+		clusterResourceBindings       []fleetv1beta1.ClusterResourceBinding
+		works                         []fleetv1beta1.Work
+		wantLastTransitionTimeChanged bool
```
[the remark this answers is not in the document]
ryanzhang-oss: removed

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -909,49 +968,1282 @@ func TestSetPlacementStatus(t *testing.T) {
...
+					},
+				},
+			},
+			clusterResourceBindings: []fleetv1beta1.ClusterResourceBinding{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "binding-1",
+						Labels: map[string]string{
+							fleetv1beta1.CRPTrackingLabel: testName,
+						},
```
ryanzhang-oss, line 1362: we don't use those

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -909,49 +968,1282 @@ func TestSetPlacementStatus(t *testing.T) {
...
+						TargetCluster:                cluster1Name,
+					},
+					Status: fleetv1beta1.ResourceBindingStatus{
+						Conditions: []metav1.Condition{
+							{
+								Status:             metav1.ConditionTrue,
+								Type:               string(fleetv1beta1.ResourceBindingBound),
+								Reason:             "resourceBindingTrue",
+								ObservedGeneration: 1,
+								LastTransitionTime: oldTransitionTime,
```
ryanzhang-oss, line 1378: the msg of the condition is supposed to be copied to the CRP so we need to add msg and verify that.

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -909,49 +968,1282 @@ func TestSetPlacementStatus(t *testing.T) {
...
+						ResourceSnapshotName:         fmt.Sprintf(fleetv1beta1.ResourceSnapshotNameFmt, testName, 0),
+						SchedulingPolicySnapshotName: fmt.Sprintf(fleetv1beta1.PolicySnapshotNameFmt, testName, 0),
+						TargetCluster:                cluster1Name,
+					},
+					Status: fleetv1beta1.ResourceBindingStatus{
+						Conditions: []metav1.Condition{
+							{
+								Status:             metav1.ConditionFalse,
+								Type:               string(fleetv1beta1.ResourceBindingBound),
+								Reason:             "resourceBindingTrue",
```
ryanzhang-oss, line 1713: resourceBindingFailed?

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -909,49 +968,1282 @@ func TestSetPlacementStatus(t *testing.T) {
...
+						Status:             metav1.ConditionTrue,
+						Type:               string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType),
+						Reason:             "Scheduled",
+						ObservedGeneration: crpGeneration,
+						LastTransitionTime: oldTransitionTime,
+					},
+					{
+						Status:             metav1.ConditionFalse,
+						Type:               string(fleetv1beta1.ClusterResourcePlacementSynchronizedConditionType),
+						Reason:             SynchronizePendingReason,
```
ryanzhang-oss, line 1741: why is the reason still "pending" when one of the resource binding status is False already?

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -909,49 +968,1282 @@ func TestSetPlacementStatus(t *testing.T) {
...
+								Status:             metav1.ConditionUnknown,
+								Type:               string(fleetv1beta1.ResourcesAppliedConditionType),
+								Reason:             ResourceApplyPendingReason,
+								ObservedGeneration: crpGeneration,
+								LastTransitionTime: metav1.NewTime(currentTime),
+							},
+							{
+								Status:             metav1.ConditionFalse,
+								Type:               string(fleetv1beta1.ResourceWorkSynchronizedConditionType),
+								Reason:             WorkSynchronizeFailedReason,
```
ryanzhang-oss, line 1761: why not copy the resource binding reason/msg here?

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -909,49 +968,1282 @@ func TestSetPlacementStatus(t *testing.T) {
...
+								ObservedGeneration: crpGeneration,
```
ryanzhang-oss, line 1762: should this still be crpGeneration? This could be work/resource generation instead. The CRP generation doesn't tell the user much about the apply or sync condition as those can happen behind CRP's. back.

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: yes

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: will fix

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: i think we discussed this before. The conclusion is, if the sync has not finished, we always set the appliedCondition as unknown. Until the sync is finished, we will set the appliedCondition as true or false.

It makes the documentation easy otherwise there will be a lot of combinations and too complicated for the user.

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: yeah, that's true. But the condition is for the CRP resource itself. The new field resourceSnapshots of status should solve this issue.

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: prefer not to validate this logic in the UT and tests will become fragile if we reword the error messages in the future.

PR #580, opened 2023-10-26
test: fix: set all test cases which involves member cluster provisioning to run in serial

michaelawyu:

This PR fixes an issue where some test cases create member clusters as part of the test while running in parallel with others, which may lead to CRP test cases picking these clusters in unfortunately situations.

I have:

N/A

N/A


zhiying-lin:
Is the e2e test failure related to the fix?

Arvindthiru:
Should we add the Serial decorator to these Describe blocks as well https://github.com/Azure/fleet/pull/580/files#diff-3365133caddd9745d1550cb7d0342383451d7cd41a15576275ed78cc30a443adL162

ryanzhang-oss (approved):

PR #583, opened 2023-10-30
fix: improve CRP requeue

zhiying-lin:

Not enqueue the request when scheduler has not selected any cluster.

Fixes #

I have:

add e2e test

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -185,7 +185,10 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	if err := r.setPlacementStatus(ctx, crp, selectedResourceIDs, latestSchedulingPolicySnapshot, latestResourceSnapshot); err != nil {
+
+	// isClusterScheduled is to indicate whether we need to requeue the CRP request to track the rollout status.
+	isClusterScheduled, err := r.setPlacementStatus(ctx, crp, selectedResourceIDs, latestSchedulingPolicySnapshot, latestResourceSnapshot)
+	if err != nil {
 		return ctrl.Result{}, err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -216,9 +219,24 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
 	}
 
-	klog.V(2).InfoS("Placement rollout has not finished yet and requeue the request", "clusterResourcePlacement", crpKObj, "status", crp.Status)
+	if !isClusterScheduled {
+		// Note:
+		// 1. If the scheduledCondition is failed, it means the placement requirement cannot be satisfied fully. For example,
+		// pickN deployment requires 5 clusters and scheduler schedules the resources on 3 clusters. And the appliedCondition
+		// could be true when resources are applied successfully on these 3 clusters and the detailed the resourcePlacementStatuses
+		// need to be populated.
+		// So that we cannot reply on the scheduledCondition as false to decide whether to requeue the request.
```
michaelawyu, line 228: A nit: `reply` -> `rely`, if I understand it correctly?

test/e2e/placement_updating_members_test.go
```diff
@@ -0,0 +1,113 @@
...
+		})
+
+		It("should not place resources on any cluster", checkIfRemovedWorkResourcesFromAllMemberClusters)
+
+		It("should update CRP status as expected", func() {
+			statusUpdatedActual := crpStatusUpdatedActual(workResourceIdentifiers(), nil, []string{memberCluster1Name})
+			Eventually(statusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
+		})
+
+		It("update the member cluster label", func() {
```
michaelawyu, line 79: If this manipulates the member cluster info, we should consider marking it as a Serial block to avoid possible unintended interruptions to other test cases.

michaelawyu: Not a blocker though; will update the test case in the next bunch of E2E PRs.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -216,9 +219,24 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
-	klog.V(2).InfoS("Placement rollout has not finished yet and requeue the request", "clusterResourcePlacement", crpKObj, "status", crp.Status)
+	if !isClusterScheduled {
```
michaelawyu, line 222: Just a clarification: this shortcut would only work when there is no cluster selected at all, right?

michaelawyu (commented):
A few minor comments; nothing blocking. Feel free to merge directly to expedite the progress.

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: yeah, now that this only applies when there is no cluster selected then probably not that urgent.

test/e2e/placement_updating_members_test.go
```diff
@@ -0,0 +1,113 @@
...
+						RollingUpdate: &placementv1beta1.RollingUpdateConfig{
+							UnavailablePeriodSeconds: pointer.Int(2),
+						},
+					},
+				},
+			}
+			Expect(hubClient.Create(ctx, crp)).To(Succeed(), "Failed to create CRP")
+		})
+
+		It("should not place resources on any cluster", checkIfRemovedWorkResourcesFromAllMemberClusters)
```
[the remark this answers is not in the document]
zhiying-lin: removed, validating CRP status should be good enough.

ryanzhang-oss (approved):

PR #586, opened 2023-11-01
fix: flaky fleet guard rail E2E where namespace is terminating

Arvindthiru:

Fixes #

I have:


Arvindthiru:
The current E2E setup runs the tests in parallel but it only uses two processes meaning the member clusters that get created are either named mc-1 or mc-2. This PR fixes the flake mentioned in the screenshot down below 
image
Where a member cluster namespace is in terminating state but resources are being created on the same namespace which tells us that two Describe container are being run in parallel with the same GinkgoParallelProcess()

ryanzhang-oss (approved):

PR #597, opened 2023-11-06
feat: check for NOC and max skew

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -161,6 +161,9 @@ func validatePolicyForPickNPolicyType(policy *placementv1beta1.PlacementPolicy) 
 		if *policy.NumberOfClusters < 0 {
 			allErr = append(allErr, fmt.Errorf("number of clusters cannot be %d for policy type %s", *policy.NumberOfClusters, placementv1beta1.PickNPlacementType))
 		}
+		if *policy.NumberOfClusters > 100 {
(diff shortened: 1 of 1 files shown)


pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -200,6 +203,10 @@ func validateClusterAffinity(clusterAffinity *placementv1beta1.ClusterAffinity,
 func validateTopologySpreadConstraints(topologyConstraints []placementv1beta1.TopologySpreadConstraint) error {
 	allErr := make([]error, 0)
 	for _, tc := range topologyConstraints {
+		if *tc.MaxSkew > 98 {
```
ryanzhang-oss, line 206: where is the number 98 coming from?

Arvindthiru: **reasoning for 98:** there are two buckets for cluster where resource can be scheduled and not scheduled. assuming the distribution is 1:99 or 99:1 the max skew has to be 98.

I spoke with @michaelawyu he suggested there is no point in validating the max skew since it's dynamic as it's based on the number of clusters and it should be done as part of the API validation. @ryanzhang-oss thoughts ?

zhiying-lin: i suggest just checking if it is > 100, and we don't need to do the strict validation here and our implementation should cover it during the runtime.

ryanzhang-oss: Agree with  zhiying and michael.
Did we put a 100 cluster limit? If not then we don't need to check it here, our implementation can cover it.

Arvindthiru:
Closing PR as we decided it's not necessary at this point

PR #622, opened 2023-11-17
fix: add webhook calling timeout

ryanzhang-oss:

add webhook calling timeout

Fixes #

I have:


pkg/webhook/webhook.go
```diff
@@ -162,15 +162,14 @@ func (w *Config) buildValidatingWebHooks() []admv1.ValidatingWebhook {
 	sideEffortsNone := admv1.SideEffectClassNone
 	namespacedScope := admv1.NamespacedScope
 	clusterScope := admv1.ClusterScope
-
+	webhookTimeoutSeconds := pointer.Int32(1)
```
Arvindthiru, line 165: @ryanzhang-oss why is the timeout only set to 1 second ?

ryanzhang-oss: That's the least we can set

Arvindthiru (approved):

PR #623, opened 2023-11-18
feat: fleet guard rail static GVK list for namespaced resources

Arvindthiru:

Fixes #

I have:

Not monitoring leases, events. Tried adding E2E for local subject access review object but faced trouble creating object

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -314,17 +314,18 @@ func (w *Config) buildValidatingWebHooks() []admv1.ValidatingWebhook {
 				Operations: []admv1.OperationType{admv1.Delete},
 				Rule:       createRule([]string{"*"}, []string{"*"}, []string{"*/*"}, &namespacedScope),
 			},
+			// TODO(ArvindThiru): not handling pods, replicasets as part of the fleet guard rail since they have validating webhooks, need to remove validating webhooks before adding these resources to fleet guard rail.
 			{
 				Operations: cuOperations,
 				Rule: createRule([]string{corev1.SchemeGroupVersion.Group}, []string{corev1.SchemeGroupVersion.Version}, []string{bindingResourceName, configMapResourceName, endPointResourceName,
-					limitRangeResourceName, persistentVolumeClaimsName, persistentVolumeClaimsName + "/status", podResourceName, podResourceName + "/status", podTemplateResourceName,
+					limitRangeResourceName, persistentVolumeClaimsName, persistentVolumeClaimsName + "/status", podTemplateResourceName,
 					replicationControllerResourceName, replicationControllerResourceName + "/status", resourceQuotaResourceName, resourceQuotaResourceName + "/status", secretResourceName,
 					serviceAccountResourceName, servicesResourceName, servicesResourceName + "/status"}, &namespacedScope),
 			},
 			{
 				Operations: cuOperations,
 				Rule: createRule([]string{appsv1.SchemeGroupVersion.Group}, []string{appsv1.SchemeGroupVersion.Version}, []string{controllerRevisionResourceName, daemonSetResourceName, daemonSetResourceName + "/status",
-					deploymentResourceName, deploymentResourceName + "/status", replicaSetResourceName, replicaSetResourceName + "/status", statefulSetResourceName, statefulSetResourceName + "/status"}, &namespacedScope),
+					deploymentResourceName, deploymentResourceName + "/status", statefulSetResourceName, statefulSetResourceName + "/status"}, &namespacedScope),
 			},
 			{
 				Operations: cuOperations,
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -23,7 +23,16 @@ import (
 	admv1 "k8s.io/api/admissionregistration/v1"
 	admv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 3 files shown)


pkg/webhook/webhook.go
```diff
@@ -268,7 +308,69 @@ func (w *Config) buildValidatingWebHooks() []admv1.ValidatingWebhook {
 		namespacedResourcesRules := []admv1.RuleWithOperations{
 			{
 				Operations: cudOperations,
```
ryanzhang-oss, line 310: can we do delete for * and only CU for static list?

pkg/webhook/webhook.go
```diff
@@ -48,11 +58,41 @@ const (
...
-	replicaSetResourceName    = "replicasets"
-	podResourceName           = "pods"
+	crdResourceName                      = "customresourcedefinitions"
+	bindingResourceName                  = "bindings"
+	configMapResourceName                = "configmaps"
+	endPointResourceName                 = "endpoints"
+	eventResourceName                    = "events"
+	limitRangeResourceName               = "limitranges"
+	persistentVolumeClaimsName           = "persistentvolumeclaims"
+	podTemplateResourceName              = "podtemplates"
```
ryanzhang-oss, line 68: I am not sure what this is

pkg/webhook/webhook.go
```diff
@@ -48,11 +58,41 @@ const (
...
+	networkPolicyResourceName            = "networkpolicies"
+	podDisruptionBudgetsResourceName     = "poddisruptionbudgets"
+	roleResourceName                     = "roles"
+	roleBindingResourceName              = "rolebindings"
+	csiStorageCapacityResourceName       = "csistoragecapacities"
+	memberClusterResourceName            = "memberclusters"
+	internalMemberClusterResourceName    = "internalmemberclusters"
+	namespaceResourceName                = "namespaces"
+	replicaSetResourceName               = "replicasets"
+	podResourceName                      = "pods"
```
ryanzhang-oss, line 95: IIRC, we have a pod webhook already.

pkg/webhook/webhook.go
```diff
@@ -268,7 +308,69 @@ func (w *Config) buildValidatingWebHooks() []admv1.ValidatingWebhook {
...
+				Operations: cudOperations,
+				Rule:       createRule([]string{networkingv1.SchemeGroupVersion.Group}, []string{networkingv1.SchemeGroupVersion.Version}, []string{ingressResourceName, ingressResourceName + "/status", networkPolicyResourceName, networkPolicyResourceName + "/status"}, &namespacedScope),
+			},
+			{
+				Operations: cudOperations,
+				Rule:       createRule([]string{policyv1.SchemeGroupVersion.Group}, []string{policyv1.SchemeGroupVersion.Version}, []string{podDisruptionBudgetsResourceName, podDisruptionBudgetsResourceName + "/status"}, &namespacedScope),
+			},
+			{
+				Operations: cudOperations,
+				Rule:       createRule([]string{rbacv1.SchemeGroupVersion.Group}, []string{rbacv1.SchemeGroupVersion.Version}, []string{roleResourceName, roleResourceName + "/status", roleBindingResourceName, roleBindingResourceName + "/status"}, &namespacedScope),
```
ryanzhang-oss, line 361: not every resource has /status

pkg/webhook/webhook.go
ryanzhang-oss: what about our fleet resources?

pkg/webhook/webhook.go
Arvindthiru: It is a namespaced resource 
image

pkg/webhook/webhook.go
Arvindthiru: We have validating webhook for pod, replicaset which block their creation is fleet-, kube- prefixed namespaces. I guess we can remove these validating webhook configurations since we have fleet guard rail,

- https://github.com/Azure/fleet/blob/main/pkg/webhook/pod/pod_validating_webhook.go
- https://github.com/Azure/fleet/blob/main/pkg/webhook/replicaset/replicaset_validating_webhook.go

michaelawyu:
Hi Arvind! For the resources that have been watched by a different webhook, you might want to leave them out in this PR and send another one that removes these webhooks and add the resources to this webhook all together.

Arvindthiru:
Not handling pods, replicasets as part of the fleet guard rail since they have validating webhooks, need to remove validating webhooks before adding these resources to fleet guard rail

Will handle both in a follow up PR as suggested

pkg/webhook/webhook.go
```diff
@@ -48,11 +57,39 @@ const (
...
+	ingressResourceName                  = "ingresses"
+	networkPolicyResourceName            = "networkpolicies"
+	podDisruptionBudgetsResourceName     = "poddisruptionbudgets"
+	roleResourceName                     = "roles"
+	roleBindingResourceName              = "rolebindings"
+	csiStorageCapacityResourceName       = "csistoragecapacities"
+	memberClusterResourceName            = "memberclusters"
+	internalMemberClusterResourceName    = "internalmemberclusters"
+	namespaceResourceName                = "namespaces"
+	replicaSetResourceName               = "replicasets"
```
michaelawyu, line 91: Hi Arvind! I think these two values are no longer used, right?

michaelawyu (approved):
LGTM. Has discussed with Arvind; will confirm the behavior with pods/replicasets later, for now the original behavior (no pods/replicasets in non-reserved namespaces) will be kept.

PR #625, opened 2023-11-22
fix: make the failedResourcePlacements json consistent with field name

zhiying-lin:
It should be safe to rename the field as the etcd stores the json.

Fixes #

I have:


Arvindthiru (approved):

ryanzhang-oss (commented):
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning

IIRC, it seems that this is safe, but I think we need to manually test the CRD change to make sure it won't break existing CRP.

zhiying-lin:
instead of updating the json, i think it's better to update the field name. Will test and send another PR.

ryanzhang-oss (approved):

PR #635, opened 2023-12-12
feat: set observedResourceIndex for CRP status

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/test/e2e/actuals_test.go
+++ b/test/e2e/actuals_test.go
@@ -82,6 +82,23 @@ func workNamespaceAndConfigMapPlacedOnClusterActual(cluster *framework.Cluster) 
 	}
 }
 
+func workNamespacePlacedOnClusterActual(cluster *framework.Cluster) func() error {
+	workNamespaceName := fmt.Sprintf(workNamespaceNameTemplate, GinkgoParallelProcess())
+
+	return func() error {
+		return validateWorkNamespaceOnCluster(cluster, types.NamespacedName{Name: workNamespaceName})
+	}
+}
+
+func workConfigMapPlacedOnClusterActual(cluster *framework.Cluster) func() error {
+	workNamespaceName := fmt.Sprintf(workNamespaceNameTemplate, GinkgoParallelProcess())
+	appConfigMapName := fmt.Sprintf(appConfigMapNameTemplate, GinkgoParallelProcess())
+
+	return func() error {
+		return validateConfigMapOnCluster(cluster, types.NamespacedName{Namespace: workNamespaceName, Name: appConfigMapName})
+	}
+}
+
 func crpSyncFailedConditions(generation int64) []metav1.Condition {
 	return []metav1.Condition{
 		{
--- a/test/e2e/actuals_test.go
+++ b/test/e2e/actuals_test.go
@@ -90,15 +90,6 @@ func workNamespacePlacedOnClusterActual(cluster *framework.Cluster) func() error
 	}
 }
 
-func workConfigMapPlacedOnClusterActual(cluster *framework.Cluster) func() error {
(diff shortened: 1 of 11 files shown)


test/e2e/actuals_test.go
```diff
@@ -253,6 +253,7 @@ func workResourceIdentifiers() []placementv1beta1.ResourceIdentifier {
 func crpStatusUpdatedActual(
 	wantSelectedResourceIdentifiers []placementv1beta1.ResourceIdentifier,
 	wantSelectedClusters, wantUnselectedClusters []string,
+	wantObservedResourceIndex string,
```
zhiying-lin, line 256:
```suggestion
	wantObservedResourceIndex int,
```

can you use the int as the input type? Using string type as the index is kind of weird.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -387,7 +387,7 @@ var _ = Describe("validating CRP when cluster-scoped and namespace-scoped resour
 	It("should update the selected resources on member clusters", checkIfPlacedWorkResourcesOnAllMemberClusters)
 
 	It("should update CRP status as expected", func() {
-		crpStatusUpdatedActual := crpStatusUpdatedActual(workResourceIdentifiers(), allMemberClusterNames, nil)
```
zhiying-lin, line 390: It could be either 1 or 2. It depends on the when our controller detects the changes.

Could you please update the test to update the resources one by one?

eg, update one resource -> validate -> update another resource -> validate

test/e2e/actuals_test.go
Arvindthiru: The ObservedResourceIndex field in the API is string https://github.com/Azure/fleet/blob/main/apis/placement/v1beta1/clusterresourceplacement_types.go#L353 that's the reason I went with string.

If we pass int we have to convert to string though

test/e2e/actuals_test.go
```diff
@@ -82,6 +82,23 @@ func workNamespaceAndConfigMapPlacedOnClusterActual(cluster *framework.Cluster)
...
 
+func workNamespacePlacedOnClusterActual(cluster *framework.Cluster) func() error {
+	workNamespaceName := fmt.Sprintf(workNamespaceNameTemplate, GinkgoParallelProcess())
+
+	return func() error {
+		return validateWorkNamespaceOnCluster(cluster, types.NamespacedName{Name: workNamespaceName})
+	}
+}
+
+func workConfigMapPlacedOnClusterActual(cluster *framework.Cluster) func() error {
```
zhiying-lin, line 93: what's the diff between this func and func workNamespaceAndConfigMapPlacedOnClusterActual?

They're the same.

test/e2e/utils_test.go
```diff
@@ -533,6 +533,24 @@ func checkIfPlacedWorkResourcesOnAllMemberClusters() {
...
+func checkIfPlacedNamespaceResourceOnAllMemberClusters() {
+	for idx := range allMemberClusters {
+		memberCluster := allMemberClusters[idx]
+
+		namespaceResourcePlacedActual := workNamespacePlacedOnClusterActual(memberCluster)
+		Eventually(namespaceResourcePlacedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place work namespace on member cluster %s", memberCluster.ClusterName)
+	}
+}
+
+func checkIfPlacedConfigMapResourceOnAllMemberClusters() {
```
zhiying-lin, line 545: it's duplicate of checkIfPlacedWorkResourcesOnAllMemberClusters()

zhiying-lin (approved):

PR #638, opened 2023-12-14
feat: add allowedPropagatingAPIs option for resource selection

### LINKED ISSUE #633: [feature] Add AllowedAPIs option for the resources propagated from the namespace
```
// SkippedPropagatingAPIs indicates comma separated resources that should be skipped for propagating.
SkippedPropagatingAPIs string
```
All resources in the namespace configured with CRP are propagated to the member clusters by default. The hub agent option [`SkippedPropagatingAPIs`](https://github.com/Azure/fleet/blob/6fd37978c0f695b702a069ca6cf9e96c0d2585dc/cmd/hubagent/options/options.go#L52C1-L52C1) can be used to exclude propagating certain APIs from the namespace. The list of skipped APIs can get long to specify all resources and it needs to be updated when the new use cases come up. Alternatively an `AllowedAPIs` option can allow us to be selective while propagating resources. Having a limited set of allowed resources should also reduce reconciliations on the CRPs as it might not trigger on any change in the namespace.

vasudev-bongale:
### Changes

- Add `--allowed-propagating-apis` option to hub-agent command.
- Validate that `--allowed-propagating-apis` and `--skipped-propagating-apis` are mutually exclusive and cannot be specified together.
- Refactor `DisabledResourceConfig` type into a generic `ResourceConfig` so that we can reuse it for both `Allowed` and `Disabled` resource configs. During initializing `DisabledResourceConfig`, fleet and built-in resources are configured.
- Update `change_detector` to watch only the allowed resources if specified.
- Update `resource_selector` to only select allowed resources if specified.
- Add a util function with unit test to determine if a ResourceConfig is empty

#### Other changes
- Add make `help` recipe for developers to list the available commands in the Makefile.

Fixes #633 

I have:

Verify that the options are mutually exclusive.
```
E1213 17:04:40.247238   91214 main.go:79] "invalid parameter" err="Options.AllowedPropagatingAPIs: Invalid value: \"policy/v1\": AllowedPropagatingAPIs and SkippedPropagatingAPIs are mutually exclusive"
exit status 1
```

Run hub-agent locally, its able to start with the right option value for `--allowed-propagating-apis`
```
❯ export POD_NAMESPACE=fleet-system
❯ go run ./cmd/hubagent/main.go --enable-v1alpha1-apis=false --enable-v1beta1-apis=true --allowed-propagating-apis=rbac.authorization.k8s.io/v1 --v=1
I1214 09:19:00.292904   85484 main.go:76] "flag:" name="add_dir_header" value="false"
I1214 09:19:00.293211   85484 main.go:76] "flag:" name="allowed-propagating-apis" value="rbac.authorization.k8s.io/v1"
```

<details>
<summary>Verify that the Allowed resources are being watched and informer is created</summary>

```
I1214 09:19:11.823118   85484 change_dector.go:205] "Watching Allowed Resource" GVK="rbac.authorization.k8s.io/v1, Kind=ClusterRole"
I1214 09:19:11.823254   85484 change_dector.go:205] "Watching Allowed Resource" GVK="rbac.authorization.k8s.io/v1, Kind=Role"
I1214 09:19:11.823345   85484 change_dector.go:205] "Watching Allowed Resource" GVK="rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding"
I1214 09:19:11.823394   85484 change_dector.go:205] "Watching Allowed Resource" GVK="rbac.authorization.k8s.io/v1, Kind=RoleBinding"
I1214 09:19:11.826453   85484 informermanager.go:123] "Added an informer for a new resource" res={"GroupVersionKind":{"Group":"rbac.authorization.k8s.io","Version":"v1","Kind":"ClusterRole"},"GroupVersionResource":{"Group":"rbac.authorization.k8s.io","Version":"v1","Resource":"clusterroles"},"IsClusterScoped":true}
I1214 09:19:11.826502   85484 informermanager.go:123] "Added an informer for a new resource" res={"GroupVersionKind":{"Group":"rbac.authorization.k8s.io","Version":"v1","Kind":"Role"},"GroupVersionResource":{"Group":"rbac.authorization.k8s.io","Version":"v1","Resource":"roles"},"IsClusterScoped":false}
I1214 09:19:11.826519   85484 informermanager.go:123] "Added an informer for a new resource" res={"GroupVersionKind":{"Group":"rbac.authorization.k8s.io","Version":"v1","Kind":"ClusterRoleBinding"},"GroupVersionResource":{"Group":"rbac.authorization.k8s.io","Version":"v1","Resource":"clusterrolebindings"},"IsClusterScoped":true}
I1214 09:19:11.826536   85484 informermanager.go:123] "Added an informer for a new resource" res={"GroupVersionKind":{"Group":"rbac.authorization.k8s.io","Version":"v1","Kind":"RoleBinding"},"GroupVersionResource":{"Group":"rbac.authorization.k8s.io","Version":"v1","Resource":"rolebindings"},"IsClusterScoped":false}
```

</details>

<details>
<summary>It selects the correct resources</summary>

```
    - lastTransitionTime: "2023-12-14T03:54:43Z"
      message: Successfully applied resources
      observedGeneration: 1
      reason: ApplySucceeded
      status: "True"
      type: ResourceApplied
  selectedResources:
  - group: rbac.authorization.k8s.io
    kind: Role
    name: vbongale-test-lease-role
    namespace: vbongale-test-ns
    version: v1
  - kind: Namespace
    name: vbongale-test-ns
    version: v1
```

</details>

```
Running tool: /opt/homebrew/bin/go test -timeout 30s -run ^TestResourceConfigIsEmpty$ go.goms.io/fleet/pkg/utils
ok  	go.goms.io/fleet/pkg/utils
```

<details>
<summary>$ make help</summary>

```

❯ make help

Usage:
  make <target>
  help             Display this help.
  lint-full        Run slower linters to detect possible issues
  fmt              Run go fmt against code.
  vet              Run go vet against code.
  test             Run tests.
  local-unit-test  Run tests.
  integration-test  Run tests.
  build            Build agent binaries.
  run-hubagent     Run a controllers from your host.
  run-memberagent  Run a controllers from your host.
  clean-bin        Remove all generated binaries
```

</details>

### DIFF
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -105,8 +105,11 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		}
 	}
 
+	// AllowedPropagatingAPIs and SkippedPropagatingAPIs are mutually exclusive.
+    // If none of them are set, the resourceConfig by default stores a list of skipped propagation APIs.
 	resourceConfig := utils.NewResourceConfig(opts.AllowedPropagatingAPIs != "")
 	if err := resourceConfig.Parse(opts.AllowedPropagatingAPIs); err != nil {
+		// The program will never go here because the parameters have been checked.
 		return err
 	}
 	if err := resourceConfig.Parse(opts.SkippedPropagatingAPIs); err != nil {
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -105,14 +105,12 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		}
 	}
 
-	disabledResourceConfig := utils.NewResourceConfig(true)
-	if err := disabledResourceConfig.Parse(opts.SkippedPropagatingAPIs); err != nil {
-		// The program will never go here because the parameters have been checked
+	disabledResourceConfig := utils.NewResourceConfig(opts.AllowedPropagatingAPIs != "")
+	if err := disabledResourceConfig.Parse(opts.AllowedPropagatingAPIs); err != nil {
 		return err
 	}
-
-	allowedResourceConfig := utils.NewResourceConfig(false)
-	if err := allowedResourceConfig.Parse(opts.AllowedPropagatingAPIs); err != nil {
+	if err := disabledResourceConfig.Parse(opts.SkippedPropagatingAPIs); err != nil {
+		// The program will never go here because the parameters have been checked
 		return err
 	}
 
@@ -138,7 +136,6 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		RestMapper:             mgr.GetRESTMapper(),
 		InformerManager:        dynamicInformerManager,
 		DisabledResourceConfig: disabledResourceConfig,
-		AllowedResourceConfig:  allowedResourceConfig,
 		SkippedNamespaces:      skippedNamespaces,
 		Scheme:                 mgr.GetScheme(),
 		UncachedReader:         mgr.GetAPIReader(),
@@ -276,7 +273,6 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		MemberClusterPlacementController:           memberClusterPlacementController,
 		InformerManager:                            dynamicInformerManager,
 		DisabledResourceConfig:                     disabledResourceConfig,
-		AllowedResourceConfig:                      allowedResourceConfig,
 		SkippedNamespaces:                          skippedNamespaces,
 		ConcurrentClusterPlacementWorker:           opts.ConcurrentClusterPlacementSyncs,
 		ConcurrentResourceChangeWorker:             opts.ConcurrentResourceChangeSyncs,
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -105,9 +105,14 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		}
 	}
 
-	disabledResourceConfig := utils.NewDisabledResourceConfig()
+	disabledResourceConfig := utils.NewResourceConfig(true)
 	if err := disabledResourceConfig.Parse(opts.SkippedPropagatingAPIs); err != nil {
 		// The program will never go here because the parameters have been checked
+		return err
+	}
+
+	allowedResourceConfig := utils.NewResourceConfig(false)
+	if err := allowedResourceConfig.Parse(opts.AllowedPropagatingAPIs); err != nil {
 		return err
 	}
 
@@ -133,6 +138,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		RestMapper:             mgr.GetRESTMapper(),
 		InformerManager:        dynamicInformerManager,
 		DisabledResourceConfig: disabledResourceConfig,
+		AllowedResourceConfig:  allowedResourceConfig,
 		SkippedNamespaces:      skippedNamespaces,
 		Scheme:                 mgr.GetScheme(),
 		UncachedReader:         mgr.GetAPIReader(),
@@ -270,6 +276,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		MemberClusterPlacementController:           memberClusterPlacementController,
 		InformerManager:                            dynamicInformerManager,
 		DisabledResourceConfig:                     disabledResourceConfig,
+		AllowedResourceConfig:                      allowedResourceConfig,
 		SkippedNamespaces:                          skippedNamespaces,
 		ConcurrentClusterPlacementWorker:           opts.ConcurrentClusterPlacementSyncs,
 		ConcurrentResourceChangeWorker:             opts.ConcurrentResourceChangeSyncs,
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -105,11 +105,11 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		}
 	}
 
-	disabledResourceConfig := utils.NewResourceConfig(opts.AllowedPropagatingAPIs != "")
-	if err := disabledResourceConfig.Parse(opts.AllowedPropagatingAPIs); err != nil {
+	resourceConfig := utils.NewResourceConfig(opts.AllowedPropagatingAPIs != "")
+	if err := resourceConfig.Parse(opts.AllowedPropagatingAPIs); err != nil {
 		return err
 	}
-	if err := disabledResourceConfig.Parse(opts.SkippedPropagatingAPIs); err != nil {
+	if err := resourceConfig.Parse(opts.SkippedPropagatingAPIs); err != nil {
 		// The program will never go here because the parameters have been checked
 		return err
 	}
@@ -131,14 +131,14 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 
 	// Set up  a custom controller to reconcile cluster resource placement
 	crpc := &clusterresourceplacement.Reconciler{
-		Client:                 mgr.GetClient(),
-		Recorder:               mgr.GetEventRecorderFor(crpControllerName),
-		RestMapper:             mgr.GetRESTMapper(),
-		InformerManager:        dynamicInformerManager,
-		DisabledResourceConfig: disabledResourceConfig,
-		SkippedNamespaces:      skippedNamespaces,
-		Scheme:                 mgr.GetScheme(),
-		UncachedReader:         mgr.GetAPIReader(),
+		Client:            mgr.GetClient(),
+		Recorder:          mgr.GetEventRecorderFor(crpControllerName),
+		RestMapper:        mgr.GetRESTMapper(),
+		InformerManager:   dynamicInformerManager,
+		ResourceConfig:    resourceConfig,
+		SkippedNamespaces: skippedNamespaces,
+		Scheme:            mgr.GetScheme(),
+		UncachedReader:    mgr.GetAPIReader(),
 	}
 
 	rateLimiter := options.DefaultControllerRateLimiter(opts.RateLimiterOpts)
@@ -272,7 +272,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 		ResourceChangeController:                   resourceChangeController,
 		MemberClusterPlacementController:           memberClusterPlacementController,
 		InformerManager:                            dynamicInformerManager,
-		DisabledResourceConfig:                     disabledResourceConfig,
+		ResourceConfig:                             resourceConfig,
 		SkippedNamespaces:                          skippedNamespaces,
 		ConcurrentClusterPlacementWorker:           opts.ConcurrentClusterPlacementSyncs,
 		ConcurrentResourceChangeWorker:             opts.ConcurrentResourceChangeSyncs,
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -106,7 +106,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 	}
 
 	// AllowedPropagatingAPIs and SkippedPropagatingAPIs are mutually exclusive.
-    // If none of them are set, the resourceConfig by default stores a list of skipped propagation APIs.
+	// If none of them are set, the resourceConfig by default stores a list of skipped propagation APIs.
 	resourceConfig := utils.NewResourceConfig(opts.AllowedPropagatingAPIs != "")
 	if err := resourceConfig.Parse(opts.AllowedPropagatingAPIs); err != nil {
 		// The program will never go here because the parameters have been checked.
--- a/pkg/controllers/clusterresourceplacement/placement_controllerv1alpha1.go
+++ b/pkg/controllers/clusterresourceplacement/placement_controllerv1alpha1.go
@@ -50,7 +50,11 @@ type Reconciler struct {
 	UncachedReader client.Reader
 
 	// DisabledResourceConfig contains all the api resources that we won't select.
-	DisabledResourceConfig *utils.DisabledResourceConfig
+	DisabledResourceConfig *utils.ResourceConfig
+
+	// AllowedResourceConfig contains all the api resources that are selected for
+	// propagation. This is mutually exclusive with DisabledResourceConfig
+	AllowedResourceConfig *utils.ResourceConfig
 
 	// SkippedNamespaces contains the namespaces that we should not propagate.
 	SkippedNamespaces map[string]bool
--- a/pkg/controllers/clusterresourceplacement/placement_controllerv1alpha1.go
+++ b/pkg/controllers/clusterresourceplacement/placement_controllerv1alpha1.go
@@ -49,7 +49,7 @@ type Reconciler struct {
 	// It's only needed by v1beta1 APIs.
 	UncachedReader client.Reader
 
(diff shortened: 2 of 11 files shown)


pkg/utils/apiresources.go
```diff
@@ -162,9 +175,9 @@ func (r *DisabledResourceConfig) parseSingle(token string) error {
 	return nil
 }
 
-// IsResourceDisabled returns whether a given GroupVersionKind is disabled.
-// a gkv is disabled if its group or group version is disabled
-func (r *DisabledResourceConfig) IsResourceDisabled(gvk schema.GroupVersionKind) bool {
```
zhiying-lin, line 167: Should we keep the original interface?

you can have something like, and then you don't need to update the caller code.

```
func (r *ResourceConfig) IsResourceDisabled(gvk schema.GroupVersionKind) bool {
	isConfigured := r.isResourceConfigured(gvk)
	if r.isAllowedList {
		return !isConfigured
	}
	return isConfigured
}
```
and change the "IsResourceConfigured" to "isResourceConfigured", which is more like an internal util func.

cmd/hubagent/workload/setup.go
```diff
@@ -105,12 +105,17 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
...
 	}
 
-	disabledResourceConfig := utils.NewDisabledResourceConfig()
+	disabledResourceConfig := utils.NewResourceConfig(true)
 	if err := disabledResourceConfig.Parse(opts.SkippedPropagatingAPIs); err != nil {
 		// The program will never go here because the parameters have been checked
 		return err
 	}
 
+	allowedResourceConfig := utils.NewResourceConfig(false)
```
zhiying-lin, line 114: only one of them can be set. You don't need to maintain two configurations. Please see the comments below.

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -287,18 +287,29 @@ func (r *Reconciler) fetchAllResourcesInOneNamespace(namespaceName string, place
 	return resources, nil
 }
 
-// shouldSelectResource returns whether a resource should be propagated
+// shouldSelectResource returns whether a resource should be selected for propagation
```
zhiying-lin, line 290:
```suggestion
// shouldSelectResource returns whether a resource should be selected for propagation.
```
missing period

pkg/utils/apiresources.go
```diff
@@ -44,41 +44,54 @@ var (
 	apiGroupSepToken = ";"
 )
 
-// DisabledResourceConfig represents the configuration that identifies the API resources should not be selected.
-type DisabledResourceConfig struct {
-	// groups holds a collection of API group, all resources under this group will be avoided.
+// ResourceConfig represents the configuration that identifies the API resources that are parsed from the
+// user input to either allow or disable propagating them.
+type ResourceConfig struct {
```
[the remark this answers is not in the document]
vasudev-bongale: Having a single Config makes sense since the callers need not change. Thanks for the suggestion. I still renamed `DisabledResourceConfig` to `ResourceConfig` so that we can use for both with a flag.

pkg/controllers/clusterresourceplacement/placement_controllerv1alpha1.go
```diff
@@ -50,7 +50,7 @@ type Reconciler struct {
 	UncachedReader client.Reader
 
 	// DisabledResourceConfig contains all the api resources that we won't select.
-	DisabledResourceConfig *utils.DisabledResourceConfig
+	DisabledResourceConfig *utils.ResourceConfig
```
ryanzhang-oss, line 53:
```suggestion
	ResourceConfig *utils.ResourceConfig
```

pkg/resourcewatcher/change_dector.go
```diff
@@ -65,7 +65,7 @@ type ChangeDetector struct {
 	InformerManager informer.Manager
 
 	// DisabledResourceConfig contains all the api resources that we won't select
-	DisabledResourceConfig *utils.DisabledResourceConfig
+	DisabledResourceConfig *utils.ResourceConfig
```
ryanzhang-oss, line 68:
```suggestion
	ResourceConfig *utils.ResourceConfig
```

pkg/utils/apiresources_test.go
```diff
@@ -305,6 +305,24 @@ func TestDisabledResourceConfigMixedParse(t *testing.T) {
 			}
 		}
 	}
+	for _, test := range tests {
+		// test allow list
+		r := NewResourceConfig(true)
+		if err := r.Parse(test.input); err != nil {
+			t.Fatalf("Unexpected error: %v", err)
```
zhiying-lin, line 312:
```suggestion
			t.Fatalf("Parse() returned error: %v", err)
```

https://go.dev/wiki/TestComments#identify-the-function
`In most tests, failure messages should include the name of the function that failed, even though it seems obvious from the name of the test function.`

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -287,9 +287,9 @@ func (r *Reconciler) fetchAllResourcesInOneNamespace(namespaceName string, place
 	return resources, nil
 }
 
-// shouldSelectResource returns whether a resource should be propagated
+// shouldSelectResource returns whether a resource should be selected for propagation.
 func (r *Reconciler) shouldSelectResource(gvr schema.GroupVersionResource) bool {
-	if r.DisabledResourceConfig == nil {
+	if r.ResourceConfig.IsEmpty() {
```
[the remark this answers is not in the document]
vasudev-bongale: `What if the isAllow is true and resourceConfig is empty? we should not select the resource.` - Agreed.

However the ResourceConfig will never be a nil, instead an empty struct is returned by the following function, right? Hence I added another method to check if the ResourceConfig is actually empty of any resources in it. 
```
func NewResourceConfig(isAllowList bool) *ResourceConfig {
	r := &ResourceConfig{
		groups:            map[string]struct{}{},
		groupVersions:     map[schema.GroupVersion]struct{}{},
		groupVersionKinds: map[schema.GroupVersionKind]struct{}{},
	}
```

pkg/utils/apiresources_test.go
```diff
@@ -288,7 +288,7 @@ func TestDisabledResourceConfigMixedParse(t *testing.T) {
 		},
 	}
 	for _, test := range tests {
-		r := NewDisabledResourceConfig()
+		r := NewResourceConfig(false)
```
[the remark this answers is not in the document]
vasudev-bongale: I refactored both `TestDisabledResourceConfigMixedParse` and the DefaultResourceConfig tests.

pkg/controllers/clusterresourceplacement/resource_selector.go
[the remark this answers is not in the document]
zhiying-lin: To make it clear, there are two issues.

1. about "nil", right now, we always initialize the reconciler using NewResourceConfig(), but you cannot prevent other caller creating a reconciler without setting the resourceConfig. It's better to add the check to prevent the regression/panic in the future.
2. about adding r.ResourceConfig.IsEmpty() as a shortcut, i'm fine with checking whether it's empty or not as a shortcut though it's redundant and existing code should cover that.

If you want to add that, it should be

1) expose the isAllowList value
2) and then based on this value to return,

eg, if r.ResourceConfig.IsEmpty() {
      return !r.ResourceConfig.isAllowList
}

Could you please add a UT to test this func to prevent the regression? today we have the tests covered for the SkippedPropagatingAPIs list, but not for the AllowedPropagatingAPIs.

pkg/utils/apiresources.go
```diff
@@ -180,17 +198,22 @@ func (r *DisabledResourceConfig) IsResourceDisabled(gvk schema.GroupVersionKind)
 	return false
 }
 
-// DisableGroup to disable group.
-func (r *DisabledResourceConfig) DisableGroup(g string) {
+// AddGroup to store group in the resource config.
```
zhiying-lin, line 201:
```suggestion
// AddGroup stores group in the resource config.
```

Could you please help us to fix the comments? The existing codes are not following the golang best practices.  :(

` Comments documenting declarations should be full sentences, even if that seems a little redundant. This approach makes them format well when extracted into godoc documentation. Comments should begin with the name of the thing being described and end in a period:`
https://go.dev/wiki/CodeReviewComments#comment-sentences

pkg/utils/apiresources_test.go
```diff
@@ -228,144 +228,196 @@ func TestResourceConfigGroupParse(t *testing.T) {
...
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			r := NewResourceConfig(test.isAllowList)
+			if err := r.Parse(input); err != nil {
+				t.Fatalf("Parse() returned error: %v", err)
+			}
+
+			for i, o := range test.disabled {
+				if ok := r.IsResourceDisabled(o); !ok {
+					t.Errorf("%d: expected resource to be disabled : %v", i, o)
```
zhiying-lin, line 312:
```suggestion
					t.Errorf("IsResourceDisabled(%v) = false, want true", o)
```
Could you please fix the test code to follow the test output format?

we're following golang best practices here
`Test outputs should output the actual value that the function returned before printing the value that was expected. A usual format for printing test outputs is “YourFunc(%v) = %v, want %v”.`
https://go.dev/wiki/TestComments#got-before-want

`Do not use the index of the test in the test table as a substitute for naming your tests or printing the inputs. Nobody wants to go through your test table and count the entries in order to figure out which test case is failing.`
https://go.dev/wiki/TestComments#identify-the-input

pkg/utils/apiresources_test.go
```diff
@@ -228,144 +228,196 @@ func TestResourceConfigGroupParse(t *testing.T) {
...
+				}
+			}
+
+			for i, o := range test.enabled {
+				if ok := r.IsResourceDisabled(o); ok {
+					t.Errorf("%d: expected resource to be enabled : %v", i, o)
```
zhiying-lin, line 318:
```suggestion
					t.Errorf("IsResourceDisabled(%v) = true, want false", o)
```

pkg/utils/apiresources_test.go
```diff
@@ -228,144 +228,196 @@ func TestResourceConfigGroupParse(t *testing.T) {
...
+				t.Fatalf("Parse() returned error: %v", err)
 			}
-		}
-		for i, o := range test.enabled {
-			ok := r.IsResourceDisabled(o)
-			if ok {
-				t.Errorf("%d: unexpected error: %v", i, o)
+			for i, o := range test.disabled {
+				if ok := r.IsResourceDisabled(o); !ok {
+					t.Errorf("%d: expected resource to be disabled : %v", i, o)
```
zhiying-lin, line 387:
```suggestion
					t.Errorf("IsResourceDisabled(%v) = false, want true", o)
```

pkg/utils/apiresources_test.go
```diff
@@ -228,144 +228,196 @@ func TestResourceConfigGroupParse(t *testing.T) {
...
+				}
 			}
-		}
+
+			for i, o := range test.enabled {
+				if ok := r.IsResourceDisabled(o); ok {
+					t.Errorf("%d: expected resource to be enabled : %v", i, o)
```
zhiying-lin, line 393:
```suggestion
					t.Errorf("IsResourceDisabled(%v) = true, want false", o)
```

pkg/utils/apiresources_test.go
```diff
@@ -228,144 +228,196 @@ func TestResourceConfigGroupParse(t *testing.T) {
...
+		})
 	}
 }
 
-func TestDefaultDisabledResourceConfigGroupVersionKindParse(t *testing.T) {
+func TestResourceConfigIsEmpty(t *testing.T) {
 	tests := []struct {
-		disabled []schema.GroupVersionKind
-		enabled  []schema.GroupVersionKind
+		input string
```
zhiying-lin, line 402: could you please add a test name?

`In most tests, your test failure messages should include the function inputs if they are short. If the relevant properties of the inputs are not obvious (for example, because the inputs are large or opaque), you should name your test cases with a description of what’s being tested, and print the description as part of your error message.`

https://go.dev/wiki/TestComments#identify-the-input

pkg/utils/apiresources_test.go
```diff
@@ -228,144 +228,196 @@ func TestResourceConfigGroupParse(t *testing.T) {
...
+			input: "v1/Node,Pod;networking.k8s.io;apps/v1;authorization.k8s.io/v1/SelfSubjectRulesReview",
+			want:  false,
 		},
 	}
 	for _, test := range tests {
-		r := NewDisabledResourceConfig()
-		if err := r.Parse(""); err != nil {
+		r := NewResourceConfig(true)
+		if err := r.Parse(test.input); err != nil {
 			t.Fatalf("Unexpected error: %v", err)
```
zhiying-lin, line 417: t.Fatalf("Parse() returned error: %v", err)

pkg/utils/apiresources_test.go
```diff
@@ -228,144 +228,196 @@ func TestResourceConfigGroupParse(t *testing.T) {
...
-				t.Errorf("%d: unexpected error: %v", i, o)
-			}
-		}
-		for i, o := range test.enabled {
-			ok := r.IsResourceDisabled(o)
-			if ok {
-				t.Errorf("%d: unexpected error: %v", i, o)
-			}
+		if got := r.IsEmpty(); got != test.want {
+			t.Errorf("Unexpected result: %v", got)
```
zhiying-lin, line 420:
```suggestion
			t.Errorf("IsEmpty() = %v, want %v", got, test.want)
```

pkg/utils/apiresources_test.go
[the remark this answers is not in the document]
zhiying-lin: LGTM

pkg/utils/apiresources_test.go
vasudev-bongale: I ended up refactoring all the tests in apiresources_test.go, I have added small utility functions to make tests more readable and not repeating code multiple times. please let me know if that looks good.

pkg/controllers/clusterresourceplacement/resource_selector.go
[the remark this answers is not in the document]
vasudev-bongale: Thanks for the explanation! I have kept the original nil check since `r.ResourceConfig.IsResourceDisabled(gvk) should return correctly if the list is empty.`

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -287,9 +287,9 @@ func (r *Reconciler) fetchAllResourcesInOneNamespace(namespaceName string, place
 	return resources, nil
 }
 
-// shouldSelectResource returns whether a resource should be propagated
+// shouldSelectResource returns whether a resource should be selected for propagation.
 func (r *Reconciler) shouldSelectResource(gvr schema.GroupVersionResource) bool {
-	if r.DisabledResourceConfig == nil {
+	if r.ResourceConfig == nil {
```
zhiying-lin, line 292:
```suggestion
       // By default, all of the APIs are allowed.
	if r.ResourceConfig == nil {
```

nit, add a comment to improve the readability

pkg/resourcewatcher/change_dector.go
```diff
@@ -186,7 +186,7 @@ func (d *ChangeDetector) discoverResources(dynamicResourceEventHandler cache.Res
 
 // gvrDisabled returns whether GroupVersionResource is disabled.
 func (d *ChangeDetector) shouldWatchResource(gvr schema.GroupVersionResource) bool {
-	if d.DisabledResourceConfig == nil {
+	if d.ResourceConfig == nil {
```
zhiying-lin, line 189:
```suggestion
         // By default, all of the APIs are allowed.
	if d.ResourceConfig == nil {
```

pkg/controllers/clusterresourceplacement/placement_controllerv1alpha1.go
```diff
@@ -49,7 +49,7 @@ type Reconciler struct {
 	// It's only needed by v1beta1 APIs.
 	UncachedReader client.Reader
 
-	// ResourceConfig contains all the api resources that we won't select based on allowed or skipped propagating apis option.
+	// ResourceConfig contains all the API resources that we won't select based on allowed or skipped propagating APIs ption.
```
zhiying-lin, line 52:
```suggestion
	// ResourceConfig contains all the API resources that we won't select based on allowed or skipped propagating APIs option.
```

zhiying-lin (approved):

PR #642, opened 2023-12-19
feat: create multiple resource snapshots for same group index

Arvindthiru:

Fixes #

I have:

This PR contains core logic and UTs

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -438,9 +438,9 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
 	}
 
 	shouldCreateNewMasterClusterSnapshot := true
-	// This index indicates the index in the split selectedResourceList, if this index is zero we start from creating the
-	// master clusterResourceSnapshot if it's greater than zero it means that the master clusterResourceSnapshot got created
-	// but not all sub-indexed clusterResourceSnapshots have been created yet.
+	// This index indicates the selected resource in the split selectedResourceList, if this index is zero we start
+	// from creating the master clusterResourceSnapshot if it's greater than zero it means that the master clusterResourceSnapshot
+	// got created but not all sub-indexed clusterResourceSnapshots have been created yet.
 	resourceSnapshotStartIndex := 0
 	if latestResourceSnapshot != nil && latestResourceSnapshotHash == resourceHash {
 		if err := r.ensureLatestResourceSnapshot(ctx, latestResourceSnapshot); err != nil {
@@ -494,16 +494,13 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
 	}
 	// split selected resources as list of lists.
 	selectedResourcesList := splitSelectedResources(resourceSnapshotSpec.SelectedResources)
-	// create master cluster resource snapshot and sub-indexed resource snapshots belonging to the same group index.
-	resourceSnapshotSubIndex := -1
 	var resourceSnapshot *fleetv1beta1.ClusterResourceSnapshot
 	for i := resourceSnapshotStartIndex; i < len(selectedResourcesList); i++ {
 		if i == 0 {
 			resourceSnapshot = buildMasterClusterResourceSnapshot(latestResourceSnapshotIndex, len(selectedResourcesList), envelopeObjCount, crp.Name, resourceHash, selectedResourcesList[i])
 			latestResourceSnapshot = resourceSnapshot
 		} else {
-			resourceSnapshotSubIndex++
-			resourceSnapshot = buildSubIndexResourceSnapshot(latestResourceSnapshotIndex, resourceSnapshotSubIndex, crp.Name, selectedResourcesList[i])
+			resourceSnapshot = buildSubIndexResourceSnapshot(latestResourceSnapshotIndex, i-1, crp.Name, selectedResourcesList[i])
 		}
 		if err = r.createResourceSnapshot(ctx, crp, resourceSnapshot); err != nil {
 			return nil, err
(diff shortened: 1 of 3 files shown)


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -34,6 +34,8 @@ import (
 	"go.goms.io/fleet/pkg/utils/labels"
 )
 
+const resourceSnapshotSize int = 1024 * (1 << 10) // 1 MB
```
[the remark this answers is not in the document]
zhiying-lin: I'm thinking of using 80% of the 1MB as the limit to leave some buffers for the other fields of the resourcesnapshot.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -455,38 +478,113 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
 		return nil, err
 	}
 
-	// create a new resource snapshot
-	latestResourceSnapshotIndex++
-	latestResourceSnapshot = &fleetv1beta1.ClusterResourceSnapshot{
+	// only increment if new cluster resource snapshots are to be created.
```
zhiying-lin, line 481: The solution should work for the positive case, but the perf is not good for the negative case when the controller crashes in the middle. And there's a bug when handling the redundant resource snapshots.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -455,38 +478,113 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
...
+		if err = r.createResourceSnapshot(ctx, crp, resourceSnapshot); err != nil {
+			return nil, err
+		}
+	}
+	return latestResourceSnapshot, nil
+}
+
+// splitSelectedResources splits selected resources in a ClusterResourcePlacement into separate lists
+// so that the total size of each split list of selected Resources is within 1MB limit.
+func splitSelectedResources(selectedResources []fleetv1beta1.ResourceContent) [][]fleetv1beta1.ResourceContent {
```
zhiying-lin, line 528: we should add UT for this important function.

pkg/controllers/clusterresourceplacement/controller.go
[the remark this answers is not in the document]
Arvindthiru: The actual allowed size for a k8s object 1.5 MB I capped it at 1MB

pkg/controllers/clusterresourceplacement/controller.go
Arvindthiru: As discussed offline opened a separate PR to add UTs https://github.com/Azure/fleet/pull/652

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -1671,17 +1702,520 @@ func TestGetOrCreateClusterResourceSnapshot(t *testing.T) {
...
+						Annotations: map[string]string{
+							fleetv1beta1.SubindexOfResourceSnapshotAnnotation: "0",
+						},
+					},
+					Spec: *resourceSnapshotSpecSplit2,
 				},
 			},
 			wantLatestSnapshotIndex: 1,
 		},
+		{
```
zhiying-lin, line 2109: please create two more test cases:

- selected resource cross clusterResourceSnapshot size limit, all clusterResourceSnapshots remain the same since no change, but the IsLatestSnapshotLabel of current latest snapshot is set as false
- selected resources cross clusterResourceSnapshot limit, revision limit is 1, delete existing clusterResourceSnapshots which are missing subindex snapshots & create new clusterResourceSnapshots

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -38,14 +38,15 @@ const (
 
 var (
 	fleetAPIVersion = fleetv1beta1.GroupVersion.String()
-	cmpOptions      = []cmp.Option{
+	sortOption      = cmpopts.SortSlices(func(r1, r2 fleetv1beta1.ClusterResourceSnapshot) bool {
```
zhiying-lin, line 41:
```suggestion
	sortClusterResourceSnapshotOption      = cmpopts.SortSlices(func(r1, r2 fleetv1beta1.ClusterResourceSnapshot) bool {
```

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -506,7 +507,8 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
 			return nil, err
 		}
 	}
-	if len(selectedResourcesList) == 0 {
+	// shouldCreateNewMasterClusterSnapshot is used here to be defensive and catch regression.
```
zhiying-lin, line 510:
```suggestion
	// shouldCreateNewMasterClusterSnapshot is used here to be defensive in case of the regression.
```

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -1126,38 +1127,68 @@ func secretResourceContentForTest(t *testing.T) *fleetv1beta1.ResourceContent {
...
+	resourceSnapshotSpecWithTwoResourcesHash := fmt.Sprintf("%x", sha256.Sum256(jsonBytes))
+	resourceSnapshotSpecSplit1 := resourceSnapshotSpecWithSingleResource
+	resourceSnapshotSpecSplit2 := &fleetv1beta1.ResourceSnapshotSpec{
+		SelectedResources: []fleetv1beta1.ResourceContent{secretResourceContent},
+	}
+	resourceSnapshotSpecSplit3 := &fleetv1beta1.ResourceSnapshotSpec{
+		SelectedResources: []fleetv1beta1.ResourceContent{deploymentResourceContent},
+	}
+	resourceSnapshotSpecSplit4 := &fleetv1beta1.ResourceSnapshotSpec{
+		SelectedResources: []fleetv1beta1.ResourceContent{serviceResourceContent, secretResourceContent},
```
zhiying-lin, line 1177: forget to post this comment, please use these values directly instead of creating a variable. From the variable name, it's hard to tell what's the usage of them.

zhiying-lin (approved):

PR #646, opened 2023-12-21
interface: define override apis

zhiying-lin:

Introduced a cluster scope resource "Override" and "ResourceOverride".

Fixes #

I have:


apis/placement/v1beta1/commons.go
```diff
@@ -81,3 +83,36 @@ const (
...
+	// Use an empty string to select resources under the core API group (e.g., services).
+	// +required
+	Group string `json:"group"`
+
+	// Version of the namespace-scoped resource.
+	// +required
+	Version string `json:"version"`
+
+	// Kind of the namespace-scoped resource.
+	// Note: When `Kind` is `namespace`, ALL the resources under the selected namespaces are selected.
```
circy9, line 100: remove the note

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,125 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// The desired state of ClusterResourcePlacement.
+	// +required
+	Spec OverrideSpec `json:"spec"`
+}
+
+// OverrideSpec defines the desired state of the Override.
+type OverrideSpec struct {
```
circy9, line 30: Better to make overridepolicy part of a placement to avoid creating N-M mappings between overrideds & placements.

You can think a placement as an application. Usually overrides are part of the application configuration.

They need to managed by the same developer. As an owner of the app, I definitely don't want others to arbitrarily override my app.

This is also how kubefed does it.

circy9: Something like:

```
PlacementSpec {
  ...
  OverridePolicies []OverridePolicy 
}

OverridePolicy {
    // Target clusters.
    ClusterNames []string
    (or) ClusterSelector

    // Target resources.
    ClusterResourceSelectors []ClusterResourceSelector
    ResourceSelectors []ResourceSelector

    // Overrides.
    JSONPatchOverrides []JSONPatchOverride
  }
```

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -477,6 +477,15 @@ const (
 	// (i.e., fleet-member-<member-name>) on the hub cluster yet.
 	ClusterResourcePlacementSynchronizedConditionType ClusterResourcePlacementConditionType = "ClusterResourcePlacementSynchronized"
 
+	// ClusterResourcePlacementOverriddenConditionType indicates whether all the selected resources have been overridden
+	// successfully before applying to the target cluster if there is any override defined.
+	// Its condition status can be one of the following:
+	// - "True" means all the selected resources are successfully overridden before applying to the target cluster or
+	// override is not needed if there is no override defined.
+	// - "False" means some of them have failed. We will place some detailed failure in the FailedResourcePlacement array.
+	// - "Unknown" means we haven't finished the override yet.
```
ryanzhang-oss, line 486: for a newly added condition, there is also a case of "not set". I assume that the default is "no override needed"?

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// ApplyMode defines how to apply resources if the resources exist on the target clusters.
+	// It can be "CreateOnly", "ServerSideApply" or "ServerSideApplyWithForceConflicts". Default is CreateOnly.
+	// +kubebuilder:validation:Enum=CreateOnly;ServerSideApply;ServerSideApplyWithForceConflicts
+	// +kubebuilder:default=CreateOnly
+	// +optional
+	ApplyMode ApplyMode `json:"applyMode,omitempty"`
+
+	// Priority is an integer defining the relative importance of this override compared to others.
+	// Lower number is considered higher priority.
```
ryanzhang-oss, line 63: any reason that lower means higher priority？

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,125 @@
...
+// OverrideSpec defines the desired state of the Override.
+type OverrideSpec struct {
+	// If none of the ClusterResourceSelectors and ResourceSelectors are specified, it means selecting all resources.
+	// The ClusterResourceSelectors and ResourceSelectors are `ORed`.
+
+	// ClusterResourceSelectors is an array of selectors used to select cluster scoped resources. The selectors are `ORed`.
+	// If a namespace is selected, ALL the resources under the namespace are selected automatically.
+	// You can have 1-100 selectors.
+	// +optional
+	ClusterResourceSelectors []ClusterResourceSelector `json:"clusterResourceSelectors,omitempty"`
```
ryanzhang-oss, line 38: does it still make sense to apply the override to all the resources in a namespace?

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,125 @@
...
+	// OverridePolices defines how to override the selected resources.
+	// +optional
+	OverridePolices *OverridePolices `json:"overridePolices,omitempty"`
+
+	// ApplyMode defines how to apply resources if the resources exist on the target clusters.
+	// It can be "CreateOnly", "ServerSideApply" or "ServerSideApplyWithForceConflicts". Default is CreateOnly.
+	// +kubebuilder:validation:Enum=CreateOnly;ServerSideApply;ServerSideApplyWithForceConflicts
+	// +kubebuilder:default=CreateOnly
+	// +optional
+	ApplyMode ApplyMode `json:"applyMode,omitempty"`
```
ryanzhang-oss, line 60: this needs work-api support

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Priority is an integer defining the relative importance of this override compared to others.
+	// Lower number is considered higher priority.
+	// And resources will be applied by order from lower priority to higher.
+	// That means override with the lowest value will win.
```
ryanzhang-oss, line 65: you mean lowest priority value?

apis/placement/v1beta1/work_types.go
```diff
@@ -63,6 +68,23 @@ type Manifest struct {
...
 }
 
+type WorkloadOverride struct {
+	// Identifier represents an identity of a resource which has defined the overrides.
+	// +required
+	Identifier WorkResourceIdentifier `json:"identifier"`
+
+	// Overrides defines a list of override applied to the resource before applying to the spoke cluster.
+	// +required
+	Overrides []Override `json:"overrides"`
```
ryanzhang-oss, line 78: I can't seem to find where "Override" is defined

apis/placement/v1beta1/work_types.go
```diff
@@ -47,6 +49,9 @@ const (
 type WorkSpec struct {
 	// Workload represents the manifest workload to be deployed on spoke cluster
 	Workload WorkloadTemplate `json:"workload,omitempty"`
+
+	// WorkloadOverrides represents a list of overrides applied to the resources.
+	WorkloadOverrides []WorkloadOverride `json:"workloadOverrides,omitempty"`
```
ryanzhang-oss, line 54: why do we need this? Which component is going to apply the overrides?

apis/placement/v1beta1/override_types.go
zhiying-lin: I have discussed this option with Ryan and Patrick. We take the current option because of:

- there is no relationship between CRP and override. The override is applying to the resources(s).
- the apply result is determined by the resource itself. The value won't be changed if multiple CRPs select the same resource but use different override policies.

apis/placement/v1beta1/clusterresourceplacement_types.go
zhiying-lin: do you mean the resource itself does not have override policy? you want to distinguish it with a separate condition?

In the current api design, if no override needed, the ClusterResourcePlacementOverridden condition will be set as true.

apis/placement/v1beta1/override_types.go
zhiying-lin: In the azure networking world, ["Azure default security rules are given the highest number with the lowest priority to ensure that custom rules are always processed first."](https://learn.microsoft.com/en-us/azure/virtual-network/network-security-groups-overview#security-rules)

Similarly, [the azure firewall policy](https://learn.microsoft.com/en-us/azure/firewall/rule-processing) follows the same pattern. "Rules are processed based on Rule Collection Group Priority and Rule Collection priority. Priority is any number between 100 (highest priority) to 65,000 (lowest priority). Highest priority Rule Collection Groups are processed first. Inside a rule collection group, Rule Collections with highest priority (lowest number) are processed first."

apis/placement/v1beta1/override_types.go
zhiying-lin: The main reason is to simplify the duplicate override configuration. For example, a user want to add a label to all of the resources under a namespace. In addition, we can keep it consistent with the CRP semantics.

apis/placement/v1beta1/override_types.go
zhiying-lin: yes

apis/placement/v1beta1/work_types.go
zhiying-lin: Right now hubagent will apply the override and then create the manifest. memberAgent need the applyMode info to know how to apply the resource, eg, createOnly or server-side apply.

In addition, this field provides the troubleshooting information.

apis/placement/v1beta1/work_types.go
zhiying-lin: It's the full structure defined in the override_types.go.

I was thinking of populating the object metadata, overridePolicies, applyMode and priority fields for the troubleshooting purpose.
It may be too much and could exceed the 1mb limit of the object. 

Now I'm leaning toward creating a new struct to include the object name and its observedGeneration.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -99,8 +99,30 @@ type ClusterResourcePlacementSpec struct {
 	// +kubebuilder:default=10
 	// +optional
 	RevisionHistoryLimit *int32 `json:"revisionHistoryLimit,omitempty"`
+
+	// ApplyMode defines how to apply resources if the resources exist on the target clusters.
+	// It can be "CreateOnly", "ServerSideApply" or "ServerSideApplyWithForceConflicts". Default is CreateOnly.
+	// +kubebuilder:validation:Enum=CreateOnly;ServerSideApply;ServerSideApplyWithForceConflicts
+	// +kubebuilder:default=CreateOnly
+	// +optional
+	ApplyMode ApplyMode `json:"applyMode,omitempty"`
```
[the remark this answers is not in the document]
ryanzhang-oss: Can this be folded into rollout strategy instead of a first level parameter?

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -99,8 +99,30 @@ type ClusterResourcePlacementSpec struct {
...
+	// ApplyModeCreateOnly creates the resources on the target clusters and will fail if the resources exist on the target
+	// cluster.
+	ApplyModeCreateOnly ApplyMode = "CreateOnly"
+	// ApplyModeServerSideApply applies the resources on the target cluster using server-side apply.
+	// https://kubernetes.io/docs/reference/using-api/server-side-apply/
+	ApplyModeServerSideApply ApplyMode = "ServerSideApply"
+	// ApplyModeServerSideApplyWithForceConflicts applies the resources on the target cluster using server-side apply with
+	// force-conflicts option. It forces the operation to succeed when resolving the conflicts.
+	ApplyModeServerSideApplyWithForceConflicts ApplyMode = "ServerSideApplyWithForceConflicts"
+)
```
ryanzhang-oss, line 124: this is essentially expressing two different controls
1. server side apply or not
2. force apply or not
I wonder if it would make more sense to expand this single mode into a struct so we can add more options in the future?

apis/placement/v1beta1/resourcesnapshot_types.go
```diff
@@ -82,6 +82,8 @@ type ResourceSnapshotSpec struct {
 	// SelectedResources contains a list of resources selected by ResourceSelectors.
 	// +required
 	SelectedResources []ResourceContent `json:"selectedResources"`
+
+	ApplyMode ApplyMode `json:"applyMode,omitempty"`
```
ryanzhang-oss, line 86: this is where the actual apply mode can be folded into each resource.

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,157 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package v1beta1
```
ryanzhang-oss, line 6: new apis normally don't start as v1beta1

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,157 @@
...
+	ClusterSelector *ClusterSelector `json:"clusterSelector,omitempty"`
+
+	// OverridePolices defines how to override the selected resources.
+	// +optional
+	OverridePolices *OverridePolices `json:"overridePolices,omitempty"`
+
+	// ApplyMode defines how to apply resources if the resources exist on the target clusters.
+	// It can be "CreateOnly", "ServerSideApply" or "ServerSideApplyWithForceConflicts". Default is CreateOnly.
+	// +kubebuilder:validation:Enum=CreateOnly;ServerSideApply;ServerSideApplyWithForceConflicts
+	// +kubebuilder:default=CreateOnly
```
ryanzhang-oss, line 51: remove now?

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,157 @@
...
+	// +required
+	Spec ClusterResourceOverrideSpec `json:"spec"`
+}
+
+// ClusterResourceOverrideSpec defines the desired state of the Override.
+type ClusterResourceOverrideSpec struct {
+	// ClusterResourceSelectors is an array of selectors used to select cluster scoped resources. The selectors are `ORed`.
+	// If a namespace is selected, ALL the resources under the namespace are selected automatically.
+	// You can have 1-20 selectors.
+	// We only support Name selector for now.
```
ryanzhang-oss, line 33: why only Name selector?

apis/placement/v1beta1/overridesnapshot_types.go
```diff
@@ -0,0 +1,80 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// The desired state of ClusterResourceOverrideSnapshotSpec.
+	// +required
+	Spec ClusterResourceOverrideSnapshotSpec `json:"spec"`
+}
+
+type ClusterResourceOverrideSnapshotSpec struct {
+	OverrideName string `json:"overrideName"`
```
ryanzhang-oss, line 37: comment on this field? Is this different from the label value "ClusterResourceOverrideTrackingLabel"?

apis/placement/v1beta1/overridesnapshot_types.go
```diff
@@ -0,0 +1,80 @@
...
+*/
+
+package v1beta1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope="Cluster",shortName=rss,categories={fleet,fleet-placement}
```
ryanzhang-oss, line 13: what should be the shortname?

apis/placement/v1beta1/overridesnapshot_types.go
```diff
@@ -0,0 +1,80 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:JSONPath=`.metadata.generation`,name="Gen",type=string
```
ryanzhang-oss, line 15: do we need this? Is it immutable?

apis/placement/v1beta1/overridesnapshot_types.go
```diff
@@ -0,0 +1,80 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ClusterResourceOverrideSnapshot is used to store a snapshot of ClusterResourceOverride.
+// Its spec is immutable.
+// We assign an ever-increasing index for snapshots.
+// The naming convention of a ClusterResourceOverrideSnapshot is {ClusterResourceOverride}-{resourceIndex}.
+// resourceIndex will begin with 0.
+// Each snapshot MUST have the following labels:
+//   - `ClusterResourceOverrideTrackingLabel` which points to its owner ClusterResourceOverride.
+//   - `IsLatestSnapshotLabel` which indicates whether the snapshot is the latest one.
```
ryanzhang-oss, line 26: do we need a sub-index for multiple snapshot?

apis/placement/v1beta1/override_types.go
```diff
@@ -0,0 +1,157 @@
...
+// ClusterResourceOverride defines a group of override policies about how to override the selected resources and how to apply these resources
+// to the target clusters.
+// Note: override may fail, and it will be reflected on the placement status.
+type ClusterResourceOverride struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// The desired state of ClusterResourceOverrideSpec.
+	// +required
+	Spec ClusterResourceOverrideSpec `json:"spec"`
```
ryanzhang-oss, line 25: Is there any case we want a status field?

apis/placement/v1beta1/overriddenresourcesnapshot_types.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// The observed status of ResourceSnapshot.
+	// +optional
+	Status OverriddenResourceSnapshotStatus `json:"status,omitempty"`
+}
+
+type OverriddenResourceSnapshotSpec struct {
+	TargetCluster string `json:"targetCluster"`
+
+	ClusterResourceSnapshotName string `json:"clusterResourceSnapshotName"`
```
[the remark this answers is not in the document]
ryanzhang-oss: add comments?

apis/placement/v1beta1/overriddenresourcesnapshot_types.go
```diff
@@ -0,0 +1,90 @@
...
+	// +required
+	Spec OverriddenResourceSnapshotSpec `json:"spec"`
+
+	// The observed status of ResourceSnapshot.
+	// +optional
+	Status OverriddenResourceSnapshotStatus `json:"status,omitempty"`
+}
+
+type OverriddenResourceSnapshotSpec struct {
+	TargetCluster string `json:"targetCluster"`
```
ryanzhang-oss, line 51: add comments?

apis/placement/v1beta1/overriddenresourcesnapshot_types.go
```diff
@@ -0,0 +1,90 @@
...
+
+	ClusterResourceSnapshotName string `json:"clusterResourceSnapshotName"`
+
+	ResourceOverrides []OverrideSnapshotIdentifier `json:"resourceOverrides"`
```
ryanzhang-oss, line 55: Have we decided to not have priority for now? Are the elements in the array sorted by anything or just random?

apis/placement/v1beta1/overriddenresourcesnapshot_types.go
```diff
@@ -0,0 +1,90 @@
...
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// NumberOfSnapshots indicates how many snapshots are created within the same group and generated from the same clusterResourceSnapshot
+	// and its overrides.
+	// +optional
+	// +kubebuilder:default=1
+	NumberOfSnapshots int `json:"numberOfSnapshots,omitempty"`
+
+	// Workload represents the manifest workload to be deployed on spoke cluster.
+	Workload WorkloadTemplate `json:"workload,omitempty"`
```
ryanzhang-oss, line 80: AFAIK, most of the fields in the status are actually known when the CR object is created instead of having another controller reconcile based on the spec. Hence, I wonder why those are in the status? I can see condition be part of status just to keep the convention.

ryanzhang-oss: Or should we actually use the work generator to reconcile this object which makes the flow a lot more natural? I know the problem is that we won't be able to "keep the last known good state" but I am not sure if that's what a state driven system normally guarantees.

apis/placement/v1beta1/overriddenresourcesnapshot_types.go
ryanzhang-oss: btw, What exactly is this field for?

zhiying-lin:
Created two separate PRs instead

PR #652, opened 2024-01-04
feat: handle multiple resource snapshots - split selected resources into lists

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -34,6 +34,10 @@ import (
 	"go.goms.io/fleet/pkg/utils/labels"
 )
 
+// The max size of an object in k8s is 1.5MB because of ETCD limit https://etcd.io/docs/v3.3/dev-guide/limit/.
+// We choose 1MB as the max size for all the selected resources within one clusterResourceSnapshot object.
+var resourceSnapshotResourceSizeLimit = 1024 * (1 << 10) // 1 MB
+
 func (r *Reconciler) Reconcile(ctx context.Context, key controller.QueueKey) (ctrl.Result, error) {
 	name, ok := key.(string)
 	if !ok {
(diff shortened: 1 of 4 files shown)


pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -2017,6 +2027,44 @@ func TestGetOrCreateClusterResourceSnapshot_failure(t *testing.T) {
 	}
 }
 
+func TestSplitSelectedResources(t *testing.T) {
```
zhiying-lin, line 2030: As discussed offline, the current UT is not good enough, we need to test different resource size combinations.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -34,6 +34,8 @@ import (
 	"go.goms.io/fleet/pkg/utils/labels"
 )
 
+var resourceSnapshotSize = 1024 * (1 << 10) // 1 MB
```
zhiying-lin, line 37: Add a comment to explain why we set the limit as 1MB and a comment to explain why we need this.
```suggestion
var resourceSnapshotResourceSizeLimit = 1024 * (1 << 10) // 1 MB
```

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -2017,6 +2079,57 @@ func TestGetOrCreateClusterResourceSnapshot_failure(t *testing.T) {
...
+		{
+			name:                       "selected resources cross individual clusterResourceSnapshot size limit - one resource in first list, two resources in second list",
+			selectedResourcesSizeLimit: 600,
+			selectedResources:          []fleetv1beta1.ResourceContent{serviceResourceContent, deploymentResourceContent, secretResourceContent},
+			wantSplitSelectedResources: [][]fleetv1beta1.ResourceContent{{serviceResourceContent}, {deploymentResourceContent, secretResourceContent}},
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			resourceSnapshotResourceSizeLimit = tc.selectedResourcesSizeLimit
```
zhiying-lin, line 2124: we need to revert the value to the original value at the end.

```suggestion
        originalResourceSnapshotResourceSizeLimit := resourceSnapshotResourceSizeLimit
        defer func() {
           resourceSnapshotResourceSizeLimit = originalResourceSnapshotResourceSizeLimit 
        }
	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			resourceSnapshotResourceSizeLimit = tc.selectedResourcesSizeLimit
```

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -2017,6 +2079,57 @@ func TestGetOrCreateClusterResourceSnapshot_failure(t *testing.T) {
 	}
 }
 
+func TestSplitSelectedResources(t *testing.T) {
+	serviceResourceContent := *serviceResourceContentForTest(t)
```
zhiying-lin, line 2083: could you please write a comment to indicate the size of content of the resource? easier for reader to understand what you're testing.

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -2017,6 +2079,57 @@ func TestGetOrCreateClusterResourceSnapshot_failure(t *testing.T) {
...
+	deploymentResourceContent := *deploymentResourceContentForTest(t)
+	secretResourceContent := *secretResourceContentForTest(t)
+	tests := []struct {
+		name                       string
+		selectedResourcesSizeLimit int
+		selectedResources          []fleetv1beta1.ResourceContent
+		wantSplitSelectedResources [][]fleetv1beta1.ResourceContent
+	}{
+		{
```
zhiying-lin, line 2092: please add a test case when a resource content size is greater than the limit already.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -489,6 +493,32 @@ func (r *Reconciler) getOrCreateClusterResourceSnapshot(ctx context.Context, crp
...
+		var snapshotResources []fleetv1beta1.ResourceContent
+		for j < len(selectedResources) {
+			currentSize += len(selectedResources[j].Raw)
+			if currentSize > resourceSnapshotResourceSizeLimit {
+				break
+			}
+			snapshotResources = append(snapshotResources, selectedResources[j])
+			j++
+		}
+		if len(snapshotResources) == 0 {
```
zhiying-lin, line 513: This assumption is not correct.

Our limit is 1mb, but it's possible that a valid resource is more than 1mb as the etcd size limit is 1.5mb.

We should handle this valid case.

zhiying-lin (approved):

PR #654, opened 2024-01-09
fix: fix flaky rollout controller integration test

britaniar:

Fixes #

I have: Fixed flaky rollout controller integration test

### DIFF
--- a/pkg/controllers/rollout/controller_integration_test.go
+++ b/pkg/controllers/rollout/controller_integration_test.go
@@ -542,7 +542,7 @@ func markBindingApplied(binding *fleetv1beta1.ClusterResourceBinding, success bo
 		if err := k8sClient.Status().Update(ctx, binding); err != nil {
 			if apierrors.IsConflict(err) {
 				// get the binding again to avoid conflict
-				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)).Should(Succeed())
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.GetName()}, binding)).Should(Succeed())
 			}
 			return err
 		}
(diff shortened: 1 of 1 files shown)


pkg/controllers/rollout/controller_integration_test.go
```diff
@@ -542,7 +542,7 @@ func markBindingApplied(binding *fleetv1beta1.ClusterResourceBinding, success bo
 		if err := k8sClient.Status().Update(ctx, binding); err != nil {
 			if apierrors.IsConflict(err) {
 				// get the binding again to avoid conflict
-				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)).Should(Succeed())
+				Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.GetName()}, binding)).Should(Succeed())
```
zhiying-lin, line 545: Both .GetName() and .Name are correct.

pkg/controllers/rollout/controller_integration_test.go
```diff
@@ -177,6 +177,10 @@ var _ = Describe("Test the rollout Controller", func() {
 			By(fmt.Sprintf("resource binding `%s` is marked as not scheduled", binding.Name))
 			deletedBindings = append(deletedBindings, binding)
 		}
+		// simulate that some of the bindings are applied
```
zhiying-lin, line 180: should we move the for loop before we set them as unscheduled?
Could you please add a comment to indicate why we move these statement?

otherwise, the rollout controller will try to delete the bindings before we mark them as applied.

zhiying-lin (commented):
Could you please update the PR title to indicate which test you're fixing?

zhiying-lin (approved):

PR #657, opened 2024-01-16
fix: upgrade upstream test env to 1.28.0

britaniar:
upgrade test env to 1.28

Fixes #

I have:


Arvindthiru:
I think the ask was to update the k8s version to 1.28 https://github.com/Azure/fleet/blob/main/Makefile#L47, not the kind version.

zhiying-lin:
> I think the ask was to update the k8s version to 1.28 https://github.com/Azure/fleet/blob/main/Makefile#L47, not the kind version.

We actually need both, one is for the integration test and the other is for the e2e test env (which is using kind).

zhiying-lin (approved):

PR #658, opened 2024-01-19
fix: fix flaky internalmembercluster contoller integration test

britaniar:

Fixes #

I have: made the RequeAfter comparison inclusive to include all +/- jitter percent in internal member cluster controller integration test.

### DIFF
--- a/pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
+++ b/pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
@@ -6,8 +6,6 @@ package v1beta1
 
 import (
 	"context"
-	"fmt"
-	"k8s.io/klog/v2"
 	"strings"
 	"time"
 
@@ -36,7 +34,7 @@ var _ = Describe("Test Internal Member Cluster Controller", func() {
 
 	BeforeEach(func() {
 		ctx = context.Background()
-		HBPeriod = int(200)
+		HBPeriod = int(utils.RandSecureInt(600))
 		memberClusterName = "rand-" + strings.ToLower(utils.RandStr()) + "-mc"
 		memberClusterNamespace = "fleet-" + memberClusterName
 		memberClusterNamespacedName = types.NamespacedName{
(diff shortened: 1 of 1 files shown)


pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
```diff
@@ -110,8 +110,8 @@ var _ = Describe("Test Internal Member Cluster Controller", func() {
 				NamespacedName: memberClusterNamespacedName,
 			})
 			// take into account the +- jitter
-			Expect(result.RequeueAfter.Milliseconds() < (1000+1000*jitterPercent/2/100)*time.Millisecond.Milliseconds()*int64(HBPeriod)).Should(BeTrue())
-			Expect(result.RequeueAfter.Milliseconds() > (1000-1000*jitterPercent/2/100)*time.Millisecond.Milliseconds()*int64(HBPeriod)).Should(BeTrue())
+			Expect(result.RequeueAfter.Milliseconds() <= (1000+1000*jitterPercent/2/100)*time.Millisecond.Milliseconds()*int64(HBPeriod)).Should(BeTrue())
```
zhiying-lin, line 113: I'm not 100 % sure if this change will help to reduce the flakiness a lot.

Suggest to update the jitter large, eg, 15?

Could you please also add a error message to print the actual result?

```suggestion
upperBoundOfWantRequeueAfter = (1000+1000*jitterPercent/2/100)*time.Millisecond.Milliseconds()*int64(HBPeriod)
			Expect(result.RequeueAfter.Milliseconds() <= upperBoundOfWantRequeue).Should(BeTrue(), "Reconcile() RequeueAfter got %v, want <= %v", result.RequeueAfter, upperBoundOfWantRequeueAfter)
```

pkg/controllers/internalmembercluster/v1beta1/member_controller.go
```diff
@@ -56,8 +56,8 @@ const (
 	// EventReasonInternalMemberClusterLeft is the event type and reason string when the agent left.
 	EventReasonInternalMemberClusterLeft = "InternalMemberClusterLeft"
 
-	// we add +-5% jitter
-	jitterPercent = 10
+	// we add +-7.5% jitter
+	jitterPercent = 15
```
britaniar, line 60: @zhiying-lin  increased the jitter here, is +-7.5% or do we want a more well-rounded number where it is +-10%?

zhiying-lin: i feel 10% is too much or you can round to 7%, using 14 instead.

Let's start with a small number. You'll know the diff after you add the log.

zhiying-lin (approved):
LGTM, leave you to decide jitterPercent.

pkg/controllers/internalmembercluster/v1beta1/member_controller.go
```diff
@@ -56,8 +56,8 @@ const (
 	// EventReasonInternalMemberClusterLeft is the event type and reason string when the agent left.
 	EventReasonInternalMemberClusterLeft = "InternalMemberClusterLeft"
 
-	// we add +-5% jitter
-	jitterPercent = 10
+	// we add +-7% jitter
+	jitterPercent = 14
```
ryanzhang-oss, line 60: I don't get why increasing the jitter percent is going to help this case.

zhiying-lin: https://github.com/Azure/fleet/actions/runs/7564476912/job/20598646789 the failure happened when we compared the requeueAfter. Relaxing the range should help. Adding error message should help us to figure out what's the diff.

zhiying-lin (dismissed):

pkg/controllers/internalmembercluster/v1beta1/member_controller.go
ryanzhang-oss: had a sync, this should not matter. We don't need to change this.

zhiying-lin (approved):

PR #659, opened 2024-01-25
fix: Changes in correlation with new GH Action Permission Changes.

Tatsinnit:
These changes are made in correlation with the following changes coming soon in the GH actions permissions. Please take a kind look. ❤️🙏☕️

https://docs.opensource.microsoft.com/github/apps/permission-changes/

Key aspects from this change referential is:

> Changes are coming that will break your GitHub Actions workflows unless you add a [permissions block](https://docs.opensource.microsoft.com/github/apps/permission-changes/#what-steps-must-i-take-to-avoid-a-bad-time). You can add this today to minimize impact to your group or your GitHub Actions may fail starting on 2024-02-01.

Followed by:

> In your GHA workflow, you must add a [permissions block](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions) to your workflow file that will require you to specify the scope and permission you are granting it to your repo. The options for access are read, write and none.

If this helps please feel free to check and this might help in being prepared for the changes coming soon from 1st Feb.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,9 @@ env:
 jobs:
   detect-noop:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
     outputs:
       noop: ${{ steps.noop.outputs.should_skip }}
     steps:
@@ -32,6 +35,9 @@ jobs:
   unit-tests:
     runs-on: ubuntu-latest
     needs: detect-noop
+    permissions:
+      contents: read
+      pull-requests: read
     if: needs.detect-noop.outputs.noop != 'true'
     steps:
       - name: Set up Go
@@ -59,6 +65,9 @@ jobs:
   
   e2e-tests-v1alpha1:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
     needs: [
       detect-noop,
     ]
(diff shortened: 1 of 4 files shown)


ryanzhang-oss:
>> Starting February 1, 2024 the default permission for the GITHUB_TOKEN will change from Read/Write to Read-only for all our Open Source GitHub orgs. This is a breaking change to many workflows.

I thought the token will become read-only if we don't do anything. This PR doesn't seem to add any write permissions. Did I miss anything?

Tatsinnit:
> > > Starting February 1, 2024 the default permission for the GITHUB_TOKEN will change from Read/Write to Read-only for all our Open Source GitHub orgs. This is a breaking change to many workflows.
> 
> I thought the token will become read-only if we don't do anything. This PR doesn't seem to add any write permissions. Did I miss anything?

Thank you @ryanzhang-oss yeah, you can leave this PR open, but here is the reason where I think it must be compulsory grade;ly post 1st Feb, this comment below makes me think that the `permission` tag will be in need, be it all empty or `read-all` et. al. 

I have highlighted the **MUST have** statement :) 

Screenshot 2024-01-26 at 10 36 39 AM

ryanzhang-oss:
stale

PR #661, opened 2024-01-26
interface: define override apis

zhiying-lin:

add new override apis

Fixes #

I have:


apis/placement/v1alpah1/override_types.go
```diff
@@ -0,0 +1,186 @@
...
+	Version string `json:"version"`
+
+	// Kind of the namespace-scoped resource.
+	// +required
+	Kind string `json:"kind"`
+
+	// Namespace of the namespace-scoped resource.
+	// +optional
+	// Default is empty, which means inherit from the parent object scope.
+	Namespace string `json:"namespace,omitempty"`
```
ryanzhang-oss, line 128: the resourceOverride is already a namespaced object, we don't need a namespace field here. Aka, we should not allow a resourceOverride object in namespace A says that it will override some object in namespace B.

apis/placement/v1alpah1/override_types.go
```diff
@@ -0,0 +1,186 @@
...
+	Policy *OverridePolicy `json:"policy"`
+}
+
+// OverridePolicy defines how to override the selected resources on the target clusters.
+// More is to be added.
+type OverridePolicy struct {
+	// OverrideRules defines an array of override rules to be applied on the selected resources.
+	// The order of the rules determines the override order.
+	// When there are two rules selecting the same fields on the target cluster, the last one will win.
+	// You can have 1-100 rules.
```
ryanzhang-oss, line 54: we only need 100 rules if we don't have templated override rules (i.e. append the cluster name to the field), right?

apis/placement/v1alpah1/override_types.go
zhiying-lin: SG. The namespace is redundant. Will remove.

apis/placement/v1alpah1/override_types.go
```diff
@@ -0,0 +1,186 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package v1alpah1
```
circy9, line 6: Does this mean you will do beta later?

apis/placement/v1alpah1/override_types.go
```diff
@@ -0,0 +1,186 @@
...
+	// Default is empty, which means inherit from the parent object scope.
+	Namespace string `json:"namespace,omitempty"`
+
+	// Name of the namespace-scoped resource.
+	// +required
+	Name string `json:"name"`
+}
+
+// JSONPatchOverride applies a JSON patch on the selected resources following [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902).
+type JSONPatchOverride struct {
```
circy9, line 136: Could you provide a few examples for the type as comments?
Like: change a field, add one item to an array, ...

apis/placement/v1alpah1/override_types.go
```diff
@@ -0,0 +1,186 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// The desired state of ClusterResourceOverrideSpec.
+	// +required
+	Spec ClusterResourceOverrideSpec `json:"spec"`
+}
+
+// ClusterResourceOverrideSpec defines the desired state of the Override.
+type ClusterResourceOverrideSpec struct {
```
circy9, line 31: Discuss the conflict-resolving behavior here?

Conflict with another CRO. Conflict with RO.

circy9: Same for RO.

apis/placement/v1alpah1/override_types.go
zhiying-lin: yes when we GA launch placement API.

apis/placement/v1alpah1/override_types.go
circy9: Could you also add what happens if one resource is selected by two CRO?

apis/placement/v1alpah1/override_types.go
```diff
@@ -0,0 +1,233 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// The desired state of ResourceOverrideSpec.
+	// +required
+	Spec ResourceOverrideSpec `json:"spec"`
+}
+
+// ResourceOverrideSpec defines the desired state of the Override.
+// If the resource is selected by both ClusterResourceOverride and ResourceOverride, ResourceOverride will win when resolving
+// conflicts.
```
circy9, line 98: Could you also add what happens if one resource is selected by two RO?

circy9 (changes requested):

circy9 (approved):

PR #665, opened 2024-02-01
feat: add draft API for metric-based scheduling

michaelawyu:

This PR adds the draft API for metric-based scheduling.

Fixes #

I have:

N/A

The PR is for early discussion preparation only; all names, fields, structures are not finalized. Any code inconsistencies are also not yet addressed.

### DIFF
--- a/apis/cluster/v1beta1/internalmembercluster_types.go
+++ b/apis/cluster/v1beta1/internalmembercluster_types.go
@@ -48,6 +48,13 @@ type InternalMemberClusterSpec struct {
 
 // InternalMemberClusterStatus defines the observed state of InternalMemberCluster.
 type InternalMemberClusterStatus struct {
+	// Conditions is an array of current observed conditions for the member cluster.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// Metrics is an array of non-resource metrics observed for the member cluster.
+	Metrics []Metric `json:"metrics,omitempty"`
+
 	// The current observed resource usage of the member cluster. It is populated by the member agent.
 	// +optional
 	ResourceUsage ResourceUsage `json:"resourceUsage,omitempty"`
--- a/apis/cluster/v1beta1/membercluster_types.go
+++ b/apis/cluster/v1beta1/membercluster_types.go
@@ -49,6 +49,14 @@ type MemberClusterSpec struct {
 	HeartbeatPeriodSeconds int32 `json:"heartbeatPeriodSeconds,omitempty"`
 }
 
+// Metric is the observed value of a non-resource metric.
+type Metric struct {
+	// The name of the metric; it should be a valid Kubernetes label name.
+	Name string `json:"name"`
+	// The observed value of the metric; it should be a sortable numeric.
+	Value string `json:"value"`
+}
+
 // MemberClusterStatus defines the observed status of MemberCluster.
 type MemberClusterStatus struct {
 	// +patchMergeKey=type
@@ -59,6 +67,9 @@ type MemberClusterStatus struct {
 	// Conditions is an array of current observed conditions for the member cluster.
 	// +optional
 	Conditions []metav1.Condition `json:"conditions"`
+
+	// Metrics is an array of non-resource metrics observed for the member cluster.
+	Metrics []Metric `json:"metrics,omitempty"`
 
 	// The current observed resource usage of the member cluster. It is copied from the corresponding InternalMemberCluster object.
 	// +optional
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -239,9 +239,9 @@ type MetricMatcher struct {
 	// Name is the name of the metric; it should be a Kubernetes label name.
 	// +required
 	Name string `json:"name"`
-	// Ranges are the required or preferred ranges for the metric.
+	// Range is the required or preferred range for the metric.
 	// +required
-	Ranges []MetricRange `json:"ranges"`
+	Range MetricRange `json:"ranges"`
 }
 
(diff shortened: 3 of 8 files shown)


apis/cluster/v1beta1/internalmembercluster_types.go
```diff
@@ -48,6 +48,13 @@ type InternalMemberClusterSpec struct {
 
 // InternalMemberClusterStatus defines the observed state of InternalMemberCluster.
 type InternalMemberClusterStatus struct {
+	// Conditions is an array of current observed conditions for the member cluster.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
```
ryanzhang-oss, line 53: what are the conditions we need to add?

apis/cluster/v1beta1/membercluster_types.go
```diff
@@ -49,6 +49,14 @@ type MemberClusterSpec struct {
 	HeartbeatPeriodSeconds int32 `json:"heartbeatPeriodSeconds,omitempty"`
 }
 
+// Metric is the observed value of a non-resource metric.
+type Metric struct {
+	// The name of the metric; it should be a valid Kubernetes label name.
```
ryanzhang-oss, line 54: do we need a metrics type? for example, static vs dynamic? integer vs fraction?

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -218,11 +218,51 @@ type PreferredClusterSelector struct {
...
+
+// MetricRange is one metric requirement when picking clusters for resource placement.
+type MetricMatcher struct {
+	// Name is the name of the metric; it should be a Kubernetes label name.
+	// +required
+	Name string `json:"name"`
+	// Ranges are the required or preferred ranges for the metric.
+	// +required
+	Ranges []MetricRange `json:"ranges"`
+}
```
ryanzhang-oss, line 245: How do we express that we want to find a cluster with a metrics key with the largest/smallest value using this?

pkg/metricprovider/interface.go
```diff
@@ -0,0 +1,36 @@
...
+
+// MetricCollectionResponse is returned by a Fleet metric provider to report metrics and
+// metric collection status.
+type MetricCollectionResponse struct {
+	// Metrics is an array of metrics and their values. The key should be the name of the
+	// metric, which is a Kubernetes label name; the value is the metric data.
+	Metrics map[string]float64
+	// Resources is a group of resources, described by their allocatable capacity and
+	// available capacity.
+	Resources clusterv1beta1.ResourceUsage
```
ryanzhang-oss, line 25: I am not sure if I get how to combine these two together.

ryanzhang-oss: and how can this be translated to internalMemberCluster fields

apis/placement/v1beta1/clusterresourceplacement_types.go
michaelawyu: Yeah, that's actually a question I would like to ask before adding the APIs:

Earlier, it has been proposed that we put metric selectors as a part of the existing cluster selector terms, i.e. a cluster selector term = some label selectors + some metric selectors. This makes great sense as it reduces confusion and achieves the use case where user might want to filter out/prefer clusters with both specific labels and specific metrics with the right values. But it also brings out a side effect: the weight is an upper level field for a cluster selector term, i.e.,

```yaml
type PreferredClusterSelector struct {
	Weight int32 `json:"weight"`
	Preference ClusterSelectorTerm `json:"preference"`
}
```

michaelawyu: Basically there are two complications:

* User might want to express prefer maximum/minimum semantics using left/right open brackets, e.g., [10, max.) or (min, 30). This would normally require two weights (one for the min/max. value and one for the set limit), yet currently there's only one weight to specify. We could either add another weight field, which makes the API a bit weird if the user only uses label selectors; or we could assume that the weight for the set limit is always 0 ([10, max) -> cluster with 10 gets a score of 0, cluster with max gets the full weight), but this is also less than satisfactory as cluster with 10 receives the same score of 0 as clusters with a value of the range.
* We do allow multiple range expressions for a metric. The weight field being in the upper level means that all ranges have to use the same weight, with or without prefer min/max semantics.

michaelawyu: The most straightforward solution is just to add a interpolate from min/max switch, and assume that a) one end of the range always receives the weight of 0 with prefer min./max. semantics and b) all range expressions of a metric use the same weight -> would this be OK?

apis/cluster/v1beta1/membercluster_types.go
michaelawyu: The static vs dynamic distinction might make less sense IMO as we don't really handle them differently (assuming that static metrics = metrics that don't change very often?). 

As for the int vs float distinction, Kubernetes does not recommend using floats directly in the API (as different programming languages support them differently); for obvious reasons, we will always need float values, so here all values are serialized into a string as recommended, which should work just fine for both ints and floats.

apis/cluster/v1beta1/internalmembercluster_types.go
michaelawyu: This field is for the metric provider so it can pass some conditions regarding the metric collection process.

pkg/metricprovider/interface.go
michaelawyu: The two are not combined; non-resource and resource metrics are kept separately. The metrics returned here are put in the `metrics` field in the `InternalMemberCluster` status (with the value being serialized as a string); the resources go to the `resourceUsage` field in the status.

michaelawyu:
Closing the PR due to latest developments; will send a new one

PR #670, opened 2024-02-05
fix: address previous comments

ryanzhang-oss:

Add some comments and return wrapped error.

Fixes #

I have:

Existing UT

### DIFF
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -679,9 +679,9 @@ func (f *framework) createBindings(ctx context.Context, toCreate []*placementv1b
 							// The binding already exists, which is fine.
 							return nil
 						}
-						klog.ErrorS(err, "failed to create a new binding", "clusterResourceBinding", klog.KObj(newBinding))
+						klog.ErrorS(err, "Failed to create a new binding", "clusterResourceBinding", klog.KObj(newBinding))
 					}
-					return err
+					return controller.NewAPIServerError(false, err)
 				})
 		})
 	}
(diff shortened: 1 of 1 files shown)


pkg/scheduler/framework/framework.go
```diff
@@ -679,9 +679,9 @@ func (f *framework) createBindings(ctx context.Context, toCreate []*placementv1b
 							// The binding already exists, which is fine.
 							return nil
 						}
-						klog.ErrorS(err, "failed to create a new binding", "clusterResourceBinding", klog.KObj(newBinding))
+						klog.ErrorS(err, "Failed to create a new binding", "clusterResourceBinding", klog.KObj(newBinding))
 					}
-					return err
+					return controller.NewAPIServerError(false, err)
```
zhiying-lin, line 684: we cannot wrap this error here, as the retry will need to use the original api error. Suggest to wrap the error in line 688.

pkg/scheduler/framework/framework.go
```diff
@@ -697,15 +697,16 @@ func (f *framework) patchBindings(ctx context.Context, toPatch []*bindingWithPat
...
 				func() error {
-					// Use JSON patch to avoid races so we shouldn't get a conflict error.
+					// we will get conflict error if the binding has been updated.
 					err := f.client.Patch(cctx, patchBinding.updated, patchBinding.patch)
 					if err != nil {
-						klog.ErrorS(err, "failed to patch a binding", "clusterResourceBinding", klog.KObj(patchBinding.updated))
+						klog.ErrorS(err, "Failed to patch a binding", "clusterResourceBinding", klog.KObj(patchBinding.updated))
 					}
-					return err
+					return controller.NewUpdateIgnoreConflictError(err)
```
zhiying-lin, line 709: ditto

pkg/scheduler/framework/framework.go
```diff
@@ -697,19 +697,20 @@ func (f *framework) patchBindings(ctx context.Context, toPatch []*bindingWithPat
 		errs.Go(func() error {
 			return retry.OnError(retry.DefaultBackoff,
 				func(err error) bool {
+					// we can't retry on conflict errors here easily as we need to fetch the original binding to patch it.
 					return apierrors.IsServiceUnavailable(err) || apierrors.IsServerTimeout(err)
 				},
 				func() error {
-					// Use JSON patch to avoid races so we shouldn't get a conflict error.
+					// we will get conflict error if the binding has been updated.
```
zhiying-lin, line 704:
```suggestion
					// We will get conflict error if the binding has been updated.
```

pkg/scheduler/framework/framework.go
```diff
@@ -697,19 +697,20 @@ func (f *framework) patchBindings(ctx context.Context, toPatch []*bindingWithPat
 		errs.Go(func() error {
 			return retry.OnError(retry.DefaultBackoff,
 				func(err error) bool {
+					// we can't retry on conflict errors here easily as we need to fetch the original binding to patch it.
```
zhiying-lin, line 700:
```suggestion
					// We can't retry on conflict errors here easily as we need to fetch the original binding to patch it.
```

zhiying-lin (approved):

PR #671, opened 2024-02-05
perf: increase controller concurrency

ryanzhang-oss:

Add an option to increase the concurrency of the controller downstream of scheduler.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -20,9 +20,10 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
-	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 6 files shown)


pkg/controllers/rollout/controller.go
```diff
@@ -20,9 +20,10 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
-	ctrl "sigs.k8s.io/controller-runtime"
+	controllerRuntime "sigs.k8s.io/controller-runtime"
```
zhiying-lin, line 23: By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps. 

rename to "runtime" or "controllerruntime"?

michaelawyu (approved):
LGTM

zhiying-lin (approved):

PR #674, opened 2024-02-06
feat: add scheduler related metrics

michaelawyu:

This PR adds a scheduler metric and enables the queue metrics for the scheduler workqueue.

I have:

- [x] Unit tests

N/A


pkg/metrics/metrics.go
```diff
@@ -50,3 +50,22 @@ var (
...
+// The scheduler related metrics.
+var (
+	// SchedulingCycleDurationMilliseconds is a Fleet scheduler metric that tracks how long it
+	// takes to complete a scheduling loop run.
+	SchedulingCycleDurationMilliseconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "scheduling_cycle_duration_milliseconds",
+			Help: "The duration of a scheduling cycle run in milliseconds",
+			Buckets: []float64{
+				10, 50, 100, 500, 1000, 5000, 10000, 50000,
```
ryanzhang-oss, line 63: any reason to use the this bucket size?

ryanzhang-oss (approved):

pkg/metrics/metrics.go
michaelawyu: Oh, the unit is milliseconds and based on prev. observations these buckets seem to be appropriate => would you prefer finer bucket sizes?

ryanzhang-oss: how about copy controller-runtime one?

michaelawyu: Oh, the controller-runtime one uses custom ExponetialBuckets (`prometheus.ExponentialBuckets(10e-9, 10, 10)`) and the unit is in seconds -> I can send a separate PR if you'd like to switch to seconds as well

michaelawyu:
I'll merge it first to unblock the progress; if there's any concern and further requested change, please let me know.

PR #682, opened 2024-02-13
feat: cluster resource override controller

ryanzhang-oss:

The main code  logic for cluster resource override controller. Most of the code changes are related to refactoring the code to have a common code for CRO and RO.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/overrider/clusterresource_controller.go
+++ b/pkg/controllers/overrider/clusterresource_controller.go
@@ -64,11 +64,9 @@ func (r *ClusterResourceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	// Ensure that we have the finalizer so we can delete all the related snapshots on cleanup
-	needUpdate, err := r.ensureFinalizer(ctx, &clusterOverride)
-	if needUpdate {
-		if err != nil {
-			klog.ErrorS(err, "Failed to ensure the finalizer", "clusterResourceOverride", overrideRef)
-		}
+	err := r.ensureFinalizer(ctx, &clusterOverride)
+	if err != nil {
+		klog.ErrorS(err, "Failed to ensure the finalizer", "clusterResourceOverride", overrideRef)
 		return ctrl.Result{}, err
 	}
 
--- a/pkg/controllers/overrider/clusterresource_controller.go
+++ b/pkg/controllers/overrider/clusterresource_controller.go
@@ -64,8 +64,7 @@ func (r *ClusterResourceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	// Ensure that we have the finalizer so we can delete all the related snapshots on cleanup
-	err := r.ensureFinalizer(ctx, &clusterOverride)
-	if err != nil {
+	if err := r.ensureFinalizer(ctx, &clusterOverride); err != nil {
 		klog.ErrorS(err, "Failed to ensure the finalizer", "clusterResourceOverride", overrideRef)
 		return ctrl.Result{}, err
 	}
--- a/pkg/controllers/overrider/clusterresource_controller.go
+++ b/pkg/controllers/overrider/clusterresource_controller.go
@@ -1,0 +1,161 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+// Package overrider features controllers to reconcile the override objects.
+package overrider
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	placementv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
+	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/utils"
+	"go.goms.io/fleet/pkg/utils/controller"
+	"go.goms.io/fleet/pkg/utils/labels"
+)
+
+// Reconciler reconciles a clusterResourceOverride object.
+type ClusterResourceReconciler struct {
+	Reconciler
+	recorder record.EventRecorder
+}
+
+// Reconcile triggers a single  reconcile round when scheduling policy has changed.
+func (r *ClusterResourceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	name := req.NamespacedName
+	clusterOverride := placementv1alpha1.ClusterResourceOverride{}
+	overrideRef := klog.KRef(name.Namespace, name.Name)
+
+	startTime := time.Now()
+	klog.V(2).InfoS("Reconciliation starts", "clusterResourceOverride", overrideRef)
+	defer func() {
+		latency := time.Since(startTime).Milliseconds()
+		klog.V(2).InfoS("Reconciliation ends", "clusterResourceOverride", overrideRef, "latency", latency)
+	}()
+
+	if err := r.Client.Get(ctx, name, &clusterOverride); err != nil {
+		if errors.IsNotFound(err) {
+			klog.V(4).InfoS("Ignoring notFound clusterResourceOverride", "clusterResourceOverride", overrideRef)
+			return ctrl.Result{}, nil
+		}
+		klog.ErrorS(err, "Failed to get clusterResourceOverride", "clusterResourceOverride", overrideRef)
+		return ctrl.Result{}, controller.NewAPIServerError(true, err)
+	}
+
+	// Check if the clusterResourceOverride is being deleted
+	if clusterOverride.DeletionTimestamp != nil {
+		klog.V(4).InfoS("The clusterResourceOverride is being deleted", "clusterResourceOverride", overrideRef)
+		return ctrl.Result{}, r.handleOverrideDeleting(ctx, &placementv1alpha1.ClusterResourceOverrideSnapshot{}, &clusterOverride)
+	}
+
+	// Ensure that we have the finalizer so we can delete all the related snapshots on cleanup
+	needUpdate, err := r.ensureFinalizer(ctx, &clusterOverride)
+	if needUpdate {
+		if err != nil {
+			klog.ErrorS(err, "Failed to ensure the finalizer", "clusterResourceOverride", overrideRef)
+		}
+		return ctrl.Result{}, err
+	}
+
+	// create or update the overrideSnapshot
+	return ctrl.Result{}, r.ensureClusterResourceOverrideSnapshot(ctx, &clusterOverride, 10)
+}
+
+func (r *ClusterResourceReconciler) ensureClusterResourceOverrideSnapshot(ctx context.Context, cro *placementv1alpha1.ClusterResourceOverride, revisionHistoryLimit int) error {
+	croKObj := klog.KObj(cro)
+	overridePolicy := cro.Spec
+	overrideSpecHash, err := utils.GenerateSpecHash(overridePolicy)
+	if err != nil {
+		klog.ErrorS(err, "Failed to generate policy hash of clusterResourceOverride", "clusterResourceOverride", croKObj)
+		return controller.NewUnexpectedBehaviorError(err)
+	}
+	// we need to list the snapshots anyway since we need to remove the extra snapshots if there are too many of them.
+	snapshotList, err := r.listSortedOverrideSnapshots(ctx, utils.ClusterResourceOverrideSnapshotKind, cro.GetName())
+	if err != nil {
+		return err
+	}
+	// delete redundant snapshot revisions before creating a new snapshot to guarantee that the number of snapshots
+	// won't exceed the limit.
+	if err = r.removeExtraSnapshot(ctx, snapshotList, revisionHistoryLimit); err != nil {
+		return err
+	}
+
+	latestSnapshotIndex := 0
+	if len(snapshotList.Items) != 0 {
+		// convert the last unstructured snapshot to the typed object
+		latestSnapshot := &placementv1alpha1.ClusterResourceOverrideSnapshot{}
+		if err = runtime.DefaultUnstructuredConverter.FromUnstructured(snapshotList.Items[len(snapshotList.Items)-1].Object, latestSnapshot); err != nil {
+			klog.ErrorS(err, "Invalid overrideSnapshot", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(&snapshotList.Items[len(snapshotList.Items)-1]))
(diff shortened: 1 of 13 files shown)


cmd/hubagent/main.go
```diff
@@ -59,6 +60,7 @@ func init() {
 	utilruntime.Must(fleetv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(workv1alpha1.AddToScheme(scheme))
 	utilruntime.Must(placementv1beta1.AddToScheme(scheme))
+	utilruntime.Must(placementv1alpha1.AddToScheme(scheme))
```
zhiying-lin, line 63: should we not include the new apis in the main func? until the crd release has been finished.

cmd/hubagent/workload/setup.go
```diff
@@ -271,6 +272,16 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 			klog.ErrorS(err, "Unable to set up memberCluster watcher for scheduler")
 			return err
 		}
+
+		klog.Info("Setting up ClusterResourceOverride controller")
```
zhiying-lin, line 276: ditto, unless we're sure the crd release has been finished.

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -0,0 +1,161 @@
...
+		// we need to figure out the last snapshot index.
+		latestSnapshotIndex, err = labels.ExtractIndex(latestSnapshot, placementv1alpha1.OverrideIndexLabel)
+		if err != nil {
+			klog.ErrorS(err, "Failed to parse the override index label", "clusterResourceOverride", croKObj, "overrideSnapshot", klog.KObj(latestSnapshot))
+			return controller.NewUnexpectedBehaviorError(err)
+		}
+	}
+
+	// Need to create new snapshot when 1) there is no snapshots or 2) the latest snapshot hash != current one.
+	latestSnapshotIndex++
```
zhiying-lin, line 128: what if there is no snapshot? the latestSnapshotIndex becomes 1?

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -0,0 +1,161 @@
...
+	newSnapshot := &placementv1alpha1.ClusterResourceOverrideSnapshot{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf(placementv1alpha1.OverrideSnapshotNameFmt, cro.Name, latestSnapshotIndex),
+			Labels: map[string]string{
+				placementv1alpha1.OverrideTrackingLabel: cro.Name,
+				placementv1beta1.IsLatestSnapshotLabel:  strconv.FormatBool(true),
+				placementv1alpha1.OverrideIndexLabel:    strconv.Itoa(latestSnapshotIndex),
+			},
+			OwnerReferences: []metav1.OwnerReference{
```
zhiying-lin, line 137: nit, to set the owner reference, you can use the controllerutil.SetControllerReference() func

pkg/controllers/overrider/clusterresource_controller_integration_test.go
```diff
@@ -0,0 +1,170 @@
...
+				return err
+			}
+			if !controllerutil.ContainsFinalizer(cro, fleetv1alpha1.OverrideFinalizer) {
+				return fmt.Errorf("finalizer not added")
+			}
+			return nil
+		}, eventuallyTimeout, interval).Should(Succeed(), "snapshot should have finalizer")
+
+		By("Checking if a new snapshot is created")
+		snapshot := getClusterResourceOverrideSnapshot(testCROName, 1) //index start from 1
```
zhiying-lin, line 54: why index not starting from 0?

pkg/controllers/overrider/clusterresource_controller_integration_test.go
```diff
@@ -0,0 +1,170 @@
...
+		Eventually(func() error {
+			return k8sClient.Get(ctx, types.NamespacedName{Name: snapshot.Name}, snapshot)
+		}, eventuallyTimeout, interval).Should(Succeed(), "snapshot should exist")
+		By("Checking if the label is correct")
+		diff := cmp.Diff(map[string]string{
+			fleetv1alpha1.OverrideIndexLabel:    strconv.Itoa(1),
+			fleetv1beta1.IsLatestSnapshotLabel:  "true",
+			fleetv1alpha1.OverrideTrackingLabel: testCROName,
+		}, snapshot.GetLabels())
+		Expect(diff).Should(BeEmpty(), diff)
```
zhiying-lin, line 64:
```suggestion
		Expect(diff).Should(BeEmpty(), diff, "Snapshot label mismatch (-want, +got)")
```
need to include error message

pkg/controllers/overrider/clusterresource_controller_integration_test.go
```diff
@@ -0,0 +1,170 @@
...
+		By("Checking if the spec is correct")
+		diff = cmp.Diff(cro.Spec, snapshot.Spec.OverrideSpec)
+		Expect(diff).Should(BeEmpty(), diff)
```
zhiying-lin, line 67: ditto

pkg/controllers/overrider/common.go
```diff
@@ -0,0 +1,122 @@
...
+	if err := r.Client.DeleteAllOf(ctx, overrideSnapshotObj, client.MatchingLabels{placementv1alpha1.OverrideTrackingLabel: parentOverrideObj.GetName()}); err != nil {
+		klog.ErrorS(err, "Failed to delete all associated overrideSnapshot", "clusterResourceOverride", croObjRef)
+		return err
+	}
+	klog.V(2).InfoS("Deleted all overrideSnapshot associated with the override", "overrideSnapshot", klog.KObj(overrideSnapshotObj), "clusterResourceOverride", croObjRef)
+
+	controllerutil.RemoveFinalizer(parentOverrideObj, placementv1alpha1.OverrideFinalizer)
+	if err := r.Client.Update(ctx, parentOverrideObj); err != nil {
+		klog.ErrorS(err, "Failed to remove crp finalizer", "clusterResourceOverride", croObjRef)
+		return err
```
zhiying-lin, line 51: add error wrapper? NewUpdateIgnoreConflictError?

pkg/controllers/overrider/common.go
```diff
@@ -0,0 +1,122 @@
...
+func (r *Reconciler) handleOverrideDeleting(ctx context.Context, overrideSnapshotObj, parentOverrideObj client.Object) error {
+	croObjRef := klog.KObj(parentOverrideObj)
+	if !controllerutil.ContainsFinalizer(parentOverrideObj, placementv1alpha1.OverrideFinalizer) {
+		klog.V(2).InfoS("No need to do anything for the deleting override without a finalizer", "clusterResourceOverride", croObjRef)
+		return nil
+	}
+	// delete all the associated snapshots
+	if err := r.Client.DeleteAllOf(ctx, overrideSnapshotObj, client.MatchingLabels{placementv1alpha1.OverrideTrackingLabel: parentOverrideObj.GetName()}); err != nil {
+		klog.ErrorS(err, "Failed to delete all associated overrideSnapshot", "clusterResourceOverride", croObjRef)
+		return err
```
zhiying-lin, line 44: add the error wrapper?

what if the associated snapshot is 0, will delete fail?

pkg/controllers/overrider/common.go
```diff
@@ -0,0 +1,122 @@
...
+	})
+
+	if len(errs) > 0 {
+		return nil, controller.NewUnexpectedBehaviorError(utilerrors.NewAggregate(errs))
+	}
+
+	return snapshotList, nil
+}
+
+func (r *Reconciler) removeExtraSnapshotReconciler(ctx context.Context, sortedSnapshotList *unstructured.UnstructuredList, limit int) error {
```
zhiying-lin, line 96:
```suggestion
func (r *Reconciler) removeExtraSnapshot(ctx context.Context, sortedSnapshotList *unstructured.UnstructuredList, limit int) error {
```

pkg/controllers/overrider/common.go
```diff
@@ -0,0 +1,122 @@
...
+	// the list is sorted by the override index, so we can just remove from the beginning
+	for i := 0; i <= len(sortedSnapshotList.Items)-limit; i++ {
+		if err := r.Client.Delete(ctx, &sortedSnapshotList.Items[i]); err != nil {
+			klog.ErrorS(err, "Failed to delete the extra override snapshot", "overrideSnapshot", klog.KObj(&sortedSnapshotList.Items[i]))
+			return err
```
zhiying-lin, line 101: add error wrapper?

ignore when the error is not found?

pkg/utils/common.go
```diff
@@ -51,8 +54,9 @@ const (
 )
 
 const (
-	PlacementFieldManagerName    = "cluster-placement-controller"
-	MCControllerFieldManagerName = "member-cluster-controller"
+	PlacementFieldManagerName          = "cluster-placement-controller"
```
zhiying-lin, line 57: i'm thinking of getting rid of the utils package as the utils is not a good package name and it's too generic.

We can always create the package, like FieldManager, etc.

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -0,0 +1,161 @@
...
+)
+
+// Reconciler reconciles a clusterResourceOverride object.
+type ClusterResourceReconciler struct {
+	Reconciler
+	recorder record.EventRecorder
+}
+
+// Reconcile triggers a single  reconcile round when scheduling policy has changed.
+func (r *ClusterResourceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
zhiying-lin, line 38: we can probably create a common reconciler for both ro and cro. We can use interface to have different implementations, so that we don't need to duplicate the code. These two controllers are almost the same.

pkg/controllers/overrider/clusterresource_controller.go
ryanzhang-oss: I was trying to set all in one place.

pkg/controllers/overrider/clusterresource_controller_integration_test.go
ryanzhang-oss: isn't diff already the error message?

pkg/controllers/overrider/common_integration_test.go
```diff
@@ -0,0 +1,277 @@
...
+			By("Removing the overrideFinalizer")
+			controllerutil.RemoveFinalizer(cro, fleetv1alpha1.OverrideFinalizer)
+			Expect(commonReconciler.handleOverrideDeleting(ctx, nil, cro)).Should(Succeed())
+		})
+
+		It("Should not fail if there is no snapshots yet", func() {
+			By("verifying that it handles no snapshot cases")
+			cro.Name = "non-existing-cro" //there is no snapshot associated with this CRO
+			// we don't persist the CRO as it will trigger the real reconcile loop so the update can only return APIServerError
+			Expect(errors.Is(commonReconciler.handleOverrideDeleting(context.Background(), getClusterResourceOverrideSnapshot(testCROName, 0), cro), controller.ErrAPIServerError)).Should(BeTrue())
```
zhiying-lin, line 123: did not get why we expect the api server error

pkg/controllers/overrider/common_integration_test.go
```diff
@@ -0,0 +1,277 @@
...
+			},
+		},
+		Spec: fleetv1alpha1.ClusterResourceOverrideSnapshotSpec{
+			OverrideHash: []byte("hash"),
+			OverrideSpec: getClusterResourceOverrideSpec(),
+		},
+	}
+}
+
+var _ = Describe("Test ClusterResourceOverride common logic", func() {
```
zhiying-lin, line 89: The integration test looks weird as it manually triggered some of the funcs, instead of by creating/update/deleting the override and then validating the controller behavior.

It looks like unit tests.

pkg/controllers/overrider/clusterresource_controller_integration_test.go
zhiying-lin: yes, but the diff don't provide enough information, eg, which two objects? and what's the "-" and "+". it will be hard to debug the failure unless we read the code.

btw, If i remember it correctly, the expect will print the "diff" automatically.

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Check if the clusterResourceOverride is being deleted
+	if clusterOverride.DeletionTimestamp != nil {
+		klog.V(4).InfoS("The clusterResourceOverride is being deleted", "clusterResourceOverride", overrideRef)
+		return ctrl.Result{}, r.handleOverrideDeleting(ctx, &placementv1alpha1.ClusterResourceOverrideSnapshot{}, &clusterOverride)
+	}
+
+	// Ensure that we have the finalizer so we can delete all the related snapshots on cleanup
+	needUpdate, err := r.ensureFinalizer(ctx, &clusterOverride)
+	if needUpdate {
```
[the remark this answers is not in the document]
zhiying-lin: needUpdate is redundant, right?

pkg/controllers/overrider/clusterresource_controller.go
zhiying-lin: we can refactor the code later to see if we can find a clean way.

pkg/controllers/overrider/common_integration_test.go
ryanzhang-oss: yep, it's a UT with a real API server. The IT was for the whole reconciler.

ryanzhang-oss: I will see if renaming it as UT will break the test.

pkg/controllers/overrider/clusterresource_controller.go
ryanzhang-oss: The tricky part is the test, it will be a lot of redundant test too, that's why I stopped at the CRO IT.

michaelawyu (commented):
LGTM ;) I saw that this PR is also impacted by the dep bumps, so some changes might be necessary

pkg/controllers/overrider/common.go
zhiying-lin: Add the NewAPIServerError error wrapper?

pkg/controllers/overrider/clusterresource_controller.go
```diff
@@ -64,11 +64,9 @@ func (r *ClusterResourceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 
 	// Ensure that we have the finalizer so we can delete all the related snapshots on cleanup
-	needUpdate, err := r.ensureFinalizer(ctx, &clusterOverride)
-	if needUpdate {
-		if err != nil {
-			klog.ErrorS(err, "Failed to ensure the finalizer", "clusterResourceOverride", overrideRef)
-		}
+	err := r.ensureFinalizer(ctx, &clusterOverride)
+	if err != nil {
+		klog.ErrorS(err, "Failed to ensure the finalizer", "clusterResourceOverride", overrideRef)
```
zhiying-lin, line 69:
```suggestion
	if err := r.ensureFinalizer(ctx, &clusterOverride); err != nil {
		klog.ErrorS(err, "Failed to ensure the finalizer", "clusterResourceOverride", overrideRef)
```

zhiying-lin (commented):
others LGTM

zhiying-lin (approved):

PR #683, opened 2024-02-14
feat: taints & toleration plugin

Arvindthiru:

Fixes #

I have:

UTs

ITs and E2Es will be added in separate PRs

### DIFF
--- a/pkg/scheduler/framework/plugins/tainttoleration/filtering.go
+++ b/pkg/scheduler/framework/plugins/tainttoleration/filtering.go
@@ -22,7 +22,7 @@ func (p *Plugin) Filter(
 	policy *placementv1beta1.ClusterSchedulingPolicySnapshot,
 	cluster *clusterv1beta1.MemberCluster,
 ) (status *framework.Status) {
-	taint, isUntolerated := findUntoleratedTaint(cluster.Spec.Taints, policy.GetTolerations())
+	taint, isUntolerated := findUntoleratedTaint(cluster.Spec.Taints, policy.Tolerations())
 	if !isUntolerated {
 		return nil
 	}
--- a/pkg/scheduler/framework/plugins/tainttoleration/filtering.go
+++ b/pkg/scheduler/framework/plugins/tainttoleration/filtering.go
@@ -9,7 +9,6 @@ import (
 	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
 	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	"go.goms.io/fleet/pkg/scheduler/framework"
-	"go.goms.io/fleet/pkg/utils"
 )
 
 var (
@@ -23,7 +22,7 @@ func (p *Plugin) Filter(
 	policy *placementv1beta1.ClusterSchedulingPolicySnapshot,
 	cluster *clusterv1beta1.MemberCluster,
 ) (status *framework.Status) {
-	taint, isUntolerated := findUntoleratedTaint(cluster.Spec.Taints, utils.GetTolerations(policy.Spec.Policy))
+	taint, isUntolerated := findUntoleratedTaint(cluster.Spec.Taints, policy.GetTolerations())
 	if !isUntolerated {
 		return nil
 	}
--- a/pkg/scheduler/framework/plugins/tainttoleration/filtering.go
+++ b/pkg/scheduler/framework/plugins/tainttoleration/filtering.go
@@ -1,0 +1,63 @@
+package tainttoleration
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+
+	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
+	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/scheduler/framework"
+	"go.goms.io/fleet/pkg/utils"
+)
+
+var (
+	reasonFmt = "taint %+v cannot be tolerated"
+)
+
+// Filter allows the plugin to connect to the Filter extension point in the scheduling framework.
+func (p *Plugin) Filter(
+	_ context.Context,
+	_ framework.CycleStatePluginReadWriter,
+	policy *placementv1beta1.ClusterSchedulingPolicySnapshot,
+	cluster *clusterv1beta1.MemberCluster,
+) (status *framework.Status) {
+	taint, isUntolerated := findUntoleratedTaint(cluster.Spec.Taints, utils.GetTolerations(policy.Spec.Policy))
+	if !isUntolerated {
+		return nil
+	}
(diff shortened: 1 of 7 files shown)


pkg/scheduler/framework/framework.go
```diff
@@ -128,9 +135,13 @@ type frameworkOptions struct {
...
 	maxUnselectedClusterDecisionCount int
 
-	// checker is the cluster eligibility checker the scheduler framework will use to check
+	// clusterEligibilityChecker is the checker the scheduler framework will use to check
 	// if a cluster is eligibile for resource placement.
 	clusterEligibilityChecker *clustereligibilitychecker.ClusterEligibilityChecker
+
+	// taintTolerationChecker is the checker the scheduler framework will use to check
+	// if taints on a member cluster can be tolerated by tolerations on cluster resource placement.
+	taintTolerationChecker *tainttolerationchecker.TaintTolerationChecker
```
Arvindthiru, line 144: This checker will be used for PickFixed placement type similar to clusterEligibilityChecker where we don't run all the filter plugins

pkg/scheduler/framework/plugins/tainttoleration/filtering.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// Filter allows the plugin to connect to the Filter extension point in the scheduling framework.
+func (p *Plugin) Filter(
+	_ context.Context,
+	_ framework.CycleStatePluginReadWriter,
+	policy *placementv1beta1.ClusterSchedulingPolicySnapshot,
+	cluster *clusterv1beta1.MemberCluster,
+) (status *framework.Status) {
+	taint, isUntolerated := p.handle.TaintTolerationChecker().FindUnToleratedTaint(cluster.Spec.Taints, utils.GetTolerations(policy.Spec.Policy))
```
zhiying-lin, line 24: what if the cluster does not have any taint?

pkg/utils/common.go
```diff
@@ -391,3 +391,11 @@ func GenerateGroupString(groups []string) string {
 	}
 	return groupString
 }
+
+// GetTolerations returns tolerations from supplied placement policy.
+func GetTolerations(policy *placementv1beta1.PlacementPolicy) []placementv1beta1.Toleration {
```
zhiying-lin, line 396: prefer to move this util into apis/placement package, something like,

func (m *ClusterResourcePlacement) Tolerations() []Toleration  {
	....
}

similar to policySnapshot.

We need to get rid of the "utils" package cause the package name is not good and does not provide any meaning.

pkg/utils/test_util.go
```diff
@@ -136,3 +142,32 @@ func (matcher AlreadyExistMatcher) FailureMessage(actual interface{}) (message s
 func (matcher AlreadyExistMatcher) NegatedFailureMessage(actual interface{}) (message string) {
 	return format.Message(actual, "not to be already exist")
 }
+
+// MockHandle framework.Handle interface for set up the plugin.
+type MockHandle struct {
```
zhiying-lin, line 147: All the test util functions need to be moved to the test/utils/ path. We should not mix the test util with the non test utils in one package.

pkg/scheduler/framework/framework.go
Arvindthiru: Discussed offline taints and tolerations plugin will not be run for pickFixed since it's similar to specifying nodeName in podSpec

pkg/scheduler/framework/plugins/tainttoleration/filtering_test.go
```diff
@@ -0,0 +1,410 @@
...
+				Spec: placementv1beta1.SchedulingPolicySnapshotSpec{
+					Policy: &placementv1beta1.PlacementPolicy{
+						PlacementType: placementv1beta1.PickAllPlacementType,
+					},
+				},
+			},
+			wantStatus: framework.NewNonErrorStatus(framework.ClusterUnschedulable, p.Name(), fmt.Sprintf(reasonFmt, &clusterv1beta1.Taint{Key: "key1", Value: "value1", Effect: corev1.TaintEffectNoSchedule})),
+		},
+		{
+			name: "taints can be tolerated based on key, value & effect with Equal operator, every taint has matching toleration - nil status",
```
zhiying-lin, line 82: add some (positive and negative) tests when taint key is present + empty value + operation is equal.

pkg/scheduler/framework/plugins/tainttoleration/plugin.go
```diff
@@ -0,0 +1,64 @@
...
+
+// WithName sets the name of the plugin.
+func WithName(name string) Option {
+	return func(o *taintTolerationPluginOptions) {
+		o.name = name
+	}
+}
+
+// New returns a new Plugin.
+func New(opts ...Option) Plugin {
```
zhiying-lin, line 45: you will add this plugin in your next PR?

Arvindthiru: Yes

pkg/scheduler/framework/plugins/tainttoleration/filtering_test.go
```diff
@@ -171,7 +171,48 @@ func TestFilter(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "csp-1",
+				},
+				Spec: placementv1beta1.SchedulingPolicySnapshotSpec{
+					Policy: &placementv1beta1.PlacementPolicy{
+						PlacementType: placementv1beta1.PickAllPlacementType,
+						Tolerations: []placementv1beta1.Toleration{
+							{
+								Key:      "key1",
+								Operator: corev1.TolerationOpEqual,
```
zhiying-lin, line 202: it's not a valid test case? i remember this crp will be rejected by webhook when the value is empty?

pkg/scheduler/framework/plugins/tainttoleration/filtering_test.go
```diff
@@ -171,7 +171,48 @@ func TestFilter(t *testing.T) {
...
+							},
+							{
+								Key:      "key2",
```
zhiying-lin, line 205: add a test case when the operation is exist?

pkg/scheduler/framework/plugins/tainttoleration/filtering_test.go
```diff
@@ -0,0 +1,446 @@
...
+								Value:    "value2",
+							},
+						},
+					},
+				},
+			},
+			wantStatus: nil,
+		},
+		{
+			name: "taints can be tolerated based on key, empty value, effect with Equal operator, every taint has matching toleration - nil status",
```
zhiying-lin, line 174: add a negative test case for taints (key present + empty value + operation is equal)?

pkg/scheduler/framework/plugins/tainttoleration/filtering_test.go
Arvindthiru: closed a fix PR to address this https://github.com/Azure/fleet/pull/704 to align with k8s

PR #685, opened 2024-02-15
feat: validate tolerations

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -232,7 +232,9 @@ func validateTolerations(tolerations []placementv1beta1.Toleration) error {
 		if toleration.Value == "" && toleration.Operator == corev1.TolerationOpEqual {
 			return fmt.Errorf(invalidTolerationErrFmt, toleration, "toleration value cannot be empty, when operator is Equal")
 		}
-
+		if toleration.Value != "" && toleration.Operator == corev1.TolerationOpExists {
+			return fmt.Errorf(invalidTolerationErrFmt, toleration, "toleration value needs to be empty, when operator is Exists")
+		}
 		if _, exists := tolerationMap[toleration]; exists {
 			return fmt.Errorf(uniqueTolerationErrFmt, toleration)
 		}
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -9,6 +9,7 @@ package validator
 import (
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiErrors "k8s.io/apimachinery/pkg/util/errors"
@@ -113,22 +114,26 @@ func IsPlacementPolicyTypeUpdated(oldPolicy, currentPolicy *placementv1beta1.Pla
 }
 
 func validatePlacementPolicy(policy *placementv1beta1.PlacementPolicy) error {
+	allErr := make([]error, 0)
 	switch policy.PlacementType {
 	case placementv1beta1.PickFixedPlacementType:
 		if err := validatePolicyForPickFixedPlacementType(policy); err != nil {
-			return err
+			allErr = append(allErr, err)
 		}
 	case placementv1beta1.PickAllPlacementType:
 		if err := validatePolicyForPickAllPlacementType(policy); err != nil {
-			return err
+			allErr = append(allErr, err)
 		}
 	case placementv1beta1.PickNPlacementType:
 		if err := validatePolicyForPickNPolicyType(policy); err != nil {
-			return err
+			allErr = append(allErr, err)
 		}
 	}
+	if err := validateTolerations(policy.Tolerations); err != nil {
+		allErr = append(allErr, err)
+	}
 
-	return nil
+	return apiErrors.NewAggregate(allErr)
 }
 
 func validatePolicyForPickFixedPlacementType(policy *placementv1beta1.PlacementPolicy) error {
@@ -211,6 +216,32 @@ func validateClusterAffinity(clusterAffinity *placementv1beta1.ClusterAffinity, 
 		}
 	}
 	return apiErrors.NewAggregate(allErr)
+}
+
+func validateTolerations(tolerations []placementv1beta1.Toleration) error {
+	errFmt := "invalid toleration %+v"
+	allErr := make([]error, 0)
+	var tolerationMap map[string]placementv1beta1.Toleration
+	for _, toleration := range tolerations {
+		if toleration.Key == "" && toleration.Operator != corev1.TolerationOpExists {
+			allErr = append(allErr, fmt.Errorf(errFmt, toleration))
+		}
+		if toleration.Value == "" && toleration.Operator == corev1.TolerationOpEqual {
+			allErr = append(allErr, fmt.Errorf(errFmt, toleration))
+		}
+		key := buildTolerationKey(toleration)
+		_, exists := tolerationMap[key]
+		if !exists {
+			tolerationMap[key] = toleration
+		} else {
+			allErr = append(allErr, fmt.Errorf("tolerations must be unique"))
+		}
+	}
+	return apiErrors.NewAggregate(allErr)
+}
+
+func buildTolerationKey(toleration placementv1beta1.Toleration) string {
+	return toleration.Key + ":" + string(toleration.Operator) + ":" + toleration.Value + ":" + string(toleration.Effect)
 }
 
 func validateTopologySpreadConstraints(topologyConstraints []placementv1beta1.TopologySpreadConstraint) error {
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -24,7 +24,7 @@ import (
 var ResourceInformer informer.Manager
 
 var (
-	invalidTolerationErrFmt = "invalid toleration %+v"
+	invalidTolerationErrFmt = "invalid toleration %+v: %s"
 	uniqueTolerationErrFmt  = "toleration %+v already exists, tolerations must be unique"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/utils/validator/clusterresourceplacement_test.go
```diff
@@ -1036,3 +1040,78 @@ func TestIsPlacementPolicyUpdateValid(t *testing.T) {
...
+					Operator: corev1.TolerationOpExists,
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+				{
+					Operator: corev1.TolerationOpExists,
+				},
+			},
+			wantErr: nil,
+		},
+		"invalid toleration, key is empty, operator is Equal": {
```
zhiying-lin, line 1068: add a test case, operator is omitted.

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -213,6 +223,30 @@ func validateClusterAffinity(clusterAffinity *placementv1beta1.ClusterAffinity,
...
 
+func validateTolerations(tolerations []placementv1beta1.Toleration) error {
+	tolerationMap := make(map[string]placementv1beta1.Toleration)
+	for _, toleration := range tolerations {
+		if toleration.Key == "" && toleration.Operator != corev1.TolerationOpExists {
+			return fmt.Errorf(invalidTolerationErrFmt, toleration)
+		}
+		if toleration.Value == "" && toleration.Operator == corev1.TolerationOpEqual {
+			return fmt.Errorf(invalidTolerationErrFmt, toleration)
+		}
```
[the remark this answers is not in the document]
zhiying-lin: can value contain special characters?

pkg/utils/validator/clusterresourceplacement_test.go
zhiying-lin: add a test case, value contains special characters

pkg/utils/validator/clusterresourceplacement_test.go
```diff
@@ -1036,3 +1040,78 @@ func TestIsPlacementPolicyUpdateValid(t *testing.T) {
...
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			wantErr: fmt.Errorf(uniqueTolerationErrFmt, placementv1beta1.Toleration{Key: "key1", Operator: corev1.TolerationOpExists, Value: "value1", Effect: corev1.TaintEffectNoSchedule}),
+		},
+	}
+	for testName, testCase := range tests {
+		t.Run(testName, func(t *testing.T) {
+			actualErr := validateTolerations(testCase.tolerations)
+			assert.Equal(t, testCase.wantErr, actualErr, utils.TestCaseMsg, testName)
```
zhiying-lin, line 1114: please try to avoid to use the assert package for the new code.

you can always compare the result using ==

test/e2e/webhook_test.go
```diff
@@ -220,3 +221,99 @@ var _ = Describe("webhook tests for CRP UPDATE operations", Ordered, func() {
...
+
+		By("deleting created work resources")
+		cleanupWorkResources()
+	})
+
+	It("should deny update on CRP with invalid toleration", func() {
+		Eventually(func(g Gomega) error {
+			var crp placementv1beta1.ClusterResourcePlacement
+			g.Expect(hubClient.Get(ctx, types.NamespacedName{Name: crpName}, &crp)).Should(Succeed())
+			invalidToleration1 := placementv1beta1.Toleration{
```
zhiying-lin, line 272: one negative test case should be fine. others should be covered by the UT.

pkg/webhook/clusterresourceplacement/v1beta1_clusterresourceplacement_validating_webhook.go
```diff
@@ -42,6 +42,10 @@ func (v *clusterResourcePlacementValidator) Handle(_ context.Context, req admiss
 			klog.ErrorS(err, "failed to decode v1beta1 CRP object for create/update operation", "userName", req.UserInfo.Username, "groups", req.UserInfo.Groups)
 			return admission.Errored(http.StatusBadRequest, err)
 		}
+		if err := validator.ValidateClusterResourcePlacement(&crp); err != nil {
+			klog.V(2).InfoS("v1beta1 cluster resource placement has invalid fields, request is denied", "operation", req.Operation, "namespacedName", types.NamespacedName{Name: crp.Name})
+			return admission.Denied(err.Error())
+		}
```
ryanzhang-oss, line 48: I didn't get why we need to move this up

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -213,6 +223,30 @@ func validateClusterAffinity(clusterAffinity *placementv1beta1.ClusterAffinity,
...
+		key := buildTolerationKey(toleration)
+		_, exists := tolerationMap[key]
```
ryanzhang-oss, line 236: IIRC, we only need to know if the key exists or not, why store the toleration as the value?

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -213,6 +223,30 @@ func validateClusterAffinity(clusterAffinity *placementv1beta1.ClusterAffinity,
...
+			tolerationMap[key] = toleration
+		} else {
+			return fmt.Errorf(uniqueTolerationErrFmt, toleration)
+		}
+	}
+	return nil
+}
+
+func buildTolerationKey(toleration placementv1beta1.Toleration) string {
+	return toleration.Key + ":" + string(toleration.Operator) + ":" + toleration.Value + ":" + string(toleration.Effect)
```
ryanzhang-oss, line 247: is this string unique? there are seem to have many ways to get to this string.

pkg/utils/validator/clusterresourceplacement_test.go
Arvindthiru: If toleration operator is "" or TaintEffect is "" kubebuilder will complain
image

pkg/webhook/clusterresourceplacement/v1beta1_clusterresourceplacement_validating_webhook.go
Arvindthiru: This ensures that we validate the CRP regardless of the operation CREATE/UPDATE and then if it's an UPDATE we check to see if the update is valid

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -213,6 +225,44 @@ func validateClusterAffinity(clusterAffinity *placementv1beta1.ClusterAffinity,
...
 }
 
+func validateTolerations(tolerations []placementv1beta1.Toleration) error {
+	allErr := make([]error, 0)
+	tolerationMap := make(map[placementv1beta1.Toleration]bool)
+	for _, toleration := range tolerations {
+		switch toleration.Operator {
+		case corev1.TolerationOpExists:
+			if toleration.Key != "" {
+				for _, msg := range validation.IsQualifiedName(toleration.Key) {
```
Arvindthiru, line 235: Reusing validation logic for toleration key and value from k8s https://github.com/kubernetes/kubernetes/blob/7080b51ee92f67623757534f3462d8ae862ef6fe/pkg/apis/core/validation/validation.go#L3908-L3930

pkg/utils/validator/clusterresourceplacement_test.go
```diff
@@ -1036,3 +1038,122 @@ func TestIsPlacementPolicyUpdateValid(t *testing.T) {
...
+					Value:    "value1",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			wantErr: fmt.Errorf(uniqueTolerationErrFmt, placementv1beta1.Toleration{Key: "key1", Operator: corev1.TolerationOpEqual, Value: "value1", Effect: corev1.TaintEffectNoSchedule}),
+		},
+	}
+	for testName, testCase := range tests {
+		t.Run(testName, func(t *testing.T) {
+			if actualErr := validateTolerations(testCase.tolerations); actualErr != nil && actualErr.Error() != testCase.wantErr.Error() {
```
zhiying-lin, line 1154: note, if you compare the error equally, the test may be fragile as it relies on the underlay util and the error string could change anytime.

zhiying-lin: there will be a bug when the toleration is valid, but you did not compare the result.

zhiying-lin, line 1154:
```suggestion
			if got := validateTolerations(testCase.tolerations); actualErr != nil && actualErr.Error() != testCase.wantErr.Error() {
```

prefer to use "got" & "want" instead of "actual" & "expected"

" Where you would write “actual” and “expected”, prefer using the words “got” and “want”, respectively."

https://google.github.io/styleguide/go/decisions#got-before-want

zhiying-lin (changes requested):

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -213,6 +225,40 @@ func validateClusterAffinity(clusterAffinity *placementv1beta1.ClusterAffinity,
...
+			}
+			if toleration.Value == "" {
+				allErr = append(allErr, fmt.Errorf(invalidTolerationErrFmt, toleration, "toleration value cannot be empty, when operator is Equal"))
+			} else {
+				for _, msg := range validation.IsValidLabelValue(toleration.Value) {
+					allErr = append(allErr, fmt.Errorf(invalidTolerationValueErrFmt, toleration, msg))
+				}
+			}
+		}
+		if _, exists := tolerationMap[toleration]; exists {
```
ryanzhang-oss, line 254: nit, the idiomatic way to check a bool map is like this, 
```suggestion
		if tolerationMap[toleration]{
```

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -213,6 +225,40 @@ func validateClusterAffinity(clusterAffinity *placementv1beta1.ClusterAffinity,
...
+	allErr := make([]error, 0)
+	tolerationMap := make(map[placementv1beta1.Toleration]bool)
+	for _, toleration := range tolerations {
+		if toleration.Key != "" {
+			for _, msg := range validation.IsQualifiedName(toleration.Key) {
+				allErr = append(allErr, fmt.Errorf(invalidTolerationKeyErrFmt, toleration, msg))
+			}
+		}
+		switch toleration.Operator {
+		case corev1.TolerationOpExists:
```
ryanzhang-oss, line 238: what happens if the operator is none of the following?

pkg/utils/validator/clusterresourceplacement_test.go
```diff
@@ -1036,3 +1038,140 @@ func TestIsPlacementPolicyUpdateValid(t *testing.T) {
...
+					Key:      "key1",
+					Operator: corev1.TolerationOpEqual,
+					Value:    "val#%/-123",
+					Effect:   corev1.TaintEffectNoSchedule,
+				},
+			},
+			wantErr:    true,
+			wantErrMsg: "a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character",
+		},
+		"invalid toleration, key is invalid, operator is Exists": {
```
ryanzhang-oss, line 1113: in this case, does the operator matter?

ryanzhang-oss (approved):

pkg/utils/validator/clusterresourceplacement_test.go
```diff
@@ -1146,13 +1153,24 @@ func TestValidateTolerations(t *testing.T) {
...
-				t.Errorf("validateTolerations() actualErr = %s, wantErr %s", actualErr.Error(), testCase.wantErr.Error())
+			gotErr := validateTolerations(testCase.tolerations)
+			if (gotErr != nil) != testCase.wantErr {
+				t.Errorf("validateTolerations() error = %v, wantErr %v", gotErr, testCase.wantErr)
+			}
+			if testCase.wantErr {
+				match, err := regexp.MatchString(testCase.wantErrMsg, gotErr.Error())
+				if err != nil {
+					t.Errorf("validateTolerations() failed to compile pattern: %s", testCase.wantErrMsg)
+				}
```
zhiying-lin, line 1170: nit, you can use strings.contains instead.

zhiying-lin (approved):

PR #689, opened 2024-02-21
feat: work agent track availability

ryanzhang-oss:
Support track resource availability on work agent and reflect it on the work status. 

I have:

### DIFF
--- a/pkg/controllers/work/apply_controller.go
+++ b/pkg/controllers/work/apply_controller.go
@@ -27,11 +27,14 @@ import (
 	"time"
 
 	"go.uber.org/atomic"
+	appv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -42,13 +45,15 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller"
+	ctrloption "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	"go.goms.io/fleet/pkg/metrics"
 	"go.goms.io/fleet/pkg/utils"
+	"go.goms.io/fleet/pkg/utils/condition"
+	"go.goms.io/fleet/pkg/utils/controller"
 	"go.goms.io/fleet/pkg/utils/resource"
 )
 
@@ -58,12 +63,18 @@ const (
 
 // WorkCondition condition reasons
 const (
-	// AppliedWorkFailedReason is the reason string of work condition when it failed to apply the work.
-	AppliedWorkFailedReason = "AppliedWorkFailed"
-	// AppliedWorkCompleteReason is the reason string of work condition when it finished applying work.
-	AppliedWorkCompleteReason = "AppliedWorkComplete"
-	// AppliedManifestFailedReason is the reason string of condition when it failed to apply manifest.
-	AppliedManifestFailedReason = "AppliedManifestFailedReason"
+	workAppliedFailedReason       = "WorkAppliedFailed"
+	workAppliedCompleteReason     = "WorkAppliedComplete"
+	workNotAvailableYetReason     = "WorkNotAvailableYet"
+	workAvailabilityUnknownReason = "WorkAvailabilityUnknown"
+	workAvailableReason           = "WorkAvailable"
+	workNotTrackableReason        = "WorkNotTrackable"
+	// ManifestApplyFailedReason is the reason string of condition when it failed to apply manifest.
+	ManifestApplyFailedReason = "ManifestApplyFailedReason"
+	// ManifestAlreadyUpToDateReason is the reason string of condition when the manifest is already up to date.
+	ManifestAlreadyUpToDateReason = "ManifestAlreadyUpToDate"
+	// ManifestNeedsUpdateReason is the reason string of condition when the manifest needs to be updated.
+	ManifestNeedsUpdateReason = "ManifestNeedsUpdate"
 )
 
 // ApplyWorkReconciler reconciles a Work object
@@ -92,22 +103,32 @@ func NewApplyWorkReconciler(hubClient client.Client, spokeDynamicClient dynamic.
 	}
 }
 
-// applyAction represents the action we take to apply the manifest
+// applyAction represents the action we take to apply the manifest.
+// It is used only internally to track the result of the apply function.
 // +enum
 type applyAction string
 
 const (
-	// ManifestCreatedAction indicates that we created the manifest for the first time.
-	ManifestCreatedAction applyAction = "ManifestCreated"
+	// manifestCreatedAction indicates that we created the manifest for the first time.
+	manifestCreatedAction applyAction = "ManifestCreated"
 
-	// ManifestThreeWayMergePatchAction indicates that we updated the manifest using three-way merge patch.
-	ManifestThreeWayMergePatchAction applyAction = "ManifestThreeWayMergePatched"
+	// manifestThreeWayMergePatchAction indicates that we updated the manifest using three-way merge patch.
+	manifestThreeWayMergePatchAction applyAction = "ManifestThreeWayMergePatched"
 
-	// ManifestServerSideAppliedAction indicates that we updated the manifest using server side apply.
-	ManifestServerSideAppliedAction applyAction = "ManifestServerSideApplied"
+	// manifestServerSideAppliedAction indicates that we updated the manifest using server side apply.
+	manifestServerSideAppliedAction applyAction = "ManifestServerSideApplied"
 
-	// ManifestNoChangeAction indicates that we don't need to change the manifest.
-	ManifestNoChangeAction applyAction = "ManifestNoChange"
+	// errorApplyAction indicates that there was an error during the apply action.
+	errorApplyAction applyAction = "ErrorApply"
+
+	// manifestNotAvailableYetAction indicates that we still need to wait for the manifest to be available.
+	manifestNotAvailableYetAction applyAction = "ManifestNotAvailableYet"
+
+	// manifestNotTrackableAction indicates that the manifest is already up to date but we don't have a way to track its availabilities.
+	manifestNotTrackableAction applyAction = "ManifestNotTrackable"
+
+	// manifestAvailableAction indicates that the manifest is available.
+	manifestAvailableAction applyAction = "ManifestAvailable"
 )
 
 // applyResult contains the result of a manifest being applied.
@@ -115,7 +136,7 @@ type applyResult struct {
 	identifier fleetv1beta1.WorkResourceIdentifier
 	generation int64
 	action     applyAction
-	err        error
+	applyErr   error
 }
 
 // Reconcile implement the control loop logic for Work object.
@@ -140,7 +161,7 @@ func (r *ApplyWorkReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	case err != nil:
 		klog.ErrorS(err, "failed to retrieve the work", "work", req.NamespacedName)
-		return ctrl.Result{}, err
+		return ctrl.Result{}, controller.NewAPIServerError(true, err)
 	}
 	logObjRef := klog.KObj(work)
 
@@ -164,7 +185,7 @@ func (r *ApplyWorkReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	// apply the manifests to the member cluster
-	results := r.applyManifests(ctx, work.Spec.Workload.Manifests, owner)
+	results := r.applyManifests(ctx, work.Spec.Workload.Manifests, owner, work.Spec.ApplyStrategy)
 
 	// collect the latency from the work update time to now.
 	lastUpdateTime, ok := work.GetAnnotations()[utils.LastWorkUpdateTimeAnnotationKey]
@@ -182,7 +203,7 @@ func (r *ApplyWorkReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	// generate the work condition based on the manifest apply result
-	errs := r.generateWorkCondition(results, work)
+	errs := constructWorkCondition(results, work)
 
 	// update the work status
 	if err = r.client.Status().Update(ctx, work, &client.SubResourceUpdateOptions{}); err != nil {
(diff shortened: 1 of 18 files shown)


pkg/controllers/work/apply_controller.go
```diff
@@ -416,21 +442,117 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+	klog.V(2).InfoS("still need to wait for daemonSet to be available", "daemonSet", klog.KObj(curObj))
+	return ManifestNotAvailableYetAction, nil
+}
+
+func trackJobAvailability(curObj *unstructured.Unstructured) (applyAction, error) {
+	var job batchv1.Job
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(curObj.Object, &job); err != nil {
+		return ErrorApplyAction, controller.NewUnexpectedBehaviorError(err)
+	}
+	if job.Status.Ready != nil {
```
michaelawyu, line 540: Hi Ryan! I am not 100% sure about the condition here: for the Job API, the pod might not stay in the ready stage; if it's no longer active (succeeded, failed), should we also consider it as available?

michaelawyu: For completed Jobs, TTL might also add some complications here -> though that might be less of a concern

apis/placement/v1beta1/binding_types.go
```diff
@@ -137,6 +137,13 @@ const (
 	// - "False" means not all the resources are created in the target cluster yet.
 	// - "Unknown" means it is unknown.
 	ResourceBindingApplied ResourceBindingConditionType = "Applied"
+
+	// ResourceBindingAvailable indicates the available condition of the given resources.
+	// Its condition status can be one of the following:
+	// - "True" means all the resources are available in the target cluster.
+	// - "False" means not all the resources are available in the target cluster yet.
+	// - "Unknown" means it is unknown.
+	ResourceBindingAvailable ResourceBindingConditionType = "Applied"
```
michaelawyu, line 146: Hi Ryan! I think this should use a different value.

michaelawyu: And, it's just a nit, but I can't help feeling that the `available` part might have many different interpretations -> though at this moment I couldn't find a more accurate alternative

pkg/controllers/work/apply_controller.go
ryanzhang-oss: I don't think we should consider a job with only failed pod as available but yes for succeeded.

apis/placement/v1beta1/binding_types.go
ryanzhang-oss: let's discuss this when we start to touch the binding. I will scope this PR to work agent only.

pkg/controllers/work/apply_controller_test.go
```diff
@@ -288,6 +284,747 @@ func TestIsManifestManagedByWork(t *testing.T) {
...
+					Status: metav1.ConditionTrue,
+					Reason: workAppliedCompleteReason,
+				},
+				{
+					Type:   fleetv1beta1.WorkConditionTypeAvailable,
+					Status: metav1.ConditionTrue,
+					Reason: workAvailableReason,
+				},
+			},
+		},
```
zhiying-lin, line 770: add test what if one of applied condition is unknown or all of the applied conditions are unknown

pkg/controllers/work/apply_controller_test.go
```diff
@@ -288,6 +284,747 @@ func TestIsManifestManagedByWork(t *testing.T) {
...
+						"availableReplicas":  2,
+						"currentReplicas":    3,
+						"updatedReplicas":    3,
+					},
+				},
+			},
+			expected: manifestNotAvailableYetAction,
+			err:      nil,
+		},
+		"Test DaemonSet Available": {
```
zhiying-lin, line 908: add test daemonset does not observe the latest generation

pkg/controllers/work/apply_controller_test.go
```diff
@@ -288,6 +284,747 @@ func TestIsManifestManagedByWork(t *testing.T) {
...
+		})
+	}
+}
+
+func TestTrackResourceAvailability(t *testing.T) {
+	tests := map[string]struct {
+		gvr      schema.GroupVersionResource
+		obj      *unstructured.Unstructured
+		expected applyAction
+		err      error
```
zhiying-lin, line 786: it's redundant, always nil

pkg/controllers/workgenerator/controller.go
```diff
@@ -659,10 +659,14 @@ func (r *Reconciler) SetupWithManager(mgr controllerruntime.Manager) error {
...
+				oldAvailableStatus := meta.FindStatusCondition(oldWork.Status.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
+				newAvailableStatus := meta.FindStatusCondition(newWork.Status.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
+
+				// we only need to handle the case the applied or available conditions is flipped between true and NOT true between the
 				// new and old work objects. Otherwise, it won't affect the binding applied condition
-				if condition.IsConditionStatusTrue(oldAppliedStatus, oldWork.GetGeneration()) == condition.IsConditionStatusTrue(newAppliedStatus, newWork.GetGeneration()) {
-					klog.V(2).InfoS("The work applied condition didn't flip between true and false, no need to reconcile", "oldWork", klog.KObj(oldWork), "newWork", klog.KObj(newWork))
+				if condition.IsConditionStatusTrue(oldAppliedStatus, oldWork.GetGeneration()) == condition.IsConditionStatusTrue(newAppliedStatus, newWork.GetGeneration()) &&
+					condition.IsConditionStatusTrue(oldAvailableStatus, oldWork.GetGeneration()) == condition.IsConditionStatusTrue(newAvailableStatus, newWork.GetGeneration()) {
```
zhiying-lin, line 668: Should it be if there are no changes between old condition and new condition then we'll skip the reconcile?

eg, oldWork applied condition is false and newWork applied condition is false,

We could use the EqualCondition func to compare https://github.com/Azure/fleet/blob/main/pkg/utils/condition/condition.go#L16 

current = old, desired = new // though old observed generation should never greater than the new one.

pkg/controllers/work/apply_controller.go
```diff
@@ -224,8 +245,13 @@ func (r *ApplyWorkReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			"work", logObjRef)
 	}
 
-	// we periodically reconcile the work to make sure the member cluster state is in sync with the work
-	// even if the reconciling succeeds in case the resources on the member cluster is removed/changed.
+	availableCond := meta.FindStatusCondition(work.Status.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
+	if !condition.IsConditionStatusTrue(availableCond, work.Generation) {
```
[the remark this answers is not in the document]
ryanzhang-oss: we return err here too so we will actually exponentially step back on failed to apply error.

pkg/controllers/work/apply_controller.go
```diff
@@ -415,21 +442,121 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(curObj.Object, &job); err != nil {
+		return errorApplyAction, controller.NewUnexpectedBehaviorError(err)
+	}
+	if job.Status.Succeeded > 0 {
+		klog.V(2).InfoS("job is available with at least one succeeded pod", "job", klog.KObj(curObj))
+		return manifestAvailableAction, nil
+	}
+	if job.Status.Ready != nil {
+		// we consider a job available if there is at least one pod ready
+		if *job.Status.Ready > 0 {
```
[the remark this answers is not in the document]
ryanzhang-oss: didn't we check at line 544?

pkg/controllers/work/apply_controller.go
```diff
@@ -653,50 +787,132 @@ func buildResourceIdentifier(index int, object *unstructured.Unstructured, gvr s
...
+	}
+	// the manifest condition should not be an empty list
+	for _, manifestCond := range manifestConditions {
+		if meta.IsStatusConditionFalse(manifestCond.Conditions, fleetv1beta1.WorkConditionTypeApplied) {
+			// we mark the entire work applied condition to false if one of the manifests is applied failed
+			applyCondition.Status = metav1.ConditionFalse
+			applyCondition.Reason = workAppliedFailedReason
+			applyCondition.Message = fmt.Sprintf("Apply manifest %+v failed", manifestCond.Identifier)
+			availableCondition.Status = metav1.ConditionUnknown
+			availableCondition.Reason = workAppliedFailedReason
```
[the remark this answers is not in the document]
ryanzhang-oss: what message should we use?

pkg/controllers/work/apply_controller.go
```diff
@@ -92,30 +103,40 @@ func NewApplyWorkReconciler(hubClient client.Client, spokeDynamicClient dynamic.
...
+	errorApplyAction applyAction = "ErrorApply"
+
+	// manifestNotAvailableYetAction indicates that we still need to wait for the manifest to be available.
+	manifestNotAvailableYetAction applyAction = "ManifestNotAvailableYet"
+
+	// manifestNotTrackableAction indicates that the manifest is already up to date but we don't have a way to track its availabilities.
+	manifestNotTrackableAction applyAction = "ManifestNotTrackable"
+
+	// manifestAvailableAction indicates that the manifest is available.
+	manifestAvailableAction applyAction = "ManifestAvailable"
```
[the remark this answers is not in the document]
ryanzhang-oss: I thought about it from the beginning, conceptually they are two status but  the problem with two status (essentially applied status and available status) is that these two are not independent. There are only 6 possible combinations just as the current number of actions. With two status, the switch case becomes a bit more evolved.

pkg/controllers/work/apply_controller.go
```diff
@@ -415,21 +442,121 @@ func (r *ApplyWorkReconciler) applyUnstructured(ctx context.Context, gvr schema.
...
-func (r *ApplyWorkReconciler) applyObject(ctx context.Context, gvr schema.GroupVersionResource,
+func trackStatefulSetAvailability(curObj *unstructured.Unstructured) (applyAction, error) {
+	var statefulSet appv1.StatefulSet
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(curObj.Object, &statefulSet); err != nil {
+		return errorApplyAction, controller.NewUnexpectedBehaviorError(err)
+	}
+	// a statefulSet is available if all the replicas are available and the currentReplicas is equal to the updatedReplicas
+	// which means there is no more update in progress.
+	if statefulSet.Status.ObservedGeneration == statefulSet.Generation &&
+		statefulSet.Status.AvailableReplicas == *statefulSet.Spec.Replicas &&
```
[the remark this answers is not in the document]
ryanzhang-oss: hmm..... I am not sure if it can be nil as it's getting from the cluster.

pkg/controllers/work/apply_controller.go
```diff
@@ -653,50 +787,132 @@ func buildResourceIdentifier(index int, object *unstructured.Unstructured, gvr s
...
+		applyCondition.Message = fmt.Sprintf("Failed to apply manifest: %v", err)
+		availableCondition.Status = metav1.ConditionUnknown
+		availableCondition.Reason = ManifestApplyFailedReason
+	} else {
+		applyCondition.Status = metav1.ConditionTrue
+		// the first three actions types means we did write to the cluster thus the availability is unknown
+		// the last three actions types means we start to track the resources
+		switch action {
+		case manifestCreatedAction:
+			applyCondition.Reason = string(manifestCreatedAction)
```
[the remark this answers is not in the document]
ryanzhang-oss: what message should we use?

pkg/controllers/work/apply_controller.go
```diff
@@ -653,50 +787,132 @@ func buildResourceIdentifier(index int, object *unstructured.Unstructured, gvr s
...
+			// we mark the entire work applied condition to false if one of the manifests is applied failed
+			applyCondition.Status = metav1.ConditionFalse
+			applyCondition.Reason = workAppliedFailedReason
+			applyCondition.Message = fmt.Sprintf("Apply manifest %+v failed", manifestCond.Identifier)
+			availableCondition.Status = metav1.ConditionUnknown
+			availableCondition.Reason = workAppliedFailedReason
+			return []metav1.Condition{applyCondition, availableCondition}
+		}
+	}
+	applyCondition.Status = metav1.ConditionTrue
```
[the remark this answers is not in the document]
ryanzhang-oss: I don't think so. We treat timeout as failed.

pkg/controllers/work/apply_controller_helper_test.go
```diff
@@ -86,6 +86,28 @@ func waitForWorkToApply(workName, workNS string) *fleetv1beta1.Work {
...
+	Eventually(func() bool {
+		err := k8sClient.Get(context.Background(), types.NamespacedName{Name: workName, Namespace: workNS}, &resultWork)
+		if err != nil {
+			return false
+		}
+		applyCond := meta.FindStatusCondition(resultWork.Status.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
+		if applyCond == nil || applyCond.Status != metav1.ConditionTrue || applyCond.ObservedGeneration != resultWork.Generation {
+			return false
+		}
+		for _, manifestCondition := range resultWork.Status.ManifestConditions {
```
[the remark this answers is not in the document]
ryanzhang-oss: to make sure this won't happen as this is a test

pkg/controllers/work/apply_controller.go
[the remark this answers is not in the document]
zhiying-lin: can we add a comment there? i missed this before

pkg/controllers/work/apply_controller.go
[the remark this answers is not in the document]
zhiying-lin: some human readable messages, something like 

"Manifest has just been updated and in the processing of checking its availability"?

pkg/controllers/work/apply_controller.go
```diff
@@ -653,50 +790,133 @@ func buildResourceIdentifier(index int, object *unstructured.Unstructured, gvr s
...
+	applyCondition.Reason = workAppliedCompleteReason
+	applyCondition.Message = "Apply work complete"
```
zhiying-lin, line 885:
```suggestion
	applyCondition.Message = "Apply work completed"
```

pkg/controllers/work/apply_controller.go
```diff
@@ -653,50 +790,133 @@ func buildResourceIdentifier(index int, object *unstructured.Unstructured, gvr s
...
+	for _, manifestCond := range manifestConditions {
+		cond := meta.FindStatusCondition(manifestCond.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
+		if cond.Reason == string(manifestNotTrackableAction) {
+			trackable = false
+			break
+		}
+	}
+	availableCondition.Status = metav1.ConditionTrue
+	if trackable {
+		availableCondition.Reason = workAvailableReason
```
zhiying-lin, line 917:
```suggestion
		availableCondition.Reason = workAvailableReason
		availableCondition.Message = "Manifests are trackable and available now"
```

pkg/controllers/work/apply_controller.go
```diff
@@ -653,50 +790,133 @@ func buildResourceIdentifier(index int, object *unstructured.Unstructured, gvr s
...
+	} else {
+		availableCondition.Reason = workNotTrackableReason
```
zhiying-lin, line 919:
```suggestion
		availableCondition.Reason = workNotTrackableReason
		availableCondition.Message = "Manifests are untrackable"
```

pkg/controllers/work/apply_controller.go
```diff
@@ -92,30 +103,40 @@ func NewApplyWorkReconciler(hubClient client.Client, spokeDynamicClient dynamic.
...
 
-	// ManifestServerSideAppliedAction indicates that we updated the manifest using server side apply.
-	ManifestServerSideAppliedAction applyAction = "ManifestServerSideApplied"
+	// manifestServerSideAppliedAction indicates that we updated the manifest using server side apply.
+	manifestServerSideAppliedAction applyAction = "ManifestServerSideApplied"
 
-	// ManifestNoChangeAction indicates that we don't need to change the manifest.
-	ManifestNoChangeAction applyAction = "ManifestNoChange"
+	// errorApplyAction indicates that there was an error during the apply action.
+	errorApplyAction applyAction = "ErrorApply"
```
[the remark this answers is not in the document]
ryanzhang-oss: what is returned is not a pointer though, there needs to be a default value. The caller is ignoring this value now but the tests are checking it.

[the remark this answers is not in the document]
zhiying-lin: That's true, otherwise you need to change the return type as the pointer.
I'm fine with either way.

pkg/controllers/work/apply_controller.go
michaelawyu: Job is a particularly tricky API type to handle. It really depends on the completion pattern I guess; if it's indexed, having one succeeded pod does not say much, and for those with workqueues, we might never see it's completion.

michaelawyu: Taking a peek at the job spec and decides in accordance with the job controller behavior seems like too much work though 😣

zhiying-lin (approved):

PR #691, opened 2024-02-21
perf: increase the number of  scheduler workers

ryanzhang-oss:

Allow the users to setup more than one worker for the scheduler.

Fixes #

I have:

### DIFF
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -218,7 +218,6 @@ func (s *Scheduler) scheduleOnce(ctx context.Context) {
 		observeSchedulingCycleMetrics(cycleStartTime, false, true)
 	}
 	observeSchedulingCycleMetrics(cycleStartTime, false, false)
-	return
 }
 
 // Run starts the scheduler.
--- a/pkg/scheduler/scheduler.go
+++ b/pkg/scheduler/scheduler.go
@@ -88,7 +88,7 @@ func NewScheduler(
 
 // ScheduleOnce performs scheduling for one single item pulled from the work queue.
 // it returns true if the context is not canceled, false otherwise.
-func (s *Scheduler) scheduleOnce(ctx context.Context) bool {
+func (s *Scheduler) scheduleOnce(ctx context.Context) {
 	// Retrieve the next item (name of a CRP) from the work queue.
 	//
 	// Note that this will block if no item is available.
@@ -96,7 +96,7 @@ func (s *Scheduler) scheduleOnce(ctx context.Context) bool {
 	if closed {
 		// End the run immediately if the work queue has been closed.
 		klog.InfoS("Work queue has been closed")
-		return false
+		return
 	}
 
 	defer func() {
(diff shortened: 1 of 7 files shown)


pkg/scheduler/scheduler.go
```diff
@@ -221,14 +227,21 @@ func (s *Scheduler) Run(ctx context.Context) {
...
-	// The loop starts in a dedicated goroutine; it exits when the context is canceled.
-	go wait.UntilWithContext(ctx, s.scheduleOnce, 0)
+	wg := &sync.WaitGroup{}
+	wg.Add(s.workerNumber)
+	for i := 0; i < s.workerNumber; i++ {
+		go func() {
+			defer wg.Done()
+			defer utilruntime.HandleCrash()
+			// Run scheduleOnce forever.
+			s.scheduleOnce(ctx)
```
michaelawyu, line 237: Hi Ryan! I think the test is failing because each worker only run once and then exits

test/e2e/README.md
```diff
@@ -14,7 +14,7 @@ test suites, follow the steps below:
     # Use a different path if the local set up is different.
     export KUBECONFIG=~/.kube/config
     export OUTPUT_TYPE=type=docker
-    ./setup.sh
+    ./setup.sh ${number of member clusers}
```
michaelawyu, line 17: Hi Ryan! A type here: it should be `clusters`

pkg/scheduler/scheduler.go
```diff
@@ -221,14 +229,23 @@ func (s *Scheduler) Run(ctx context.Context) {
...
-	go wait.UntilWithContext(ctx, s.scheduleOnce, 0)
+	wg := &sync.WaitGroup{}
+	wg.Add(s.workerNumber)
+	for i := 0; i < s.workerNumber; i++ {
+		go func() {
+			defer wg.Done()
+			defer utilruntime.HandleCrash()
+			// Run scheduleOnce forever.
+			//revive:disable:empty-block
+			for s.scheduleOnce(ctx) {
```
michaelawyu, line 240: Hi Ryan! Just out of curiosity, is there any particular reason why 

```
for {
    scheduleOnce(ctx)
}
```

does not work here?

michaelawyu: Also, the returned bool does not seem to be used at all; is it for future uses?

michaelawyu (commented):
Added a few comments, PTAL

pkg/scheduler/scheduler.go
ryanzhang-oss: > Hi Ryan! Just out of curiosity, is there any particular reason why
> 
> ```
> for {
>     scheduleOnce(ctx)
> }
> ```
> 
> does not work here?

how will this for loop end?

ryanzhang-oss: > Also, the returned bool does not seem to be used at all; is it for future uses?

the return bool is used so that the for loop will end if the ctx is done.

michaelawyu: Oh, not sure what I was thinking before 😣 Though, in this case, just a nit: would be it more idiomatic to use

```
for {
    select {
    case <-ctx.Done():
        return
    default:
        scheduleOnce(ctx)
    }
}
```

michaelawyu (approved):
LGTM ;)

PR #698, opened 2024-02-28
fix: remove unused informers

britaniar:

Removed unused informers from the informer manager

Fixes #

I have:


pkg/utils/informer/informermanager.go
```diff
@@ -119,12 +121,19 @@ func (s *informerManagerImpl) AddDynamicResources(dynResources []APIResourceMeta
 			s.apiResources[newRes.GroupVersionKind] = &newRes
 			// TODO (rzhang): remember the ResourceEventHandlerRegistration and remove it when the resource is deleted
 			// TODO: handle error which only happens if the informer is stopped
-			_, _ = s.informerFactory.ForResource(newRes.GroupVersionResource).Informer().AddEventHandler(handler)
+			reg, err = s.informerFactory.ForResource(newRes.GroupVersionResource).Informer().AddEventHandler(handler)
+			if err != nil {
+				klog.ErrorS(err, "Failed to add event handler")
+			}
 			klog.InfoS("Added an informer for a new resource", "res", newRes)
 		} else if !dynRes.isPresent {
```
ryanzhang-oss, line 129: we don't need isPresent field now that we can remove the handler

pkg/utils/informer/informermanager.go
```diff
@@ -213,6 +226,40 @@ func (s *informerManagerImpl) IsClusterScopedResources(gvk schema.GroupVersionKi
 	return resMeta.IsClusterScoped
 }
 
+func (s *informerManagerImpl) Remove(gvk schema.GroupVersionKind, registration cache.ResourceEventHandlerRegistration) {
```
ryanzhang-oss, line 229: what is this for?

britaniar: This is where I was confusing myself. Originally, I thought I had to create/implement the remove function but then I looked when I worked/looked at the section above that's when I realized that we probably didn't need this since it would be removed above.

pkg/utils/informer/informermanager_test.go
```diff
@@ -0,0 +1,125 @@
...
+			impl.AddDynamicResources([]APIResourceMeta{dynResources[0]}, handler, true)
+
+			// verify the map. Check that the dynamic resource was removed from the informer manager
+			_, ok := impl.apiResources[dynResources[1].GroupVersionKind]
+			Expect(ok).ShouldNot(BeTrue(), "Expected dynamic resource %v to be removed from informer manager", dynResources[1])
+			// Remove the dynamic resource from the informer manager
+			impl.AddDynamicResources([]APIResourceMeta{}, handler, true)
+
+			// verify the map. Check that the dynamic resource was removed from the informer manager
+			_, ok = impl.apiResources[dynResources[0].GroupVersionKind]
```
ryanzhang-oss, line 121: we need to test more than if the imp has it, we don't need IT to test that.

pkg/utils/informer/informermanager_test.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
+)
+
+var _ = Describe("Informer Manager Suite", func() {
+	Context("add then remove", Ordered, func() {
```
ryanzhang-oss, line 12: this should be a UT

go.mod
```diff
@@ -1,17 +1,17 @@
 module go.goms.io/fleet
 
-go 1.20
+go 1.21
```
michaelawyu, line 3: Hi Britania! Is this Go + deps version bump intentional?

britaniar: I updated the controller-runtime packages I guess I updated the other as well with it.

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -273,9 +273,9 @@ func (r *Reconciler) fetchAllResourcesInOneNamespace(namespaceName string, place
 		if !r.shouldSelectResource(gvr) {
 			continue
 		}
-		if !r.InformerManager.IsInformerSynced(gvr) {
-			return nil, controller.NewExpectedBehaviorError(fmt.Errorf("informer cache for %+v is not synced yet", gvr))
-		}
+		//if !r.InformerManager.IsInformerSynced(gvr) {
+		//	return nil, controller.NewExpectedBehaviorError(fmt.Errorf("informer cache for %+v is not synced yet", gvr))
+		//}
```
britaniar, line 278: Had to change to use GVK to be able to get the informer but GVK does not used in this function

pkg/utils/informer/informermanager.go
```diff
@@ -145,13 +162,37 @@ func (s *informerManagerImpl) AddDynamicResources(dynResources []APIResourceMeta
...
+					// context is not done, so the error is unexpected
+					panic(err)
+				}
+				return
+			}
+			obj, err := s.scheme.New(gvk)
+			if err != nil {
+				klog.V(2).ErrorS(err, "Could not get object of a disappeared resource", "res", dynRes)
+			}
+			err = s.cache.RemoveInformer(s.ctx, obj.(client.Object))
```
britaniar, line 184: Added RemoveInformer() function here.

pkg/utils/informer/informermanager.go
```diff
@@ -101,32 +114,36 @@ type informerManagerImpl struct {
...
 
 func (s *informerManagerImpl) AddDynamicResources(dynResources []APIResourceMeta, handler cache.ResourceEventHandler, listComplete bool) {
 	newGVKs := make(map[schema.GroupVersionKind]bool, len(dynResources))
-
 	addInformerFunc := func(newRes APIResourceMeta) {
-		dynRes, exist := s.apiResources[newRes.GroupVersionKind]
+		_, exist := s.apiResources[newRes.GroupVersionKind]
 		if !exist {
-			newRes.isPresent = true
+			// TODO: how to add GVK to scheme?
```
britaniar, line 129: The old method (using dynamic informer factory) automatically added the dynamic resources to the scheme, but informer cache does not. Currently, running into an issue where the not how to add the dynamic resources to the scheme using only the GVK?

britaniar (commented):
We are trying out to switch from dynamic informer factory to informer cache to be able to remove unused informers. There is currently some errors that I have been running into.

Arvind helped me look into it further and some links that may be helpful:
- there are no tests which test the case of adding dynamic resources in this [file](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/cache/cache_test.go)
- [This commit](https://github.com/kubernetes-sigs/controller-runtime/commit/bc35946fac1b56071def3ebf299996b281c77442) that shows it allows dynamic addition/removal of informers not the ability to add informers for dynamic resources

PR #699, opened 2024-02-29
test: ITs for taints & tolerations plugin

Arvindthiru:

Fixes #

I have:

Integration tests with the plugin enabled only for tests.

Main Workflows covered:
- Special case for PickFixed adding taints shouldn't matter since plugin won't be run
- Add taint to existing cluster, Create CRP without tolerations, cluster is not picked
- Add taint to existing cluster, Create CRP with tolerations, cluster is picked with matching toleration
- Add taint to existing cluster, Create CRP without tolerations, , cluster is not picked, CRP is updated with tolerations, cluster is picked with matching toleration after CRP udpate
- Create CRP without tolerations, Create new cluster with taints and mark it as healthy to trigger scheduling cycle, cluster is not picked
- Create CRP with tolerations, Create new cluster with taints and mark it as healthy to trigger scheduling cycle, cluster is not picked

Specific Workflows tested:

- Add taints on eligible existing clusters to be selected, Create PickFixed CRP which selecets member clusters with taints - clusters get selected as plugin is not run for PickFixed
- Add taints on some eligible existing clusters to be selected, Create PickAll CRP with no tolerations - clusters with taints aren't picked
- Add taints on some eligible existing clusters to be selected, Create PickAll CRP with affinity, tolerations which tolerate some taints - cluster with untolerated taints aren't picked, cluster with tolerated taints are picked
- Add taints on existing elighible clusters to be seleceted, PickN CRP with affinity - no cluster is picked, then CRP is updated with matching tolerations, clusters are picked
- Create PickAll CRP, Add new cluster with taint - new cluster is not picked
- Create PickAll CRP with matching toleration for new cluster's taint, Add new cluster with taint is added - new cluster is picked
- Create PickN CRP, Add new cluster with taint, upscale doesn't pick new cluster
- Create PickN CRP with matching toleration for new cluster's taint, add new cluster with taint, upscale picks new cluster

### DIFF
--- a/test/scheduler/actuals_test.go
+++ b/test/scheduler/actuals_test.go
@@ -462,7 +462,7 @@ func pickNPolicySnapshotStatusUpdatedActual(
 			cmpopts.SortSlices(lessFuncClusterDecision),
 			cmpopts.EquateEmpty(),
 			// for PickN ignore unselected clusters since there are two possible states for policy snapshot status based on whether status update is successful.
-			cmpopts.IgnoreSliceElements(ignoreUnSelectedClusterDecision),
+			cmpopts.IgnoreSliceElements(ignoreUnselectedClusterDecision),
 		); diff != "" {
 			return fmt.Errorf("policy snapshot status cluster decisions (-got, +want): %s", diff)
 		}
--- a/test/scheduler/actuals_test.go
+++ b/test/scheduler/actuals_test.go
@@ -420,6 +420,7 @@ func pickNPolicySnapshotStatusUpdatedActual(
 	picked, notPicked, filtered []string,
 	scoreByCluster map[string]*placementv1beta1.ClusterScore,
 	policySnapshotName string,
+	opts []cmp.Option,
 ) func() error {
 	return func() error {
 		policySnapshot := &placementv1beta1.ClusterSchedulingPolicySnapshot{}
@@ -458,11 +459,7 @@ func pickNPolicySnapshotStatusUpdatedActual(
 		}
 		if diff := cmp.Diff(
 			policySnapshot.Status.ClusterDecisions, wantClusterDecisions,
-			ignoreClusterDecisionReasonField,
-			cmpopts.SortSlices(lessFuncClusterDecision),
-			cmpopts.EquateEmpty(),
-			// for PickN ignore unselected clusters since there are two possible states for policy snapshot status based on whether status update is successful.
-			cmpopts.IgnoreSliceElements(ignoreUnselectedClusterDecision),
+			opts...,
 		); diff != "" {
 			return fmt.Errorf("policy snapshot status cluster decisions (-got, +want): %s", diff)
 		}
--- a/test/scheduler/actuals_test.go
+++ b/test/scheduler/actuals_test.go
@@ -461,6 +461,8 @@ func pickNPolicySnapshotStatusUpdatedActual(
 			ignoreClusterDecisionReasonField,
 			cmpopts.SortSlices(lessFuncClusterDecision),
 			cmpopts.EquateEmpty(),
+			// for PickN ignore unselected clusters since there are two possible states for policy snapshot status based on whether status update is successful.
+			cmpopts.IgnoreSliceElements(ignoreUnSelectedClusterDecision),
 		); diff != "" {
 			return fmt.Errorf("policy snapshot status cluster decisions (-got, +want): %s", diff)
 		}
--- a/test/scheduler/tainttoleration_integration_test.go
+++ b/test/scheduler/tainttoleration_integration_test.go
@@ -2,6 +2,7 @@
 Copyright (c) Microsoft Corporation.
 Licensed under the MIT license.
 */
+
 package tests
 
 import (
--- a/test/scheduler/tainttoleration_integration_test.go
+++ b/test/scheduler/tainttoleration_integration_test.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -17,6 +19,16 @@ import (
 
 	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
 	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+)
+
+var (
+	taintTolerationCmpOpts = []cmp.Option{
+		ignoreClusterDecisionReasonField,
+		cmpopts.SortSlices(lessFuncClusterDecision),
+		cmpopts.EquateEmpty(),
+		// for PickN ignore unselected clusters since there are two possible states for policy snapshot status based on whether status update is successful.
+		cmpopts.IgnoreSliceElements(ignoreUnselectedClusterDecision),
+	}
 )
 
 var _ = Describe("scheduling CRPs on member clusters with taints & tolerations", func() {
@@ -341,7 +353,7 @@ var _ = Describe("scheduling CRPs on member clusters with taints & tolerations",
 		})
 
 		It("should report status correctly", func() {
-			crpStatusUpdatedActual := pickNPolicySnapshotStatusUpdatedActual(2, []string{}, []string{}, wantFilteredClusters, zeroScoreByCluster, policySnapshotName)
+			crpStatusUpdatedActual := pickNPolicySnapshotStatusUpdatedActual(2, []string{}, []string{}, wantFilteredClusters, zeroScoreByCluster, policySnapshotName, taintTolerationCmpOpts)
 			Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to report status correctly")
 			Consistently(crpStatusUpdatedActual, consistentlyDuration, consistentlyInterval).Should(Succeed(), "Failed to report status correctly")
 		})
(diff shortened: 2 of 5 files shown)


Arvindthiru:
There was one failed unit test failure due to conflict and subsequent backoff from the controller https://github.com/Azure/fleet/actions/runs/8178390730/job/22362196109?pr=699 seems intermittent.

We can increase the timeout ?

test/scheduler/tainttoleration_integration_test.go
```diff
@@ -0,0 +1,702 @@
...
+			removeTaintsFromMemberClusters(taintClusters)
+			// Delete the CRP.
+			ensureCRPAndAllRelatedResourcesDeletion(crpName)
+		})
+	})
+
+	// This is a serial test as adding taints, tolerations can affect other tests.
+	Context("pick N clusters with affinity specified, ignore valid clusters with taints, CRP has some matching tolerations after update", Serial, Ordered, func() {
+		crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+		policySnapshotName1 := fmt.Sprintf(policySnapshotNameTemplate, crpName, 1)
```
zhiying-lin, line 196:
```suggestion
		currentPolicySnapshotName := fmt.Sprintf(policySnapshotNameTemplate, crpName, 1)
```

or oldPolicySnapshotName

test/scheduler/tainttoleration_integration_test.go
```diff
@@ -0,0 +1,702 @@
...
+		policySnapshotName2 := fmt.Sprintf(policySnapshotNameTemplate, crpName, 2)
```
zhiying-lin, line 197:
```suggestion
		newPolicySnapshotName := fmt.Sprintf(policySnapshotNameTemplate, crpName, 2)
```

test/scheduler/tainttoleration_integration_test.go
```diff
@@ -0,0 +1,702 @@
...
+
+		It("should add scheduler cleanup finalizer to the CRP", func() {
+			finalizerAddedActual := crpSchedulerFinalizerAddedActual(crpName)
+			Eventually(finalizerAddedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to add scheduler cleanup finalizer to CRP")
+		})
+
+		It("should create N bindings", func() {
+			hasNScheduledOrBoundBindingsActual := hasNScheduledOrBoundBindingsPresentActual(crpName, 0)
+			Eventually(hasNScheduledOrBoundBindingsActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to create N bindings")
+			Consistently(hasNScheduledOrBoundBindingsActual, consistentlyDuration, consistentlyInterval).Should(Succeed(), "Failed to create N bindings")
```
zhiying-lin, line 253: not need to check the eventually, as you've already checked the consistently. duplicate functionality

test/scheduler/tainttoleration_integration_test.go
```diff
@@ -0,0 +1,702 @@
...
+
+	// This is a serial test as adding a new member cluster may interrupt other test cases.
+	Context("pickAll, add a new healthy cluster with taint and matching toleration", Serial, Ordered, func() {
+		crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+		policySnapshotName := fmt.Sprintf(policySnapshotNameTemplate, crpName, 1)
+		// Prepare a new cluster to avoid interrupting other concurrently running test cases.
+		newUnhealthyMemberClusterName := fmt.Sprintf(provisionalClusterNameTemplate, GinkgoParallelProcess())
+		updatedHealthyClusters := healthyClusters
+		updatedHealthyClusters = append(updatedHealthyClusters, newUnhealthyMemberClusterName)
+		// Copy the map to avoid interrupting other concurrently running test cases.
```
zhiying-lin, line 343: The purpose is to make our test env clean? the test is serial and ordered

test/scheduler/tainttoleration_integration_test.go
```diff
@@ -0,0 +1,702 @@
...
+	})
+
+	// This is a serial test as adding a new member cluster may interrupt other test cases.
+	Context("pickAll, add a new healthy cluster with taint and matching toleration", Serial, Ordered, func() {
+		crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+		policySnapshotName := fmt.Sprintf(policySnapshotNameTemplate, crpName, 1)
+		// Prepare a new cluster to avoid interrupting other concurrently running test cases.
+		newUnhealthyMemberClusterName := fmt.Sprintf(provisionalClusterNameTemplate, GinkgoParallelProcess())
+		updatedHealthyClusters := healthyClusters
+		updatedHealthyClusters = append(updatedHealthyClusters, newUnhealthyMemberClusterName)
```
zhiying-lin, line 342: did not get it, the newUnhealthyMemberClusterName is unhealthy and why it will be added into the healthy group?

the name is wrong?

test/scheduler/tainttoleration_integration_test.go
```diff
@@ -0,0 +1,702 @@
...
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+
+	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
+	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+)
+
+var _ = Describe("scheduling CRPs on member clusters with taints & tolerations", func() {
```
zhiying-lin, line 21: add a test, create crp, add a taint to the existing cluster (selected by the crp), to validate if the placement will be removed from the cluster or not

test/scheduler/tainttoleration_integration_test.go
```diff
@@ -0,0 +1,702 @@
...
+		AfterAll(func() {
+			// Delete the CRP.
+			ensureCRPAndAllRelatedResourcesDeletion(crpName)
+			// Delete the provisional cluster.
+			ensureProvisionalClusterDeletion(newUnhealthyMemberClusterName)
+		})
+	})
+
+	// This is a serial test as adding a new member cluster may interrupt other test cases.
+	Context("pickN with required topology spread constraints, add new cluster with taint, upscaling doesn't pick new cluster", Serial, Ordered, func() {
```
zhiying-lin, line 380: can we make the test simple, focusing on the taint feature by removing the topology spread?

The topology spread has been already covered.

test/scheduler/tainttoleration_integration_test.go
```diff
@@ -0,0 +1,702 @@
...
+		crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+		policySnapshotName := fmt.Sprintf(policySnapshotNameTemplate, crpName, 1)
+		// Prepare a new cluster to avoid interrupting other concurrently running test cases.
+		newUnhealthyMemberClusterName := fmt.Sprintf(provisionalClusterNameTemplate, GinkgoParallelProcess())
```
zhiying-lin, line 384: unhealthy member cluster should not been picked.

test/scheduler/tainttoleration_integration_test.go
zhiying-lin: we add a complicated policy in the e2e tests.

test/scheduler/tainttoleration_integration_test.go
zhiying-lin: my bad, we need the eventually here to make sure the binding has been created, and then use the consistently to make sure it won't be updated during the "duration". Eventually is required.

test/scheduler/actuals_test.go
```diff
@@ -461,6 +461,8 @@ func pickNPolicySnapshotStatusUpdatedActual(
 			ignoreClusterDecisionReasonField,
 			cmpopts.SortSlices(lessFuncClusterDecision),
 			cmpopts.EquateEmpty(),
+			// for PickN ignore unselected clusters since there are two possible states for policy snapshot status based on whether status update is successful.
+			cmpopts.IgnoreSliceElements(ignoreUnSelectedClusterDecision),
```
zhiying-lin, line 465: please confirm with Chen if he's ok with this change.

test/scheduler/utils_test.go
```diff
@@ -89,6 +89,9 @@ var (
 	lessFuncClusterDecision = func(decision1, decision2 placementv1beta1.ClusterDecision) bool {
 		return decision1.ClusterName < decision2.ClusterName
 	}
+	ignoreUnSelectedClusterDecision = func(decision placementv1beta1.ClusterDecision) bool {
```
zhiying-lin, line 92:
```suggestion
	ignoreUnselectedClusterDecision = func(decision placementv1beta1.ClusterDecision) bool {
```

test/scheduler/utils_test.go
```diff
@@ -89,6 +89,9 @@ var (
...
+		return decision.Selected
```
zhiying-lin, line 93:
```suggestion
		return !decision.Selected
```
https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreSliceElements
" Elements are ignored if the function reports true."

you're ignoring the selected clusters?

zhiying-lin (dismissed):

michaelawyu:
Hi Arvind! Thanks for reaching out to me about this issue. I would appreciate if we could keep checking for the details about unselected clusters in the existing test cases, primarily for two reasons:

michaelawyu:
a) as you might have already found out, this test suite is set up with the `envtest` package, which means that the K8s API server runs locally, and consequently the chances of having a network error are extremely low; and
b) the suite covers all scheduler components and remains the only viable place where we could verify if the scheduler is producing decisions about unselected clusters as we expected (specifically their scores). Admittedly it's true that if a partial failure is encountered, info. about unselected clusters so far will be swept, but I do feel that this should be handled as a special case in the test code (which should be best addressed in a separate PR) rather than directly ignoring all of them. Fortunately our current track record seems to suggest that such occurrences are extremely rare. With that said, I also agree that not all test cases need to verify unselected cluster conditions.

michaelawyu:
And, regarding your question about the existing test cases, it's worth noting that our scheduler is designed to produce decisions in a deterministic manner. Specifically, in the case of scheduling with topology spread constraints, we can tell for sure 1) how many iterations there will be 2) which clusters to be selected and unselected in each iteration. Note that the status is only updated in the last iteration, and this means we can compose the expected status result accordingly.

zhiying-lin (approved):

PR #702, opened 2024-03-01
feat: property-based scheduling: add trackers for AKS metric provider

michaelawyu:

This PR adds a few trackers in use by the AKS metric provider.

I have:

- [x] Unit tests

Some unit tests have been temporarily removed to control PR size. Further integration and E2E tests will be added in future PRs.


pkg/metricprovider/aks/trackers/trackers.go
```diff
@@ -0,0 +1,543 @@
...
+	// of the cluster, respectively.
+	totalCapacity    corev1.ResourceList
+	totalAllocatable corev1.ResourceList
+	// perCPUCoreHourlyCost and perGBMemoryHourlyCost are the average per CPU core and per GB memory
+	// costs in the cluster, respectively.
+	perCPUCoreHourlyCost  float64
+	perGBMemoryHourlyCost float64
+
+	// Below are a list of maps that tracks information about individual nodes in the cluster.
+	capacityByNode    map[string]corev1.ResourceList
```
ryanzhang-oss, line 49: do we still care about capcity?

pkg/metricprovider/aks/trackers/trackers.go
```diff
@@ -0,0 +1,543 @@
...
+	nodeSetBySKU      map[string]NodeSet
+	skuByNode         map[string]string
+
+	// costLastUpdated is the timestamp when the per-resource-unit costs are last calculated.
+	costLastUpdated time.Time
+	// costErr tracks any error that occurs during cost calculation.
+	costErr error
+
+	// pp is a pricing provider that facilitates cost calculation.
+	pp PricingProvider
```
ryanzhang-oss, line 60: something a bit more readable?

pkg/metricprovider/aks/trackers/trackers.go
```diff
@@ -0,0 +1,543 @@
...
+
+// NodeSet is a set of nodes.
+type NodeSet map[string]bool
+
+// NodeTracker helps track specific stats about nodes in a Kubernetes cluster, e.g., its count.
+type NodeTracker struct {
+	// totalCapacity and totalAllocatable are the total capacity and allocatable capacity
+	// of the cluster, respectively.
+	totalCapacity    corev1.ResourceList
+	totalAllocatable corev1.ResourceList
```
ryanzhang-oss, line 42: do we compute the available on the fly?

pkg/metricprovider/aks/trackers/pricing.go
```diff
@@ -0,0 +1,53 @@
...
+	OnDemandPrice(instanceType string) (float64, bool)
+	LastUpdated() time.Time
+}
+
+var _ PricingProvider = &AKSKarpenterPricingClient{}
+
+// AKSKarpenterPricingClient is a thin wrapper around the AKS Karpenter pricing client, which
+// implements the PricingProvider interface.
+type AKSKarpenterPricingClient struct {
+	kp *pricing.Provider
```
ryanzhang-oss, line 30: a better name?

pkg/metricprovider/aks/trackers/pricing.go
```diff
@@ -0,0 +1,53 @@
...
+func (k *AKSKarpenterPricingClient) OnDemandPrice(instanceType string) (float64, bool) {
+	return k.kp.OnDemandPrice(instanceType)
+}
+
+// LastUpdated returns the last time the pricing information was updated.
+func (k *AKSKarpenterPricingClient) LastUpdated() time.Time {
+	return k.kp.OnDemandLastUpdated()
+}
+
+func NewAKSKarpenterPricingClient(ctx context.Context, region string) *AKSKarpenterPricingClient {
```
ryanzhang-oss, line 43: is that we need to have one KP client per  region?

pkg/metricprovider/aks/trackers/trackers.go
michaelawyu: Hi Ryan! The total capacity is kept for compatibility reasons; it's also used to calculate costs

pkg/metricprovider/aks/trackers/trackers.go
michaelawyu: Done.

pkg/metricprovider/aks/trackers/trackers.go
michaelawyu: Hi Ryan! There's a separate tracker for pods that tallies the requested resources; when the available capacity info is requested, the provider simply subtracts one tally from another.

pkg/metricprovider/aks/trackers/pricing.go
michaelawyu: Done.

pkg/metricprovider/aks/trackers/pricing.go
michaelawyu: Yeah, that's how the Karpenter pricing client is built. Though, considering that this provider is for AKS clusters only, and one AKS cluster is bound to one region, this should be alright I assume?

pkg/propertyprovider/aks/trackers/trackers.go
```diff
@@ -0,0 +1,570 @@
...
+	// Below are a list of maps that tracks information about individual nodes in the cluster.
+	capacityByNode    map[string]corev1.ResourceList
+	allocatableByNode map[string]corev1.ResourceList
+	nodeSetBySKU      map[string]NodeSet
+	skuByNode         map[string]string
+
+	// costLastUpdated is the timestamp when the per-resource-unit costs are last calculated.
+	costLastUpdated time.Time
+	// costErr tracks any error that occurs during cost calculation.
+	costErr error
```
zhiying-lin, line 58: do you mean the perXXX fields?

create a separate struct to group them together?

pkg/propertyprovider/aks/trackers/trackers.go
```diff
@@ -0,0 +1,570 @@
...
+func (nt *NodeTracker) Costs() (perCPUCoreCost, perGBMemoryCost float64, err error) {
+	nt.mu.Lock()
+	defer nt.mu.Unlock()
+	if nt.costLastUpdated.Before(nt.pricingProvider.LastUpdated()) {
+		nt.calculateCosts()
+	}
+	return nt.perCPUCoreHourlyCost, nt.perGBMemoryHourlyCost, nt.costErr
+}
+
+// NodeTracker helps track specific stats about nodes in a Kubernetes cluster, e.g., the sum
```
zhiying-lin, line 449: wrong comment?

pkg/propertyprovider/aks/trackers/trackers.go
```diff
@@ -0,0 +1,570 @@
...
+// of its requested resources.
+type PodTracker struct {
```
zhiying-lin, line 451: can we move it into a separate file, podtracker.go?

pkg/propertyprovider/aks/trackers/trackers.go
```diff
@@ -0,0 +1,570 @@
...
+// tracked.
+func (pt *PodTracker) AddOrUpdate(pod *corev1.Pod) {
+	pt.mu.Lock()
+	defer pt.mu.Unlock()
+
+	requestsAcrossAllContainers := make(corev1.ResourceList)
+	for _, container := range pod.Spec.Containers {
+		for _, rn := range supportedResourceNames {
+			r := requestsAcrossAllContainers[rn]
+			r.Add(container.Resources.Requests[rn])
```
zhiying-lin, line 484: what if request is not set?
what if both request and limit is not set?

pkg/propertyprovider/aks/trackers/trackers.go
```diff
@@ -0,0 +1,570 @@
...
+			requestsAcrossAllContainers[rn] = r
+		}
+	}
+
+	podIdentifier := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
```
zhiying-lin, line 489: do we want to specially handle the "default"?

based on the API definition
// Namespace defines the space within which each name must be unique. An empty namespace is
// equivalent to the "default" namespace, but "default" is the canonical representation.

pkg/propertyprovider/aks/trackers/pricing.go
```diff
@@ -0,0 +1,53 @@
...
+func (k *AKSKarpenterPricingClient) OnDemandPrice(instanceType string) (float64, bool) {
+	return k.karpenterPricingClient.OnDemandPrice(instanceType)
+}
+
+// LastUpdated returns the last time the pricing information was updated.
+func (k *AKSKarpenterPricingClient) LastUpdated() time.Time {
+	return k.karpenterPricingClient.OnDemandLastUpdated()
+}
+
+func NewAKSKarpenterPricingClient(ctx context.Context, region string) *AKSKarpenterPricingClient {
```
zhiying-lin, line 43: a comment?

pkg/propertyprovider/aks/trackers/trackers.go
michaelawyu: Hi Zhiying! You mean the error part? Only cost has that, and it's not one for each cost property, to streamline the experience.

pkg/propertyprovider/aks/trackers/trackers.go
michaelawyu: Per Go's language spec, reading a nil map is the same as reading an empty map, so it will yield a zero resource.Quantity.

pkg/propertyprovider/aks/trackers/trackers.go
michaelawyu: Oh, for this one, the object we read from the K8s server will have the name set actually (e.g., the user specifies "", and in the server it populates `default`). Even if it doesn't the identifier here would still function as `/` is not allowed in K8s object names.

pkg/propertyprovider/aks/trackers/trackers.go
michaelawyu: Done.

pkg/propertyprovider/aks/trackers/pricing.go
michaelawyu: Done.

pkg/propertyprovider/aks/trackers/trackers.go
michaelawyu: Fixed.

pkg/propertyprovider/aks/trackers/trackers.go
zhiying-lin: it's about line 55 comment, "costLastUpdated is the timestamp when the per-resource-unit costs are last calculated."

I assume both costLastUpdate and costErr are all talking about the "per-resource-unit costs".

Will it be better to be grouped into a separate structure, eg

type struct PerResourceCost {
     perCPUCoreHourlyCost  float64
     perGBMemoryHourlyCost float64
     costLastUpdated time.Time
	// costErr tracks any error that occurs during cost calculation.
	costErr error
}

pkg/propertyprovider/aks/trackers/trackers.go
zhiying-lin: so , if the request is not set, we will ignore this pod. If i understand it correctly, the cost may be not accurate because of missing this config.

Could you please add a comment to record this behavior?

pkg/propertyprovider/aks/trackers/trackers.go
michaelawyu: Discussed offline.

pkg/propertyprovider/aks/trackers/trackers.go
michaelawyu: Discussed offline. For reference, the costs are not calc'd using requested capacities; and due to technical limitations at this moment we cannot reliably tell how much resource a pod currently utilizes

pkg/propertyprovider/aks/trackers/pods.go
```diff
@@ -0,0 +1,132 @@
...
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// PodTracker helps track specific stats about pods in a Kubernetes cluster, e.g., the sum
+// of its requested resources.
+type PodTracker struct {
+	totalRequested corev1.ResourceList
```
zhiying-lin, line 19: add a comment to indicate we'll ignore the container/pod if requested not set?

michaelawyu: The cost info part has already included the formulas on that front. As for the pod tracker, I don't we will be ignoring the pod/container per se; the pod/container literally requested zero resource

zhiying-lin: makes sense

zhiying-lin (approved):

PR #711, opened 2024-03-05
feat: populate the crp status using new conditions

zhiying-lin:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -206,6 +206,35 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		klog.V(2).InfoS("Placement has been scheduled", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 		r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementScheduleSuccess", "Successfully scheduled the placement")
 	}
+
+	if r.UseNewConditions {
+		// We skip checking the last resource condition (available) because it will be covered by checking isRolloutCompleted func.
+		for i := RolloutStartedCondition; i < TotalCondition-1; i++ {
+			oldCond := oldCRP.GetCondition(string(i.ClusterResourcePlacementConditionType()))
+			newCond := crp.GetCondition(string(i.ClusterResourcePlacementConditionType()))
+			if !condition.IsConditionStatusTrue(oldCond, oldCRP.Generation) &&
+				condition.IsConditionStatusTrue(newCond, crp.Generation) {
+				klog.V(2).InfoS("Placement resource condition status has been changed to true", "clusterResourcePlacement", crpKObj, "generation", crp.Generation, "condition", i.ClusterResourcePlacementConditionType())
+				r.Recorder.Event(crp, corev1.EventTypeNormal, i.EventReasonForTrue(), i.EventMessageForTrue())
+			}
+		}
+
+		// There is no need to check if the CRP is available or not.
+		// If the available condition is true, it means the rollout is completed.
+		if isRolloutCompleted(r.UseNewConditions, crp) {
+			if !isRolloutCompleted(r.UseNewConditions, oldCRP) {
+				klog.V(2).InfoS("Placement rollout has finished and resources are available", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
+				r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Resources are available in the selected clusters")
+			}
+			// We don't need to requeue any request now by watching the binding changes
+			return ctrl.Result{}, nil
+		}
+		klog.V(2).InfoS("Placement rollout has not finished yet or resources are not available yet", "clusterResourcePlacement", crpKObj, "status", crp.Status, "generation", crp.Generation)
+		// We don't need to requeue any request now by watching the binding changes.
+		// Once the scheduler makes the decision, the binding will be changed by the scheduler at the same time.
+		return ctrl.Result{}, nil
+	}
+
 	if !isCRPSynchronized(oldCRP) && isCRPSynchronized(crp) {
 		klog.V(2).InfoS("Placement has been synchronized", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 		r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementSyncSuccess", "Successfully synchronized the placement")
@@ -213,8 +242,8 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 
 	// There is no need to check if the CRP is applied or not.
 	// If the applied condition is true, it means the scheduling is done & works have been synchronized which means the rollout is completed.
-	if isRolloutCompleted(crp) {
-		if !isRolloutCompleted(oldCRP) {
+	if isRolloutCompleted(r.UseNewConditions, crp) {
+		if !isRolloutCompleted(r.UseNewConditions, oldCRP) {
 			klog.V(2).InfoS("Placement rollout has finished", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 			r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Resources have been applied to the selected clusters")
 		}
@@ -903,23 +932,27 @@ func (r *Reconciler) setPlacementStatus(ctx context.Context, crp *fleetv1beta1.C
 	// could be true when resources are applied successfully on these 3 clusters and the detailed the resourcePlacementStatuses
 	// need to be populated.
 	if scheduledCondition.Status == metav1.ConditionUnknown {
-		conditions := []metav1.Condition{
-			{
-				Status:             metav1.ConditionUnknown,
-				Type:               string(fleetv1beta1.ClusterResourcePlacementSynchronizedConditionType),
-				Reason:             SynchronizePendingReason,
-				Message:            "Scheduling has not completed",
-				ObservedGeneration: crp.Generation,
-			},
-			{
-				Status:             metav1.ConditionUnknown,
-				Type:               string(fleetv1beta1.ClusterResourcePlacementAppliedConditionType),
-				Reason:             ApplyPendingReason,
-				Message:            "Scheduling has not completed",
-				ObservedGeneration: crp.Generation,
-			},
+		// For the new conditions, we no longer populate the remaining otherwise it's too complicated and the default condition
+		// will be unknown.
+		if !r.UseNewConditions {
+			conditions := []metav1.Condition{
+				{
+					Status:             metav1.ConditionUnknown,
+					Type:               string(fleetv1beta1.ClusterResourcePlacementSynchronizedConditionType),
+					Reason:             SynchronizePendingReason,
+					Message:            "Scheduling has not completed",
+					ObservedGeneration: crp.Generation,
+				},
+				{
+					Status:             metav1.ConditionUnknown,
+					Type:               string(fleetv1beta1.ClusterResourcePlacementAppliedConditionType),
+					Reason:             ApplyPendingReason,
+					Message:            "Scheduling has not completed",
+					ObservedGeneration: crp.Generation,
+				},
+			}
+			crp.SetConditions(conditions...)
 		}
-		crp.SetConditions(conditions...)
 		// skip populating detailed resourcePlacementStatus & work related conditions
 		// reset other status fields
 		// TODO: need to track whether we have deleted the resources for the last decisions.
@@ -927,6 +960,10 @@ func (r *Reconciler) setPlacementStatus(ctx context.Context, crp *fleetv1beta1.C
 		// Today, we only track the resources progress if the same cluster is selected again.
 		crp.Status.PlacementStatuses = []fleetv1beta1.ResourcePlacementStatus{}
 		return false, nil
+	}
+
+	if r.UseNewConditions {
+		return r.setResourceConditions(ctx, crp, latestSchedulingPolicySnapshot, latestResourceSnapshot)
 	}
 
 	return r.setResourcePlacementStatusAndResourceConditions(ctx, crp, latestSchedulingPolicySnapshot, latestResourceSnapshot)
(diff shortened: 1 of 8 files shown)


pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -2505,3 +2520,3179 @@ func TestBuildFailedResourcePlacements(t *testing.T) {
 		})
 	}
 }
+
+func TestSetPlacementStatus_useNewConditions(t *testing.T) {
```
ryanzhang-oss, line 2524: this UT's attack surface is actually very large, it's a little bit hard to make sure it covers all the combinations. I normally build some basic UT first before testing the top function.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -342,3 +342,340 @@ func buildFailedResourcePlacements(work *fleetv1beta1.Work) (isPending bool, res
 	}
 	return false, res
 }
+
+// --------------------------------- populating the new conditions------------
+
+// setResourceConditions sets the resource related conditions by looking at the bindings and work, excluding the scheduled
+// condition.
```
ryanzhang-oss, line 349: what does first boolean return value mean?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -342,3 +342,340 @@ func buildFailedResourcePlacements(work *fleetv1beta1.Work) (isPending bool, res
...
+			break
+		} else if clusterConditionStatusRes[i][FalseConditionStatus] > 0 {
+			crp.SetConditions(i.FalseClusterResourcePlacementCondition(crp.Generation, clusterConditionStatusRes[i][FalseConditionStatus]))
+			break
+		} else {
+			crp.SetConditions(i.TrueClusterResourcePlacementCondition(crp.Generation, clusterConditionStatusRes[i][TrueConditionStatus]))
+		}
+	}
+
+	return isClusterScheduled, nil
```
ryanzhang-oss, line 452: isClusterScheduled is always true by now?

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -2505,3 +2520,3179 @@ func TestBuildFailedResourcePlacements(t *testing.T) {
...
+						fleetv1beta1.ResourceGroupHashAnnotation:         "hash",
+						fleetv1beta1.NumberOfResourceSnapshotsAnnotation: "1",
+					},
+				},
+			},
+			want: false,
+			wantStatus: &fleetv1beta1.ClusterResourcePlacementStatus{
+				SelectedResources:     selectedResources,
+				ObservedResourceIndex: "0",
+				Conditions: []metav1.Condition{
```
ryanzhang-oss, line 2693: I can't seem to find where this is set in the setPlacementStatus code at all.

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -2505,3 +2520,3179 @@ func TestBuildFailedResourcePlacements(t *testing.T) {
...
+							BlockOwnerDeletion: ptr.To(true),
+							Controller:         ptr.To(true),
+							APIVersion:         fleetAPIVersion,
+							Kind:               "ClusterResourcePlacement",
+						},
+					},
+					Annotations: map[string]string{
+						fleetv1beta1.ResourceGroupHashAnnotation:         "hash",
+						fleetv1beta1.NumberOfResourceSnapshotsAnnotation: "1",
+					},
```
ryanzhang-oss, line 2686: do those matter in this test?

pkg/controllers/clusterresourceplacement/placement_status_test.go
ryanzhang-oss: given that we didn't provide crp in this test, I don't get why we are testing this wantStatus

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -2505,3 +2520,3179 @@ func TestBuildFailedResourcePlacements(t *testing.T) {
...
+				},
+			},
+			want: false,
+			wantStatus: &fleetv1beta1.ClusterResourcePlacementStatus{
+				SelectedResources:     selectedResources,
+				ObservedResourceIndex: "0",
+				Conditions: []metav1.Condition{
+
```
ryanzhang-oss, line 2616: empty line

ryanzhang-oss (commented):
I didn't finish all the test review

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -206,15 +206,44 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		klog.V(2).InfoS("Placement has been scheduled", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 		r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementScheduleSuccess", "Successfully scheduled the placement")
 	}
+
+	if r.UseNewConditions {
```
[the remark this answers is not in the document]
zhiying-lin: >It seems that this is not set to true in this PR. I assume that it's just for dry run?

 The flag can be removed when we finish all the changes in other controllers. Otherwise, cx will see empty conditions in the crp.

>is this PR safe to release?

Yes, as the code won't run during the runtime and it should not matter whether the cluster has crd installed or not.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -206,15 +206,44 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
+
+		// There is no need to check if the CRP is available or not.
+		// If the available condition is true, it means the rollout is completed.
+		if isRolloutCompleted(r.UseNewConditions, crp) {
+			if !isRolloutCompleted(r.UseNewConditions, oldCRP) {
+				klog.V(2).InfoS("Placement rollout has finished and resources are available", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
+				r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Resources are available in the selected clusters")
+			}
+			// We don't need to requeue any request now by watching the binding changes
+			return ctrl.Result{}, nil
```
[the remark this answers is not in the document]
zhiying-lin: > where is the change to watch binding?

It will be a separate PR, we need to add a binding watcher for the crp.

> I assume that needs to be tested in at least integration test?
Will do

pkg/controllers/clusterresourceplacement/placement_status.go
zhiying-lin: Added a comment.

pkg/controllers/clusterresourceplacement/placement_status.go
zhiying-lin: depends on line 414. left a TODO on line 435 to handle when isClusterScheduled is false.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -342,3 +342,340 @@ func buildFailedResourcePlacements(work *fleetv1beta1.Work) (isPending bool, res
...
+
+			switch i {
+			case RolloutStartedCondition:
+				if bindingCond.Status == metav1.ConditionTrue {
+					status.ApplicableResourceOverrides = binding.Spec.ResourceOverrideSnapshots
+					status.ApplicableClusterResourceOverrides = binding.Spec.ClusterResourceOverrideSnapshots
+				}
+			case AppliedCondition, AvailableCondition:
+				if bindingCond.Status == metav1.ConditionFalse {
+					if err := r.setFailedPlacementsPerCluster(ctx, crp, binding, status); err != nil {
```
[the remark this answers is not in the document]
zhiying-lin: These two conditions are not designed for per resource.

For example, we treat the override as one operation, apply the overrides on the resources one by one. So it's hard to tell which resource causes the failure.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -342,3 +342,340 @@ func buildFailedResourcePlacements(work *fleetv1beta1.Work) (isPending bool, res
...
+		}
+		if isEnveloped {
+			failedManifest.ResourceIdentifier.Envelope = &fleetv1beta1.EnvelopeIdentifier{
+				Name:      envelopObjName,
+				Namespace: envelopObjNamespace,
+				Type:      fleetv1beta1.EnvelopeType(envelopeType),
+			}
+		}
+
+		appliedCond = meta.FindStatusCondition(manifestCondition.Conditions, fleetv1beta1.WorkConditionTypeApplied)
```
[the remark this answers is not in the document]
zhiying-lin: added a comment to explain why we not check.

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: let me see if i can split.

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: it cares about the label to populate the observed resource index field.

The pros is the tests are not fragile if we change the implementation, eg, populate other fields by looking at the resourceSnapshot.
The cons is the tests are large.

I feel large test file is ok.

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: The conditions are populated in https://github.com/Azure/fleet/blob/main/pkg/controllers/clusterresourceplacement/controller.go#L896

pkg/controllers/clusterresourceplacement/placement_status.go
ryanzhang-oss: I don't get it, if isClusterScheduled  is false then it won't reach here, right?

pkg/controllers/clusterresourceplacement/placement_status_test.go
ryanzhang-oss: but it's not in the function we try to test, right? I didn't get why this matters.

pkg/controllers/clusterresourceplacement/placement_status.go
zhiying-lin: oh, yes, misunderstood your comment. when it reaches to line 452, it's always true. I can change to "true" if it's your concern.

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: oh, here it is testing setPlacementStatus function, from the upper level, so that it can catch the regression in the future for the new conditions.

I will change to validate the internal func one by one instead of validating the upper level only, in order to split the tests and make all the combination clear.

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: Removed ownerReference, which won't be used in the future even.

pkg/controllers/clusterresourceplacement/placement_status_test.go
zhiying-lin: To validate this SetPlacementStatus func, i still need to setup all other objects and this test size itself cannot be reduced. Added few other UTs to cover internal funcs. At least, we can spot the failure funcs easily when there is a regression.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -342,3 +342,360 @@ func buildFailedResourcePlacements(work *fleetv1beta1.Work) (isPending bool, res
...
+	}
+
+	if len(failedResourcePlacements) > maxFailedResourcePlacementLimit {
+		failedResourcePlacements = failedResourcePlacements[0:maxFailedResourcePlacementLimit]
+	}
+	status.FailedPlacements = failedResourcePlacements
+	return nil
+}
+
+func buildFailedResourcePlacementsPerCluster(work *fleetv1beta1.Work) []fleetv1beta1.FailedResourcePlacement {
```
ryanzhang-oss, line 623: this function name is very similar "setFailedPlacementsPerCluster" which is doing a very different job. I think this function should be called something like   "extractFailedResourcePlacementsFromWork" to distinguish these two

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -342,3 +342,360 @@ func buildFailedResourcePlacements(work *fleetv1beta1.Work) (isPending bool, res
...
+
+			switch i {
+			case RolloutStartedCondition:
+				if bindingCond.Status == metav1.ConditionTrue {
+					status.ApplicableResourceOverrides = binding.Spec.ResourceOverrideSnapshots
+					status.ApplicableClusterResourceOverrides = binding.Spec.ClusterResourceOverrideSnapshots
+				}
+			case AppliedCondition, AvailableCondition:
+				if bindingCond.Status == metav1.ConditionFalse {
+					if err := r.setFailedPlacementsPerCluster(ctx, crp, binding, status); err != nil {
```
ryanzhang-oss, line 545: IIRC, we will try to create the failedPlacements if the binding condition is false but the real condition is coming from work. At line 630, we will not create any failure if the work conditions are not current. Is this going to lead to any inconsistent status?

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -342,3 +342,360 @@ func buildFailedResourcePlacements(work *fleetv1beta1.Work) (isPending bool, res
...
+						return nil, err
+					}
+				}
+			}
+
+			meta.SetStatusCondition(&status.Conditions, cond)
+			res = append(res, bindingCond.Status)
```
ryanzhang-oss, line 552: I wonder why is the switch case inserted in the middle of the line 529 which defines var cond and line 551 which uses it?

ryanzhang-oss (commented):
There seems to be very few info logs in the status logic, is there enough log for us to debug live site issues?

pkg/controllers/clusterresourceplacement/placement_status.go
zhiying-lin: moved

ryanzhang-oss (approved):

pkg/controllers/clusterresourceplacement/placement_status_test.go
```diff
@@ -2505,3 +2520,4880 @@ func TestBuildFailedResourcePlacements(t *testing.T) {
...
+	resourceSnapshotName := "snapshot-1"
+	cluster := "member-1"
+	bindingName := "binding-1"
+	tests := []struct {
+		name       string
+		binding    *fleetv1beta1.ClusterResourceBinding
+		works      []fleetv1beta1.Work
+		want       []metav1.ConditionStatus
+		wantStatus fleetv1beta1.ResourcePlacementStatus
+	}{
```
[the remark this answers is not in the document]
zhiying-lin: will address it in the next PR if we decide to keep the existing implementation

PR #718, opened 2024-03-12
test: E2E for taint & toleration plugin

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/test/e2e/taint_toleration_test.go
+++ b/test/e2e/taint_toleration_test.go
@@ -54,7 +54,7 @@ var _ = Describe("placing resource using a cluster resource placement with pickF
 	})
 })
 
-var _ = Describe("placing resources using a cluster resource placement with no placement policy specified, taint cluster, update cluster resource placement with tolerations", Serial, Ordered, func() {
+var _ = Describe("placing resources using a cluster resource placement with no placement policy specified, taint clusters, update cluster resource placement with tolerations", Serial, Ordered, func() {
 	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
 	taintClusterNames := []string{memberCluster1EastProdName, memberCluster2EastCanaryName}
 	selectedClusterNames := []string{memberCluster3WestProdName}
--- a/test/e2e/taint_toleration_test.go
+++ b/test/e2e/taint_toleration_test.go
@@ -1,0 +1,212 @@
+package e2e
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+
+	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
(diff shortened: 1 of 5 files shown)


test/e2e/utils_test.go
```diff
@@ -658,3 +662,75 @@ func verifyWorkPropagationAndMarkAsApplied(memberClusterName, crpName string, re
...
 	}
 }
+
+func buildTaints(memberClusterNames []string) []clusterv1beta1.Taint {
+	var labels map[string]string
+	taints := make([]clusterv1beta1.Taint, len(memberClusterNames))
+	for i, name := range memberClusterNames {
+		labels = labelsByClusterName[name]
+		taints[i].Key = labels[regionLabelName]
+		taints[i].Value = labels[envLabelName]
```
michaelawyu, line 672: Hi Arvind! The taint's key is the region label name, but the value is using an env label name. Is this expected?

michaelawyu: Same for the other few occurrences below.

michaelawyu: Oh, I know that the form is correct; it's just normally the k/v pair should be regionLabel <-> regionValue (a label key and its value) rather than regionLabel <-> envLabel (two label keys).

test/e2e/utils_test.go
```diff
@@ -658,3 +662,75 @@ func verifyWorkPropagationAndMarkAsApplied(memberClusterName, crpName string, re
 		}, eventuallyDuration, eventuallyInterval).Should(Succeed())
 	}
 }
+
+func buildTaints(memberClusterNames []string) []clusterv1beta1.Taint {
+	var labels map[string]string
+	taints := make([]clusterv1beta1.Taint, len(memberClusterNames))
+	for i, name := range memberClusterNames {
+		labels = labelsByClusterName[name]
```
michaelawyu, line 670: Also, I am not 100% sure about deriving taints/tolerations from the labelsByClusterName top-level map -> there's no guarantee that all clusters have the required labels.

michaelawyu: This implicit assumption can be troublesome if the global map is updated later.

test/e2e/taint_toleration_test.go
```diff
@@ -0,0 +1,212 @@
...
+			Eventually(resourceRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to check if resources doesn't exist on member cluster")
+		}
+	})
+
+	AfterAll(func() {
+		// Remove taint from member cluster 1, 2.
+		removeTaintsFromMemberClusters(taintClusterNames)
+		ensureCRPAndRelatedResourcesDeletion(crpName, []*framework.Cluster{memberCluster2EastCanary})
+	})
+})
```
michaelawyu, line 212: Also, could you add a test case to verify the NoSchedule behavior? i.e., adding a taint won't trigger the removal of resources from a cluster that has been selected?

michaelawyu: It can be done in separate PR, of course.

michaelawyu (approved):
LGTM ;) Thanks for making the changes 🙏

PR #721, opened 2024-03-14
feat: add cluster resource override validation webhook configuration

britaniar:

Fixes #

I have: added webhook configuration for ClusterResourceOverride

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -46,6 +46,7 @@ import (
 	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	fleetv1alpha1 "go.goms.io/fleet/apis/v1alpha1"
 	"go.goms.io/fleet/cmd/hubagent/options"
+	"go.goms.io/fleet/pkg/webhook/clusterresourceoverride"
 	"go.goms.io/fleet/pkg/webhook/clusterresourceplacement"
 	"go.goms.io/fleet/pkg/webhook/fleetresourcehandler"
 	"go.goms.io/fleet/pkg/webhook/membercluster"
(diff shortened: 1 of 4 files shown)


zhiying-lin:
we don't need to create a separate validation webhook. it can be part of the existing one. You can refer to how we validate member cluster CR. https://github.com/Azure/fleet/blob/main/pkg/webhook/webhook.go#L294

https://github.com/Azure/fleet/blob/main/pkg/webhook/membercluster/membercluster_validating_webhook.go

The core func is in the https://github.com/Azure/fleet/blob/main/pkg/utils/validator/membercluster.go#L19.

Can you create the first PR to implement such util for the override first?

pkg/webhook/webhook.go
```diff
@@ -307,6 +309,22 @@ func (w *Config) buildFleetValidatingWebhooks() []admv1.ValidatingWebhook {
 			},
 			TimeoutSeconds: longWebhookTimeout,
 		},
+		{
+			Name:                    "fleet.clusterresourceoverride.validating",
+			ClientConfig:            w.createClientConfig(clusterresourceoverride.ValidationPath),
+			FailurePolicy:           &ignoreFailurePolicy,
```
zhiying-lin, line 315: the policy should be "fail"

pkg/webhook/webhook.go
```diff
@@ -307,6 +309,22 @@ func (w *Config) buildFleetValidatingWebhooks() []admv1.ValidatingWebhook {
...
+			AdmissionReviewVersions: admissionReviewVersions,
+			Rules: []admv1.RuleWithOperations{
+				{
+					Operations: []admv1.OperationType{
+						admv1.Create,
+						admv1.Update},
+					Rule: createRule([]string{fleetv1alpha1.GroupVersion.Group}, []string{fleetv1alpha1.GroupVersion.Version}, []string{clusterResourceOverrideName}, &clusterScope),
+				},
+			},
+			TimeoutSeconds: shortWebhookTimeout,
```
zhiying-lin, line 326: should use long webhook timeout

zhiying-lin (approved):

PR #723, opened 2024-03-15
feat: rollout controller supports override

zhiying-lin:

The bindings is out of date if the resourcesnapshot and their overrides are not latest.

Fixes #

I have:

UT tests, need to add ITs.

### DIFF
--- a/charts/hub-agent/templates/crds/placement.kubernetes-fleet.io_clusterresourceoverrides.yaml
+++ b/charts/hub-agent/templates/crds/placement.kubernetes-fleet.io_clusterresourceoverrides.yaml
@@ -1,0 +1,1 @@
+../../../../config/crd/bases/placement.kubernetes-fleet.io_clusterresourceoverrides.yaml
--- a/charts/hub-agent/templates/crds/placement.kubernetes-fleet.io_resourceoverrides.yaml
+++ b/charts/hub-agent/templates/crds/placement.kubernetes-fleet.io_resourceoverrides.yaml
@@ -1,0 +1,1 @@
+../../../../config/crd/bases/placement.kubernetes-fleet.io_resourceoverrides.yaml
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"golang.org/x/sync/errgroup"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -29,6 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
 	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	"go.goms.io/fleet/pkg/utils/condition"
 	"go.goms.io/fleet/pkg/utils/controller"
@@ -127,13 +129,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		return runtime.Result{}, controller.NewUnexpectedBehaviorError(err)
 	}
 
+	matchedCRO, matchedRO, err := r.fetchAllMatchingOverridesForResourceSnapshot(ctx, crp.Name, latestResourceSnapshot)
+	if err != nil {
+		klog.ErrorS(err, "Failed to find all matching overrides for the clusterResourcePlacement", "clusterResourcePlacement", crpName)
+		return runtime.Result{}, err
+	}
+
 	// pick the bindings to be updated according to the rollout plan
-	toBeUpdatedBindings, needRoll := pickBindingsToRoll(allBindings, latestResourceSnapshot.Name, &crp)
+	// staleBoundBindings is a list of "Bound" bindings and are not selected in this round because of the rollout strategy.
+	toBeUpdatedBindings, staleBoundBindings, needRoll, err := r.pickBindingsToRoll(ctx, allBindings, latestResourceSnapshot, &crp, matchedCRO, matchedRO)
+	if err != nil {
+		klog.ErrorS(err, "Failed to pick the bindings to roll", "clusterResourcePlacement", crpName)
+		return runtime.Result{}, err
+	}
+
 	if !needRoll {
 		klog.V(2).InfoS("No bindings are out of date, stop rolling", "clusterResourcePlacement", crpName)
 		return runtime.Result{}, nil
 	}
-	klog.V(2).InfoS("Picked the bindings to be updated", "clusterResourcePlacement", crpName, "numberOfBindings", len(toBeUpdatedBindings))
+	klog.V(2).InfoS("Picked the bindings to be updated", "clusterResourcePlacement", crpName, "numberOfBindings", len(toBeUpdatedBindings), "numberOfStaleBindings", len(staleBoundBindings))
+
+	// Update the status first, so that if the rolling out (updateBindings func) fails in the middle, the controller will
+	// recompute the list and the result may be different.
+	// As far as now, these bindings are blocked by the rollout strategy.
+	if err := r.updateStaleBindingsStatus(ctx, staleBoundBindings); err != nil {
+		return runtime.Result{}, err
+	}
+	klog.V(2).InfoS("Successfully updated status of the stale bindings", "clusterResourcePlacement", crpName, "numberOfStaleBindings", len(staleBoundBindings))
 
 	// Update all the bindings in parallel according to the rollout plan.
 	// We need to requeue the request regardless if the binding updates succeed or not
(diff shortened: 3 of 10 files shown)


charts/hub-agent/templates/crds/placement.kubernetes-fleet.io_clusterresourceoverrides.yaml
michaelawyu, line 1: Hi Zhiying! IIRC we added the `../../crdbased` directory to make sure that the old and the new agents install only the CRDs they need (e.g., the old hub agent has no need for cluster resource snapshot, etc.). If I understand correctly this object is also a new agent only API, it might be better to make this new agent only.

charts/hub-agent/templates/crds/placement.kubernetes-fleet.io_resourceoverrides.yaml
michaelawyu, line 1: Same for this object as well I think.

pkg/controllers/rollout/controller.go
```diff
@@ -244,43 +277,63 @@ func pickBindingsToRoll(allBindings []*fleetv1beta1.ClusterResourceBinding, late
...
 
 	// calculate the cutoff time for a binding to be applied before so that it can be considered ready
 	readyTimeCutOff := time.Now().Add(-time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds) * time.Second)
 
 	// classify the bindings into different categories
 	// TODO: calculate the time we need to wait for the first applied but not ready binding to be ready.
+	crpKObj := klog.KObj(crp)
 	for idx := range allBindings {
 		binding := allBindings[idx]
+		desiredBinding := binding.DeepCopy()
```
michaelawyu, line 301: When there's not many bindings to actually update, this seems to be wasting a lot of cycles deep-copying objects and setting their fields, which are not going to be used at all.

pkg/controllers/rollout/controller.go
```diff
@@ -224,10 +246,21 @@ func waitForResourcesToCleanUp(allBindings []*fleetv1beta1.ClusterResourceBindin
 	return false, nil
 }
 
-// pickBindingsToRoll go through all bindings associated with a CRP and returns the bindings that are ready to be updated.
-// There could be cases that no bindings are ready to be updated because of the maxSurge/maxUnavailable constraints even if there are out of sync bindings.
-// Thus, it also returns a bool indicating whether there are out of sync bindings to be rolled to differentiate those two cases.
-func pickBindingsToRoll(allBindings []*fleetv1beta1.ClusterResourceBinding, latestResourceSnapshotName string, crp *fleetv1beta1.ClusterResourcePlacement) ([]*fleetv1beta1.ClusterResourceBinding, bool) {
+// toBeUpdatedBinding is the stale binding which should be rolled out.
```
michaelawyu, line 249: This struct is also used for the bindings that cannot be rolled out I believe?

michaelawyu: Maybe it's just me but the name gets me a little bit confused, esp. considering that the method below `pickBindingsToRoll` returns two array of this struct (one of them includes the actual rollout targets, and the other needs only a status refresh)

pkg/controllers/rollout/controller.go
```diff
@@ -385,14 +441,26 @@ func pickBindingsToRoll(allBindings []*fleetv1beta1.ClusterResourceBinding, late
...
+	klog.V(2).InfoS("Calculated the max number of bindings to add", "clusterResourcePlacement", crpKObj,
 		"maxSurgeNumber", maxSurgeNumber, "maxReadyNumber", maxReadyNumber, "upperBoundReadyBindings",
 		upperBoundReadyNumber, "maxNumberOfBindingsToAdd", maxNumberToAdd)
-	for i := 0; i < maxNumberToAdd && i < len(boundingCandidates); i++ {
-		toBeUpdatedBinding = append(toBeUpdatedBinding, boundingCandidates[i])
+
+	// boundingCandidatesUnselectedIndex stores the last index of the boundingCandidates which are not selected to be updated.
+	// The rolloutStarted condition of these elements from this index should be updated.
+	boundingCandidatesUnselectedIndex := 0
+	for ; boundingCandidatesUnselectedIndex < maxNumberToAdd && boundingCandidatesUnselectedIndex < len(boundingCandidates); boundingCandidatesUnselectedIndex++ {
```
michaelawyu, line 451: It' just a personal opinion, but, would it be easier if we just re-slice the array?

pkg/controllers/rollout/controller.go
```diff
@@ -224,10 +246,21 @@ func waitForResourcesToCleanUp(allBindings []*fleetv1beta1.ClusterResourceBindin
...
 }
 
-// pickBindingsToRoll go through all bindings associated with a CRP and returns the bindings that are ready to be updated.
-// There could be cases that no bindings are ready to be updated because of the maxSurge/maxUnavailable constraints even if there are out of sync bindings.
-// Thus, it also returns a bool indicating whether there are out of sync bindings to be rolled to differentiate those two cases.
-func pickBindingsToRoll(allBindings []*fleetv1beta1.ClusterResourceBinding, latestResourceSnapshotName string, crp *fleetv1beta1.ClusterResourcePlacement) ([]*fleetv1beta1.ClusterResourceBinding, bool) {
+// toBeUpdatedBinding is the stale binding which should be rolled out.
+type toBeUpdatedBinding struct {
+	currentBinding *fleetv1beta1.ClusterResourceBinding
+	desiredBinding *fleetv1beta1.ClusterResourceBinding // only valid for scheduled or bounded binding
```
michaelawyu, line 252: Just a nit: it should be `bound`, right?

pkg/controllers/rollout/suite_test.go
```diff
@@ -12,6 +12,8 @@ import (
 	"path/filepath"
 	"testing"
 
+	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
```
michaelawyu, line 15: Also a nit, to avoid confusion, would it be better if we use the alias `placementv1alpha1`?

pkg/controllers/rollout/controller.go
```diff
@@ -548,3 +613,52 @@ func handleResourceBinding(binding client.Object, q workqueue.RateLimitingInterf
...
+		Reason:             condition.RolloutNotStartedYetReason,
+		Message:            "The resources cannot be updated to the latest because of the rollout strategy",
+	}
+	if rolloutStarted {
+		cond = metav1.Condition{
+			Type:               string(fleetv1beta1.ResourceBindingRolloutStarted),
+			Status:             metav1.ConditionTrue,
+			ObservedGeneration: binding.Generation,
+			Reason:             condition.RolloutStartedReason,
+			Message:            "Updated the resource to the latest one and starting rolling out",
```
michaelawyu, line 654: Just a nit: the `starting rollout out` part might be a little bit confusing as this condition might persist for a long while.

michaelawyu (commented):
Added some comments, PTAL.

charts/hub-agent/templates/crds/placement.kubernetes-fleet.io_clusterresourceoverrides.yaml
zhiying-lin: noted, left the crd there as we eventually will deprecate the v1alpha1 APIs.

pkg/controllers/rollout/controller.go
zhiying-lin: updated,  right now it will only build the updateInfo when the resources have been changed.

pkg/controllers/rollout/controller.go
zhiying-lin: add a todo there, there are few places. Could do the refactor in a separate PR.

pkg/controllers/rollout/controller_test.go
```diff
@@ -8,21 +8,24 @@ package rollout
...
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllertest"
 
+	"go.goms.io/fleet/pkg/utils/condition"
+
+	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
+	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
```
michaelawyu, line 28: Hi Zhiying! Just a nit: would it be better if here we use placementv1alpha1/placementv1beta1 as well?

pkg/controllers/rollout/controller_test.go
```diff
@@ -35,6 +38,30 @@ var (
...
+		cmpopts.IgnoreFields(fleetv1beta1.ClusterResourceBinding{}, "TypeMeta"),
+		cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion"),
+		cmpopts.SortSlices(func(b1, b2 fleetv1beta1.ClusterResourceBinding) bool {
+			return b1.Name < b2.Name
+		}),
+		cmpopts.IgnoreFields(metav1.Condition{}, "Message"),
+		cmpopts.SortSlices(func(c1, c2 metav1.Condition) bool {
+			return c1.Type < c2.Type
+		}),
+		cmp.Comparer(func(t1, t2 metav1.Time) bool {
```
michaelawyu, line 54: Hi Zhiying! Just being curious, would cmpopts.EquateApproxTime work here?

michaelawyu: Also, for confirmation, the case where the function will return true if either time is zero is an intentional one, right?

michaelawyu (approved):
LGTM except for a few nit comments. Feel free to merge it to unblock the progress ;)

pkg/controllers/rollout/controller_test.go
zhiying-lin: yes, the statusCmpOptions will be used in multiple places. Some of them do not need to compare the timestamp though, so i skip checking them by setting as true.

Could split them in a separate PR.

pkg/controllers/rollout/controller_test.go
zhiying-lin: will address in a separate PR.

PR #727, opened 2024-03-18
feat: cluster resource override validator functions

britaniar:

Fixes #

I have:

### DIFF
--- a/pkg/utils/validator/clusterresourceoverride.go
+++ b/pkg/utils/validator/clusterresourceoverride.go
@@ -1,55 +1,48 @@
 package validator
 
 import (
-	"context"
 	"fmt"
 
 	admissionv1 "k8s.io/api/admission/v1"
+
 	"k8s.io/apimachinery/pkg/util/errors"
-	"k8s.io/klog/v2"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
 	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 )
 
 // ValidateClusterResourceOverride validates cluster resource override fields and returns error.
-func ValidateClusterResourceOverride(ctx context.Context, client client.Client, operation admissionv1.Operation, cro fleetv1alpha1.ClusterResourceOverride) error {
+func ValidateClusterResourceOverride(cro fleetv1alpha1.ClusterResourceOverride, croList *fleetv1alpha1.ClusterResourceOverrideList) error {
 	allErr := make([]error, 0)
 
 	// Check if the resource is being selected by resource name
-	if err := validateResourceSelected(cro); err != nil {
+	if err := validateClusterResourceSelectors(cro); err != nil {
 		return err
-	}
-	// List of cluster resource overrides
-	croList, err := listClusterResourceOverrideList(ctx, client)
-	if err != nil {
-		return err
-	}
-	// Check if the override count limit has been reached
-	if !ValidateClusterResourceOverrideLimit(operation, croList) {
-		allErr = append(allErr, fmt.Errorf("at most 100 cluster resource overrides can be created"))
 	}
 
 	// Check if the override count limit for the resources has been reached
-	if err = ValidateClusterResourceOverrideResourceLimit(operation, cro, croList); err != nil {
+	if err := ValidateClusterResourceOverrideResourceLimit(cro, croList); err != nil {
 		allErr = append(allErr, err)
 	}
 	return errors.NewAggregate(allErr)
 }
 
-// validateResourceSelected checks if override is selecting resource by name.
-func validateResourceSelected(cro fleetv1alpha1.ClusterResourceOverride) error {
+// validateClusterResourceSelectors checks if override is selecting resource by name.
+func validateClusterResourceSelectors(cro fleetv1alpha1.ClusterResourceOverride) error {
+	selectorMap := make(map[fleetv1beta1.ClusterResourceSelector]bool)
 	for _, selector := range cro.Spec.ClusterResourceSelectors {
 		// Check if the resource is not being selected by label selector
 		if selector.LabelSelector != nil {
-			if selector.Name != "" {
-				return fmt.Errorf("the labelSelector and name fields are mutually exclusive in selector %+v", selector)
-			}
 			return fmt.Errorf("label selector is not supported for resource selection %+v", selector)
 		} else if selector.Name == "" {
 			return fmt.Errorf("resource name is required for resource selection %+v", selector)
 		}
+
+		// Check if there are any duplicate selectors
+		if selectorMap[selector] {
+			return fmt.Errorf("duplicate selector %+v", selector)
+		}
+		selectorMap[selector] = true
 	}
 	return nil
 }
@@ -65,7 +58,7 @@ func ValidateClusterResourceOverrideLimit(operation admissionv1.Operation, croLi
 
 // ValidateClusterResourceOverrideResourceLimit checks if there is only 1 cluster resource override per resource,
 // assuming the resource will be selected by the name only.
-func ValidateClusterResourceOverrideResourceLimit(operation admissionv1.Operation, cro fleetv1alpha1.ClusterResourceOverride, croList *fleetv1alpha1.ClusterResourceOverrideList) error {
+func ValidateClusterResourceOverrideResourceLimit(cro fleetv1alpha1.ClusterResourceOverride, croList *fleetv1alpha1.ClusterResourceOverrideList) error {
 	overrideMap := make(map[fleetv1beta1.ClusterResourceSelector]string)
 	// Add overrides and its selectors to the map
 	for _, override := range croList.Items {
@@ -78,22 +71,12 @@ func ValidateClusterResourceOverrideResourceLimit(operation admissionv1.Operatio
 	// Check if any of the cro selectors exist in the override map
 	for _, croSelector := range cro.Spec.ClusterResourceSelectors {
 		if overrideMap[croSelector] != "" {
-			// If update operation, ignore the same cluster resource override
-			if cro.GetName() == overrideMap[croSelector] && operation == admissionv1.Update {
+			// Ignore the same cluster resource override
+			if cro.GetName() == overrideMap[croSelector] {
 				continue
 			}
-			return fmt.Errorf("the resource %v has been selected by both %v and %v, which are not supported", croSelector.Name, cro.GetName(), overrideMap[croSelector])
+			return fmt.Errorf("the resource %v has been selected by both %v and %v, which is not supported", croSelector.Name, cro.GetName(), overrideMap[croSelector])
 		}
 	}
 	return nil
-}
-
-// listClusterResourceOverrideList returns a list of cluster resource overrides.
-func listClusterResourceOverrideList(ctx context.Context, client client.Client) (*fleetv1alpha1.ClusterResourceOverrideList, error) {
-	croList := &fleetv1alpha1.ClusterResourceOverrideList{}
-	if err := client.List(ctx, croList); err != nil {
-		klog.ErrorS(err, "Failed to list cluster resource overrides when validating")
-		return nil, fmt.Errorf("failed to list cluster resource overrides, please retry the request: %w", err)
-	}
-	return croList, nil
 }
--- a/pkg/utils/validator/clusterresourceoverride.go
+++ b/pkg/utils/validator/clusterresourceoverride.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	admissionv1 "k8s.io/api/admission/v1"
-
 	"k8s.io/apimachinery/pkg/util/errors"
 
 	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
--- a/pkg/utils/validator/clusterresourceoverride.go
+++ b/pkg/utils/validator/clusterresourceoverride.go
@@ -3,71 +3,90 @@ package validator
 import (
 	"context"
 	"fmt"
-	"reflect"
 
-	apiErrors "k8s.io/apimachinery/pkg/util/errors"
+	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 
 	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
 )
 
 // ValidateClusterResourceOverride validates cluster resource override fields and returns error.
-func ValidateClusterResourceOverride(ctx context.Context, client client.Client, cro fleetv1alpha1.ClusterResourceOverride) error {
+func ValidateClusterResourceOverride(ctx context.Context, client client.Client, operation admissionv1.Operation, cro fleetv1alpha1.ClusterResourceOverride) error {
 	allErr := make([]error, 0)
 
 	// Check if the resource is being selected by resource name
-	if !ValidateResourceSelectedByName(cro) {
-		allErr = append(allErr, fmt.Errorf("resource is not being selected by resource name"))
+	if err := ValidateResourceSelectedByName(cro); err != nil {
+		allErr = append(allErr, err)
+	}
+	// Get the list of cluster resource overrides
+	croList, err := getClusterResourceOverrideList(ctx, client)
+	if err != nil {
+		allErr = append(allErr, err)
 	}
 	// Check if the override count limit has been reached
-	if !ValidateCROLimit(ctx, client) {
+	if !ValidateClusterResourceOverrideLimit(operation, croList) {
 		allErr = append(allErr, fmt.Errorf("at most 100 cluster resource overrides can be created"))
 	}
+
 	// Check if the override count limit has been reached
-	if !ValidateCROResourceLimit(ctx, client, cro) {
+	if !ValidateClusterResourceOverrideResourceLimit(cro, croList) {
 		allErr = append(allErr, fmt.Errorf("only 1 cluster resource override per resource can be created"))
 	}
-	return apiErrors.NewAggregate(allErr)
+	return errors.NewAggregate(allErr)
 }
 
-// ValidateResourceSelectedByName checks if override is selecting resource by name
-func ValidateResourceSelectedByName(cro fleetv1alpha1.ClusterResourceOverride) bool {
+// ValidateResourceSelectedByName checks if override is selecting resource by name.
+func ValidateResourceSelectedByName(cro fleetv1alpha1.ClusterResourceOverride) error {
 	for _, selector := range cro.Spec.ClusterResourceSelectors {
 		// Check if the resource is being selected by resource name
-		if selector.Name == "" || selector.LabelSelector != nil {
+		if selector.LabelSelector != nil {
+			return fmt.Errorf("label selector is not supported for resource selection")
+		} else if selector.Name == "" {
+			return fmt.Errorf("resource name is required for resource selection")
+		}
+	}
+	return nil
+}
+
+// ValidateClusterResourceOverrideLimit checks if there are at most 100 cluster resource overrides.
+func ValidateClusterResourceOverrideLimit(operation admissionv1.Operation, croList *fleetv1alpha1.ClusterResourceOverrideList) bool {
+	// Check if the override count limit has been reached
+	if operation == admissionv1.Create {
+		return len(croList.Items) < 100
+	}
+	return len(croList.Items) <= 100
+}
+
+// ValidateClusterResourceOverrideResourceLimit checks if there is only 1 cluster resource override per resource.
+func ValidateClusterResourceOverrideResourceLimit(cro fleetv1alpha1.ClusterResourceOverride, croList *fleetv1alpha1.ClusterResourceOverrideList) bool {
+	overrideMap := make(map[fleetv1beta1.ClusterResourceSelector]bool)
+	// Add overrides and its selectors to the map
+	for _, override := range croList.Items {
+		selectors := override.Spec.ClusterResourceSelectors
+		for _, selector := range selectors {
+			overrideMap[selector] = true
+		}
+	}
+
+	// Check if any of the cro selectors exist in the override map
+	for _, croSelector := range cro.Spec.ClusterResourceSelectors {
+		if ok := overrideMap[croSelector]; ok {
 			return false
 		}
 	}
 	return true
 }
 
-// ValidateCROLimit checks if there are at most 100 cluster resource overrides
-func ValidateCROLimit(ctx context.Context, client client.Client) bool {
-	// Create a list of cluster resource overrides with the given resource type
+// getClusterResourceOverrideList returns a list of cluster resource overrides.
+func getClusterResourceOverrideList(ctx context.Context, client client.Client) (*fleetv1alpha1.ClusterResourceOverrideList, error) {
 	croList := &fleetv1alpha1.ClusterResourceOverrideList{}
 	if err := client.List(ctx, croList); err != nil {
-		return false
+		klog.Errorf("Failed to list cluster resource overrides: %v", err)
+		return nil, fmt.Errorf("failed to list cluster resource overrides, please retry the request: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,73 @@
...
+	croList := &fleetv1alpha1.ClusterResourceOverrideList{}
+	if err := client.List(ctx, croList); err != nil {
+		return false
+	}
+	// Check if the override count limit has been reached
+	return len(croList.Items) < 100
+}
+
+// ValidateCROResourceLimit checks if there is only 1 cluster resource override per resource
+func ValidateCROResourceLimit(ctx context.Context, client client.Client, cro fleetv1alpha1.ClusterResourceOverride) bool {
```
zhiying-lin, line 56: can you return the error here? so that we can construct the detailed error msg to user to tell them which one has the same config.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,73 @@
...
+	// Create a list of cluster resource overrides with the given resource type
+	croList := &fleetv1alpha1.ClusterResourceOverrideList{}
```
zhiying-lin, line 58: API list call is expensive. Could you please reduce to one API call?

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,73 @@
...
+	if err := client.List(ctx, croList); err != nil {
+		return false
+	}
+	var count int64
+	for _, override := range croList.Items {
```
zhiying-lin, line 63: can you use the hashmap to record the resource config? then lookup time is O(1). The current time complexity is O(m *n) and perf is not good.

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -0,0 +1,209 @@
...
+							Group:   "group",
+							Version: "v1",
+							Kind:    "Kind",
+							Name:    "valid-name",
+						},
+					},
+				},
+			},
+			want:       true,
+			wantErrMsg: "",
```
zhiying-lin, line 58: wantErrMsg is not validated at all?

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -0,0 +1,209 @@
...
+			for i := 0; i < tt.overrideCount; i++ {
+				cro := &fleetv1alpha1.ClusterResourceOverride{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: fmt.Sprintf("override-%d", i),
+					},
+				}
+				if err := fakeClient.Create(ctx, cro); err != nil {
+					t.Fatalf("failed to create override: %v", err)
+				}
+			}
```
zhiying-lin, line 115: refer to this code to construct the objects in the api server,

https://github.com/Azure/fleet/blob/5cae325eb495d25578ce2d007f7086ab74fe7116/pkg/utils/controller/controller_test.go#L423-L431

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -0,0 +1,209 @@
...
+		},
+	}
+
+	// Run the tests
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			// Create a fake client for testing
+			scheme := runtime.NewScheme()
+			if err := fleetv1alpha1.AddToScheme(scheme); err != nil {
+				t.Fatalf("failed to add scheme: %v", err)
```
zhiying-lin, line 100:
```suggestion
				t.Fatalf("Failed to add scheme: %v", err)
```

log msg starts with Uppercase

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+	"context"
+	"fmt"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+
```
zhiying-lin, line 13: newline is not needed here

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+	// Check if the override count limit has been reached
+	if !ValidateClusterResourceOverrideResourceLimit(cro, croList) {
+		allErr = append(allErr, fmt.Errorf("only 1 cluster resource override per resource can be created"))
+	}
+	return errors.NewAggregate(allErr)
+}
+
+// ValidateResourceSelectedByName checks if override is selecting resource by name.
+func ValidateResourceSelectedByName(cro fleetv1alpha1.ClusterResourceOverride) error {
```
zhiying-lin, line 43:
```suggestion
func validateResourceSelected(cro fleetv1alpha1.ClusterResourceOverride) error {
```
internal func should be good enough as it's only used in one place.

Can we rename to "validateResourceSelected"? later we may support label selector, and we don't need to rename this func at that time.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+	allErr := make([]error, 0)
+
+	// Check if the resource is being selected by resource name
+	if err := ValidateResourceSelectedByName(cro); err != nil {
+		allErr = append(allErr, err)
+	}
+	// Get the list of cluster resource overrides
+	croList, err := getClusterResourceOverrideList(ctx, client)
+	if err != nil {
+		allErr = append(allErr, err)
```
zhiying-lin, line 28: if we cannot get the list, we cannot validate the following so we can return directly.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+
+// ValidateResourceSelectedByName checks if override is selecting resource by name.
+func ValidateResourceSelectedByName(cro fleetv1alpha1.ClusterResourceOverride) error {
+	for _, selector := range cro.Spec.ClusterResourceSelectors {
+		// Check if the resource is being selected by resource name
+		if selector.LabelSelector != nil {
+			return fmt.Errorf("label selector is not supported for resource selection")
+		} else if selector.Name == "" {
+			return fmt.Errorf("resource name is required for resource selection")
+		}
```
zhiying-lin, line 50:
```suggestion
		if selector.LabelSelector != nil {
			return fmt.Errorf("label selector is not supported for resource selection")
		} else if selector.Name == "" {
			return fmt.Errorf("resource name is required for resource selection")
		}
```

should we validate all the selectors similar to https://github.com/Azure/fleet/blob/main/pkg/utils/validator/clusterresourceplacement.go#L44-L49

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// ValidateResourceSelectedByName checks if override is selecting resource by name.
+func ValidateResourceSelectedByName(cro fleetv1alpha1.ClusterResourceOverride) error {
+	for _, selector := range cro.Spec.ClusterResourceSelectors {
+		// Check if the resource is being selected by resource name
+		if selector.LabelSelector != nil {
+			return fmt.Errorf("label selector is not supported for resource selection")
+		} else if selector.Name == "" {
+			return fmt.Errorf("resource name is required for resource selection")
```
zhiying-lin, line 49:
```suggestion
			return fmt.Errorf("resource name is required for resource selection %+v", selector)
```

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return errors.NewAggregate(allErr)
+}
+
+// ValidateResourceSelectedByName checks if override is selecting resource by name.
+func ValidateResourceSelectedByName(cro fleetv1alpha1.ClusterResourceOverride) error {
+	for _, selector := range cro.Spec.ClusterResourceSelectors {
+		// Check if the resource is being selected by resource name
+		if selector.LabelSelector != nil {
+			return fmt.Errorf("label selector is not supported for resource selection")
```
zhiying-lin, line 47:
```suggestion
			return fmt.Errorf("label selector is not supported for resource selection %+v", selector)
```

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+
+	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
+)
+
+// ValidateClusterResourceOverride validates cluster resource override fields and returns error.
+func ValidateClusterResourceOverride(ctx context.Context, client client.Client, operation admissionv1.Operation, cro fleetv1alpha1.ClusterResourceOverride) error {
+	allErr := make([]error, 0)
+
+	// Check if the resource is being selected by resource name
+	if err := ValidateResourceSelectedByName(cro); err != nil {
```
zhiying-lin, line 22: I'm thinking of skipping the following check when it's not nil. especially the ValidateClusterResourceOverrideResourceLimit func assumes the resource is selected by the name, not label.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+		selectors := override.Spec.ClusterResourceSelectors
+		for _, selector := range selectors {
+			overrideMap[selector] = true
+		}
+	}
+
+	// Check if any of the cro selectors exist in the override map
+	for _, croSelector := range cro.Spec.ClusterResourceSelectors {
+		if ok := overrideMap[croSelector]; ok {
+			return false
```
zhiying-lin, line 78: returning false is not good enough, can we return the error instead? as we need to point out which overrides are selecting the resource at the same time to help cx to troubleshoot.

sample error message, the resource %v has been selected by both %v and %v, which are not supported.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+	return true
+}
+
+// getClusterResourceOverrideList returns a list of cluster resource overrides.
+func getClusterResourceOverrideList(ctx context.Context, client client.Client) (*fleetv1alpha1.ClusterResourceOverrideList, error) {
```
zhiying-lin, line 85:
```suggestion
func listClusterResourceOverride(ctx context.Context, client client.Client) (*fleetv1alpha1.ClusterResourceOverrideList, error) {
```

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+	croList := &fleetv1alpha1.ClusterResourceOverrideList{}
+	if err := client.List(ctx, croList); err != nil {
+		klog.Errorf("Failed to list cluster resource overrides: %v", err)
```
zhiying-lin, line 88:
```suggestion
		klog.ErrorS(err, "Failed to list cluster resource overrides when validating")
```

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"fmt"
+	"testing"
+
+	admissionv1 "k8s.io/api/admission/v1"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
```
zhiying-lin, line 14: [group the imports](https://go.dev/wiki/CodeReviewComments#imports)

Imports are organized in groups, with blank lines between them. The standard library packages are always in the first group.

the convension is
standard,
third party
fleet internal

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -0,0 +1,187 @@
...
+					Spec: fleetv1alpha1.ClusterResourceOverrideSpec{
+						ClusterResourceSelectors: []fleetv1beta1.ClusterResourceSelector{
+							{
+								Name: fmt.Sprintf("example-%d", i),
+							},
+						},
+					},
+				}
+				croList.Items = append(croList.Items, cro)
+			}
```
zhiying-lin, line 181: can move this list before the line 164. we don't need to build the inputs for each tests as you only read these data and won't write them.

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -0,0 +1,187 @@
...
+		want bool
+	}{
+		// TODO: Add test cases.
+		"create one cluster resource override for resource foo": {
+			cro: fleetv1alpha1.ClusterResourceOverride{
+				Spec: fleetv1alpha1.ClusterResourceOverrideSpec{
+					ClusterResourceSelectors: []fleetv1beta1.ClusterResourceSelector{
+						{
+							Name: "foo",
+						},
```
zhiying-lin, line 127: can you use a valid selector (populating the required field)? The equality is based on the selector struct, not just name.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,92 @@
...
+func ValidateClusterResourceOverrideLimit(operation admissionv1.Operation, croList *fleetv1alpha1.ClusterResourceOverrideList) bool {
+	// Check if the override count limit has been reached
+	if operation == admissionv1.Create {
+		return len(croList.Items) < 100
+	}
+	return len(croList.Items) <= 100
+}
+
+// ValidateClusterResourceOverrideResourceLimit checks if there is only 1 cluster resource override per resource.
+func ValidateClusterResourceOverrideResourceLimit(cro fleetv1alpha1.ClusterResourceOverride, croList *fleetv1alpha1.ClusterResourceOverrideList) bool {
```
zhiying-lin, line 65: add a comment here to indicate this func assumes the resource will be selected by the name only.

If we allow the label, the validation is not correct.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// validateResourceSelected checks if override is selecting resource by name.
+func validateResourceSelected(cro fleetv1alpha1.ClusterResourceOverride) error {
+	for _, selector := range cro.Spec.ClusterResourceSelectors {
+		// Check if the resource is not being selected by label selector
+		if selector.LabelSelector != nil {
+			if selector.Name != "" {
+				return fmt.Errorf("the labelSelector and name fields are mutually exclusive in selector %+v", selector)
+			}
```
zhiying-lin, line 48:
```suggestion
			if selector.Name != "" {
				return fmt.Errorf("the labelSelector and name fields are mutually exclusive in selector %+v", selector)
			}
```
you don't need to do this check. as we don't support the label at all.

If customer fixes this error by removing the name, they will get unsupported error which is annoying. :)

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,99 @@
...
+		for _, selector := range selectors {
+			overrideMap[selector] = override.GetName()
+		}
+	}
+
+	// Check if any of the cro selectors exist in the override map
+	for _, croSelector := range cro.Spec.ClusterResourceSelectors {
+		if overrideMap[croSelector] != "" {
+			// If update operation, ignore the same cluster resource override
+			if cro.GetName() == overrideMap[croSelector] && operation == admissionv1.Update {
```
zhiying-lin, line 82: oh, now i get your question.
You're right, if it's the update operation, we should not return error.

It can be changed to
```suggestion
			if cro.GetName() == overrideMap[croSelector] {
```
no matter what's the operation.

For the create, it's not possible that the cro.GetName() == overrideMap[croSelector]

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,99 @@
...
+	fleetv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+)
+
+// ValidateClusterResourceOverride validates cluster resource override fields and returns error.
+func ValidateClusterResourceOverride(ctx context.Context, client client.Client, operation admissionv1.Operation, cro fleetv1alpha1.ClusterResourceOverride) error {
+	allErr := make([]error, 0)
+
+	// Check if the resource is being selected by resource name
+	if err := validateResourceSelected(cro); err != nil {
```
zhiying-lin, line 21: validate if this override has duplicate resource selector.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	// Check if the override count limit has been reached
+	if !ValidateClusterResourceOverrideLimit(operation, croList) {
+		allErr = append(allErr, fmt.Errorf("at most 100 cluster resource overrides can be created"))
+	}
+
+	// Check if the override count limit for the resources has been reached
+	if err = ValidateClusterResourceOverrideResourceLimit(operation, cro, croList); err != nil {
+		allErr = append(allErr, err)
+	}
```
zhiying-lin, line 37: as discussed, let's treat these checks as a special validation for CREATE/UPDATE and ResourceLimit is expensive, so skip for other operations too.

Please refer to CRP validation webhook.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Check if any of the cro selectors exist in the override map
+	for _, croSelector := range cro.Spec.ClusterResourceSelectors {
+		if overrideMap[croSelector] != "" {
+			// If update operation, ignore the same cluster resource override
+			if cro.GetName() == overrideMap[croSelector] && operation == admissionv1.Update {
+				continue
+			}
+			return fmt.Errorf("the resource %v has been selected by both %v and %v, which are not supported", croSelector.Name, cro.GetName(), overrideMap[croSelector])
```
zhiying-lin, line 85:
```suggestion
			return fmt.Errorf("the resource %v has been selected by both %v and %v, which is not supported", croSelector.Name, cro.GetName(), overrideMap[croSelector])
```

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,99 @@
...
+	if err := validateResourceSelected(cro); err != nil {
+		return err
+	}
+	// List of cluster resource overrides
+	croList, err := listClusterResourceOverrideList(ctx, client)
+	if err != nil {
+		return err
+	}
+	// Check if the override count limit has been reached
+	if !ValidateClusterResourceOverrideLimit(operation, croList) {
```
zhiying-lin, line 30: we could treat these checks as a special validation for CREATE/UPDATE to improve the readability.

You could refer to CRP validation webhook.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Check if the override count limit for the resources has been reached
+	if err = ValidateClusterResourceOverrideResourceLimit(operation, cro, croList); err != nil {
+		allErr = append(allErr, err)
+	}
+	return errors.NewAggregate(allErr)
+}
+
+// validateResourceSelected checks if override is selecting resource by name.
+func validateResourceSelected(cro fleetv1alpha1.ClusterResourceOverride) error {
```
zhiying-lin, line 42:
```suggestion
func validateClusterResourceSelectors(cro fleetv1alpha1.ClusterResourceOverride) error {
```

pkg/webhook/clusterresourceoverride/clusterresourceoverride_validating_webhook.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// List of cluster resource overrides
+	croList, err := listClusterResourceOverrideList(ctx, v.client)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	// Check if the override count limit has been reached, if there are at most 100 cluster resource overrides
+	if req.Operation == admissionv1.Create || req.Operation == admissionv1.Update {
+		if !validator.ValidateClusterResourceOverrideLimit(req.Operation, croList) {
```
zhiying-lin, line 54: can be simplified as :)

```suggestion
	     if  operation == admissionv1.Create && len(croList.Items) >= 100 {
		return admission.Denied("clusterResourceOverride limit has been reached. At most 100 cluster resources can be created.")
	      }
```

please add a error log

britaniar: Is the function `validator.ValidateClusterResourceOverrideLimit` still needed if we are doing the check in the Handler rather than in ValidateClusterOverrideLimit?

britaniar: Also do we do not need a check for the update operation, because it can be equal to 100 if it updates?

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -0,0 +1,81 @@
...
+	// Check if the override count limit has been reached
+	if operation == admissionv1.Create {
+		return len(croList.Items) < 100
+	}
+	return len(croList.Items) <= 100
+}
+
+// ValidateClusterResourceOverrideResourceLimit checks if there is only 1 cluster resource override per resource,
+// assuming the resource will be selected by the name only.
+func ValidateClusterResourceOverrideResourceLimit(cro fleetv1alpha1.ClusterResourceOverride, croList *fleetv1alpha1.ClusterResourceOverrideList) error {
```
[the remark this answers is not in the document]
britaniar: if it's nil does that mean it will be an error or just that no ClusterResourceOverrides exist?

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -0,0 +1,435 @@
...
+			croList:    nil,
+			wantErrMsg: nil,
+		},
+	}
+	for testName, tt := range tests {
+		t.Run(testName, func(t *testing.T) {
+			got := ValidateClusterResourceOverride(tt.cro, tt.croList)
+			if got != nil {
+				if !strings.Contains(got.Error(), tt.wantErrMsg.Error()) {
+					t.Errorf("validateClusterResourceSelectors() = %v, want %v", got, tt.wantErrMsg)
```
zhiying-lin, line 428:
```suggestion
					t.Errorf("ValidateClusterResourceOverride() = %v, want %v", got, tt.wantErrMsg)
```

pkg/webhook/clusterresourceoverride/clusterresourceoverride_validating_webhook.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if err := validator.ValidateClusterResourceOverride(cro, croList); err != nil {
+		klog.V(2).ErrorS(err, "clusterResourceOverride has invalid fields, request is denied", "operation", req.Operation)
+		return admission.Denied(err.Error())
+	}
+	return admission.Allowed("clusterResourceOverride has valid fields")
+}
+
+// listClusterResourceOverrideList returns a list of cluster resource overrides.
+func listClusterResourceOverrideList(ctx context.Context, client client.Client) (*fleetv1alpha1.ClusterResourceOverrideList, error) {
```
zhiying-lin, line 72:
```suggestion
func listClusterResourceOverrides(ctx context.Context, client client.Client) (*fleetv1alpha1.ClusterResourceOverrideList, error) {
```

zhiying-lin (approved):
You can merge the PR after fixing these two minor comments

pkg/webhook/clusterresourceoverride/clusterresourceoverride_validating_webhook.go
```diff
@@ -43,22 +43,22 @@ func Add(mgr manager.Manager) error {
 // Handle clusterResourceOverrideValidator checks to see if cluster resource override is valid
 func (v *clusterResourceOverrideValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
 	var cro fleetv1alpha1.ClusterResourceOverride
-	klog.V(2).InfoS("Validating webhook handling cluster resource override", "operation", req.Operation)
+	klog.V(2).InfoS("validating webhook handling cluster resource override", "operation", req.Operation)
```
zhiying-lin, line 46: log string should begin with upper case

zhiying-lin (approved):

PR #733, opened 2024-03-21
feat: support apply strategy

zhiying-lin:
- Refactor code to support server side apply
- Add the conflict work check for both types. If the existing resource is owned by other work, the apply strategy should be the same.

Fixes #

I have:

Add new UT
existing tests should make sure no regression.

upstream does not have a fake client to support apply method very well.
Will add IT/e2e tests in a separate PR.

### DIFF
--- a/pkg/controllers/work/applier.go
+++ b/pkg/controllers/work/applier.go
@@ -95,10 +95,12 @@ func validateOwnerReference(ctx context.Context, hubClient client.Client, namesp
 		return applyConflictBetweenPlacements, controller.NewUserError(err)
 	}
 
-	// check if the existing manifest is managed by the work
+	// skip if the current resource is not derived from the work and co-ownership is disallowed
+	// Note: if the co-ownership is added afterwards, e.g., by a controller using on the user-side, all resource changes
+	// will fail to be updated.
 	if !strategy.AllowCoOwnership && !isManifestManagedByWork(ownerRefs) {
-		err := fmt.Errorf("resource exists and is not managed by the fleet controller")
-		klog.ErrorS(err, "Co-ownership is not allowed", "ownerRefs", ownerRefs)
+		err := fmt.Errorf("resource exists and is not managed by the fleet controller and co-ownernship is disallowed")
+		klog.ErrorS(err, "Skip applying a manifest managed by non-fleet applier", "ownerRefs", ownerRefs)
 		return manifestAlreadyOwnedByOthers, controller.NewUserError(err)
 	}
 	return "", nil
--- a/pkg/controllers/work/applier.go
+++ b/pkg/controllers/work/applier.go
@@ -1,0 +1,83 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package work
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/utils/controller"
+	"go.goms.io/fleet/pkg/utils/defaulter"
+)
+
+var (
+	conflictBetweenPlacementsErrorFormat = "manifest is already managed by placement %s but with different apply strategy or the placement strategy is changed"
+)
+
+// Applier is the interface to apply the resources on the member clusters.
+type Applier interface {
+	ApplyUnstructured(ctx context.Context, applyStrategy *fleetv1beta1.ApplyStrategy, gvr schema.GroupVersionResource, manifestObj *unstructured.Unstructured) (*unstructured.Unstructured, ApplyAction, error)
+}
+
+// serverSideApply uses server side apply to apply the manifest.
+func serverSideApply(ctx context.Context, client dynamic.Interface, force bool, gvr schema.GroupVersionResource,
+	manifestObj *unstructured.Unstructured) (*unstructured.Unstructured, ApplyAction, error) {
+	manifestRef := klog.KObj(manifestObj)
+	options := metav1.ApplyOptions{
+		FieldManager: workFieldManagerName,
+		Force:        force,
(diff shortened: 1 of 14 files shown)


pkg/controllers/work/apply_controller.go
```diff
@@ -729,13 +652,18 @@ func computeManifestHash(obj *unstructured.Unstructured) (string, error) {
...
+	if len(ownerRefs) == 0 {
+		return false
+	}
+
+	// an object is NOT managed by the work if any of its owner reference is not of type appliedWork
+	// We'll fail the operation if the resource is owned by other applier.
 	for _, ownerRef := range ownerRefs {
-		if ownerRef.APIVersion == fleetv1beta1.GroupVersion.String() && ownerRef.Kind == fleetv1beta1.AppliedWorkKind {
-			return true
+		if ownerRef.APIVersion != fleetv1beta1.GroupVersion.String() || ownerRef.Kind != fleetv1beta1.AppliedWorkKind {
```
michaelawyu, line 662: Would we like to explicitly ban co-ownership?

pkg/utils/defaulter/work_test.go
```diff
@@ -0,0 +1,90 @@
...
+						ServerSideApplyConfig: &fleetv1beta1.ServerSideApplyConfig{ForceConflicts: true},
+					},
+				},
+			},
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			SetDefaultsWork(&tc.work)
+			if diff := cmp.Diff(tc.work, tc.work); diff != "" {
```
michaelawyu, line 85: It should be diff between `tc.work` and `tc.want` I assume?

michaelawyu (commented):
Added some comments, PTAL.

pkg/controllers/work/applier_fail_if_exists.go
```diff
@@ -0,0 +1,140 @@
...
+		return createFunc()
+	case err != nil:
+		return nil, errorApplyAction, controller.NewAPIServerError(false, err)
+	}
+
+	conflictedWork, err := findConflictedWork(ctx, applier.HubClient, applier.WorkNamespace, applyStrategy, curObj.GetOwnerReferences())
+	if err != nil {
+		return nil, errorApplyAction, err
+	}
+	if conflictedWork != nil {
```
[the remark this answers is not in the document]
zhiying-lin: yes, we could, but i would like to log more error information, eg, gvr, manifestRef.

If it's a common func, too many arguments should be passed in, so I use the current way.

pkg/controllers/work/applier_fail_if_exists.go
```diff
@@ -0,0 +1,140 @@
+/*
```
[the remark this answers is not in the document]
zhiying-lin: @ryanzhang-oss Chen raised a valid point. The current API is more like enum. technically, it could have 2*2 combinations. We only support 2 types of them. 

For example, if we support allow co-own & client-side apply, the troubleshooting will be nightmare and the result is hard to predict.

In addition, i cannot think of a case why user does not want co-own if using server-side apply, as they can resolve the conflicts by using field manager.

WDYT? Will the current API confuse users? Not sure if we should call out the allow co-own explicitly in the API or not.

pkg/controllers/work/apply_controller.go
```diff
@@ -729,13 +655,18 @@ func computeManifestHash(obj *unstructured.Unstructured) (string, error) {
 
 // isManifestManagedByWork determines if an object is managed by the work controller.
 func isManifestManagedByWork(ownerRefs []metav1.OwnerReference) bool {
```
zhiying-lin, line 657: @ryanzhang-oss made some changes to the existing behavior. Please let me know if it's good.

pkg/controllers/work/applier.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if conflictedWork != nil {
+		placement := conflictedWork.Labels[fleetv1beta1.CRPTrackingLabel]
+		err := fmt.Errorf("manifest is already managed by placement %s but with different apply strategy or the placement strategy is changed", placement)
+		klog.ErrorS(err, "Skip applying a manifest managed by another placement but with different apply strategy",
+			"conflictedWork", conflictedWork.Name, "conflictedPlacement", placement, "conflictedWorkApplyStrategy", conflictedWork.Spec.ApplyStrategy)
+		return applyConflictBetweenPlacements, controller.NewUserError(err)
+	}
+
+	// check if the existing manifest is managed by the work
```
michaelawyu, line 98: Hi Zhiying! Just a nit: would you mind updating the comments to be a bit more specific (e.g., -> `skip if the current resource is not derived from the work and co-ownership is disallowed`)?

pkg/controllers/work/applier.go
```diff
@@ -0,0 +1,105 @@
...
+	if !strategy.AllowCoOwnership && !isManifestManagedByWork(ownerRefs) {
+		err := fmt.Errorf("resource exists and is not managed by the fleet controller")
```
michaelawyu, line 100: Same for this error message as well 🙏

pkg/controllers/work/applier.go
michaelawyu: Also, do you think it would be necessary if we add a comment here noting that with the current `isManifestManagedByWork` behavior change, when the co-ownership is added afterwards, e.g., by a controller using on the user-side, all resource changes will cease to sync?

pkg/controllers/work/applier_server_side_test.go
```diff
@@ -0,0 +1,257 @@
...
+	"go.goms.io/fleet/pkg/utils/controller"
+)
+
+func TestApplyUnstructured(t *testing.T) {
+	tests := []struct {
+		name             string
+		allowCoOwnership bool
+		manifest         *unstructured.Unstructured
+		owners           []metav1.OwnerReference
+		doeExist         bool // return whether the deployment exists
```
michaelawyu, line 33: A nit: it should be `doesExist`?

pkg/controllers/work/applier_test.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+}
+
+func TestValidateOwnerReference(t *testing.T) {
+	tests := []struct {
+		name          string
+		applyStrategy placementv1beta1.ApplyStrategy
+		ownerRefs     []metav1.OwnerReference
+		works         []placementv1beta1.Work
+		want          ApplyAction
```
michaelawyu, line 230: Just a nit: maybe `wantAction` is better, considering that there's already a `wantErr`?

michaelawyu (commented):
Just a few nit comments, the rest LGTM ;)

pkg/controllers/work/applier_test.go
zhiying-lin: I usually name the output (excluding error) when there are multiple ones. I feel it's ok for the single output.

michaelawyu (approved):
LGTM ;)

PR #737, opened 2024-03-26
interface: update apply strategy to call out allowing co-ownership explicitly

zhiying-lin:

Fixes #

I have:

Since the current default behavior is client side apply + fail if co-own, add a new field called "allowCoOwnership" so that the default value is false.

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -432,13 +432,17 @@ type RolloutStrategy struct {
 // reason ApplyConflictBetweenPlacements.
 type ApplyStrategy struct {
 	// Type defines the type of strategy to use. Default to ClientSideApply.
+	// Server-side apply is a safer choice. Read more about the differences between server-side apply and client-side
+	// apply: https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply.
 	// +kubebuilder:default=ClientSideApply
 	// +kubebuilder:validation:Enum=ClientSideApply;ServerSideApply
 	// +optional
 	Type ApplyStrategyType `json:"type,omitempty"`
 
-	// AllowCoOwnership defines whether it allows the resource in the target cluster co-owned by other non-fleet applier.
-	// If not, the resource apply will fail if its owner reference includes types other than `appliedWork`.
+	// AllowCoOwnership defines whether to apply the resource if it already exists in the target cluster and is not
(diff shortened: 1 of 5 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -427,16 +427,20 @@ type RolloutStrategy struct {
...
-	// +kubebuilder:default=FailIfExists
-	// +kubebuilder:validation:Enum=FailIfExists;ServerSideApply
+	// Type defines the type of strategy to use. Default to ClientSideApply.
+	// +kubebuilder:default=ClientSideApply
+	// +kubebuilder:validation:Enum=ClientSideApply;ServerSideApply
 	// +optional
 	Type ApplyStrategyType `json:"type,omitempty"`
 
+	// AllowCoOwnership defines whether it allows the resource in the target cluster co-owned by other non-fleet applier.
+	// If not, the resource apply will fail if its owner reference includes types other than `appliedWork`.
```
circy9, line 441: `appliedWork` is not part of the public API. Don't mention internal concepts in public API comments.

The comment describes more about the implementation rather than the user intended behavior.

What about this?

AllowCoOwnership defines whether to apply the resource if it already exists in the target cluster and is not solely owned by fleet (i.e., `metadata.ownerReferences` contains only fleet custom resources).

If true, apply the resource and add fleet as a co-owner.

If false, leave the resource unchanged and fail the apply.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -427,16 +427,20 @@ type RolloutStrategy struct {
...
 // ApplyStrategy describes how to resolve the conflict if the resource to be placed already exists in the target cluster
-// and is owned by other appliers.
+// and whether it's allowed to be co-owned by other non-fleet appliers.
 // Note: If multiple CRPs try to place the same resource with different apply strategy, the later ones will fail with the
 // reason ApplyConflictBetweenPlacements.
 type ApplyStrategy struct {
-	// Type defines the type of strategy to use. Default to FailIfExists.
-	// +kubebuilder:default=FailIfExists
-	// +kubebuilder:validation:Enum=FailIfExists;ServerSideApply
+	// Type defines the type of strategy to use. Default to ClientSideApply.
```
circy9, line 434: Add:

Server-side apply is a safer choice. Read more about the differences between server-side apply and client-side apply: https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply.

circy9 (approved):

PR #738, opened 2024-03-26
fix: use the override snapshot when looking for matched overrides

zhiying-lin:

- use override snapshot instead of override
- rename the api package

Fixes #

I have:

### DIFF
--- a/charts/hub-agent/crdbases/placement.kubernetes-fleet.io_clusterresourceoverridesnapshots.yaml
+++ b/charts/hub-agent/crdbases/placement.kubernetes-fleet.io_clusterresourceoverridesnapshots.yaml
@@ -1,1 +1,0 @@
-../../../config/crd/bases/placement.kubernetes-fleet.io_clusterresourceoverridesnapshots.yaml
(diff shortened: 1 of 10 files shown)


charts/hub-agent/crdbases/placement.kubernetes-fleet.io_clusterresourceoverridesnapshots.yaml
```diff
@@ -0,0 +1 @@
+../../../config/crd/bases/placement.kubernetes-fleet.io_clusterresourceoverridesnapshots.yaml
```
michaelawyu, line 1: Hi Zhiying! Considering that this is for the new agents only and v1alpha1 support is going to be away soon, this soft link is not necessary I believe. It can be directly added in the `crds` directory.

charts/hub-agent/crdbases/placement.kubernetes-fleet.io_resourceoverridesnapshots.yaml
```diff
@@ -0,0 +1 @@
+../../../config/crd/bases/placement.kubernetes-fleet.io_resourceoverridesnapshots.yaml
```
michaelawyu, line 1: Same for this one as well.

michaelawyu (dismissed):
LGTM except for the two comments.

michaelawyu (approved):
LGTM ;)

PR #741, opened 2024-03-27
feat: mark binding as available according to the work

ryanzhang-oss:

mark binding as available according to the work
Fixes #

I have:

### DIFF
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -48,11 +48,13 @@ import (
 )
 
 const (
-	allWorkSyncedReason  = "AllWorkSynced"
-	syncWorkFailedReason = "SyncWorkFailed"
-	workNeedSyncedReason = "StillNeedToSyncWork"
-	workNotAppliedReason = "NotAllWorkHasBeenApplied"
-	allWorkAppliedReason = "AllWorkHasBeenApplied"
+	allWorkSyncedReason    = "AllWorkSynced"
+	syncWorkFailedReason   = "SyncWorkFailed"
+	workNeedSyncedReason   = "StillNeedToSyncWork"
+	workNotAppliedReason   = "NotAllWorkHaveBeenApplied"
+	allWorkAppliedReason   = "AllWorkHaveBeenApplied"
+	workNotAvailableReason = "NotAllWorkAreAvailable"
+	allWorkAvailableReason = "AllWorkAreAvailable"
 )
 
 var (
@@ -138,7 +140,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 			})
 		} else {
 			// try to gather the resource binding applied status if we didn't update any associated work spec this time
-			resourceBinding.SetConditions(buildAllWorkAppliedCondition(works, &resourceBinding))
+			resourceBinding.SetConditions(buildAllWorkAppliedCondition(works, &resourceBinding), buildAllWorkAvailableCondition(works, &resourceBinding))
 		}
 	}
 
@@ -175,7 +177,6 @@ func (r *Reconciler) handleDelete(ctx context.Context, resourceBinding *fleetv1b
 	// background/foreground cascade deletion. This may render the finalizer unnecessary.
 	for workName := range works {
 		work := works[workName]
-
 		if err := r.Client.Delete(ctx, work); err != nil && !apierrors.IsNotFound(err) {
 			return controllerruntime.Result{}, controller.NewAPIServerError(false, err)
 		}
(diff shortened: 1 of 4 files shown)


zhiying-lin:
hi ryan, could you please include creating the workCreated condition? thanks

pkg/controllers/workgenerator/controller.go
```diff
@@ -623,7 +655,7 @@ func (r *Reconciler) SetupWithManager(mgr controllerruntime.Manager) error {
 
 				// we only need to handle the case the applied or available condition is changed between the
 				// new and old work objects. Otherwise, it won't affect the binding applied condition
-				if condition.EqualCondition(oldAppliedStatus, newAppliedStatus) && condition.EqualCondition(oldAvailableStatus, newAvailableStatus) {
+				if condition.EqualCondition(newAppliedStatus, oldAppliedStatus) && condition.EqualCondition(newAvailableStatus, oldAvailableStatus) {
```
zhiying-lin, line 658: we cannot change the order,

if newCondition.ObservedGeneration > oldCondition.ObservedGeneration, we need to requeue the request as the spec has been changed and we need to update other status fields.

For example, status keeps false but the failedManifest set are different.

pkg/controllers/workgenerator/controller.go
```diff
@@ -530,7 +533,36 @@ func buildAllWorkAppliedCondition(works map[string]*fleetv1beta1.Work, binding *
 		Status:             metav1.ConditionFalse,
 		Type:               string(fleetv1beta1.ResourceBindingApplied),
 		Reason:             workNotAppliedReason,
-		Message:            "not all corresponding work objects are applied",
+		Message:            fmt.Sprintf("work object %s is not applied", notAppliedWork),
+		ObservedGeneration: binding.GetGeneration(),
+	}
+}
+
+func buildAllWorkAvailableCondition(works map[string]*fleetv1beta1.Work, binding *fleetv1beta1.ClusterResourceBinding) metav1.Condition {
```
zhiying-lin, line 541: when crp populates the status, if the last condition is unknown or false, it won't populate the remaining in order to keep the our doc simple (easy to explain to user). Otherwise, we have too many combinations. 

Just wonder if we would like to keep them consistent?
so that we only populate the available condition when the applied condition is true.

pkg/controllers/workgenerator/controller.go
```diff
@@ -511,14 +512,16 @@ func getWorkNamePrefixFromSnapshotName(resourceSnapshot *fleetv1beta1.ClusterRes
...
+			notAppliedWork = work.Name
 			break
 		}
 	}
 	if allApplied {
-		klog.V(2).InfoS("All works associated with the binding is applied", "binding", klog.KObj(binding))
+		klog.V(2).InfoS("All works associated with the binding are applied", "binding", klog.KObj(binding))
 		return metav1.Condition{
 			Status:             metav1.ConditionTrue,
 			Type:               string(fleetv1beta1.ResourceBindingApplied),
```
zhiying-lin, line 527: could you please add a message here. CRP will need this.

pkg/controllers/workgenerator/controller.go
```diff
@@ -530,7 +533,36 @@ func buildAllWorkAppliedCondition(works map[string]*fleetv1beta1.Work, binding *
 		Status:             metav1.ConditionFalse,
 		Type:               string(fleetv1beta1.ResourceBindingApplied),
 		Reason:             workNotAppliedReason,
-		Message:            "not all corresponding work objects are applied",
+		Message:            fmt.Sprintf("work object %s is not applied", notAppliedWork),
```
zhiying-lin, line 536:
```suggestion
		Message:            fmt.Sprintf("Work object %s is not applied", notAppliedWork),
```
message begins with upper case?

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: ha, the original order is wrong but now that we care about failedManifest, I am not sure if we can do this comparison anymore.

ryanzhang-oss: we will keep this order and use a back off loop to requeue in the case that not all manifests are available.

pkg/controllers/workgenerator/controller.go
```diff
@@ -511,26 +517,58 @@ func getWorkNamePrefixFromSnapshotName(resourceSnapshot *fleetv1beta1.ClusterRes
...
+			Reason:             allWorkAvailableReason,
+			Message:            "All corresponding work objects are available",
+			ObservedGeneration: binding.GetGeneration(),
+		}
+	}
+	return metav1.Condition{
+		Status:             metav1.ConditionFalse,
+		Type:               string(fleetv1beta1.ResourceBindingAvailable),
+		Reason:             workNotAvailableReason,
+		Message:            fmt.Sprintf("work object %s is not available", notAvailableWork),
```
zhiying-lin, line 571:
```suggestion
		Message:            fmt.Sprintf("Work object %s is not available", notAvailableWork),
```

PR #747, opened 2024-04-02
feat: add validation for cluster selector for override rule

britaniar:

Fixes #

I have:

### DIFF
--- a/pkg/utils/validator/clusterresourceoverride.go
+++ b/pkg/utils/validator/clusterresourceoverride.go
@@ -31,8 +31,10 @@ func ValidateClusterResourceOverride(cro fleetv1alpha1.ClusterResourceOverride, 
 	}
 
 	// Check if override rule is using label selector
-	if err := validateClusterResourceOverrideRuleSelector(cro); err != nil {
-		allErr = append(allErr, err)
+	if cro.Spec.Policy != nil {
+		if err := validateClusterResourceOverridePolicy(cro); err != nil {
+			allErr = append(allErr, err)
+		}
 	}
 
 	return errors.NewAggregate(allErr)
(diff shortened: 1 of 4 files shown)


pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -79,3 +90,20 @@ func validateClusterResourceOverrideResourceLimit(cro fleetv1alpha1.ClusterResou
...
+func validateClusterResourceOverrideRuleSelector(cro fleetv1alpha1.ClusterResourceOverride) error {
+	allErr := make([]error, 0)
+	for _, rule := range cro.Spec.Policy.OverrideRules {
+		if rule.ClusterSelector == nil {
+			continue
+		}
+		for _, selector := range rule.ClusterSelector.ClusterSelectorTerms {
+			// Check that only label selector is supported
+			if selector.LabelSelector == nil || selector.PropertySelector != nil || selector.PropertySorter != nil {
+				allErr = append(allErr, fmt.Errorf("label selector is only supported for resource selection %+v", selector))
```
zhiying-lin, line 104:
```suggestion
			if selector.PropertySelector != nil || selector.PropertySorter != nil {
				allErr = append(allErr, fmt.Errorf("invalid clusterSelector %v: only labelSelector is supported", selector))
```

then we check if the selector.LabelSelector is nil or not.

zhiying-lin: we need to call validateClusterSelector func to validate the selector?

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -373,11 +403,41 @@ func TestValidateClusterResourceOverride(t *testing.T) {
 							Name:    "test-cluster-role",
 						},
 					},
+					Policy: validPolicy,
 				},
 			},
 			croList:    nil,
 			wantErrMsg: nil,
 		},
+		"invalid cluster resource override - fail validateClusterResourceOverrideRuleSelector": {
```
zhiying-lin, line 412: add more tests? eg, policy is nil, nil labelSelector, empty terms? empty matchLabels map and empty matchExpressions?

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -79,3 +90,20 @@ func validateClusterResourceOverrideResourceLimit(cro fleetv1alpha1.ClusterResou
...
+			}
```
zhiying-lin, line 105: we also need to validate JSONPatchOverride to validate value https://github.com/Azure/fleet/blob/main/apis/placement/v1alpha1/override_types.go#L152

btw, the api will be changed in #748 and you can validate it in a separate PR once #748 is merged.

pkg/utils/validator/clusterresourceoverride.go
```diff
@@ -79,3 +92,28 @@ func validateClusterResourceOverrideResourceLimit(cro fleetv1alpha1.ClusterResou
 	}
 	return errors.NewAggregate(allErr)
 }
+
+// validateClusterResourceOverridePolicy checks if override rule is selecting resource by name.
+func validateClusterResourceOverridePolicy(cro fleetv1alpha1.ClusterResourceOverride) error {
+	allErr := make([]error, 0)
+	for _, rule := range cro.Spec.Policy.OverrideRules {
```
ryanzhang-oss, line 99: isn't this the same as the resourceOverride? We should try to avoid code duplication

ryanzhang-oss (approved):

pkg/utils/validator/clusterresourceoverride_test.go
```diff
@@ -1,12 +1,13 @@
...
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"testing"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/errors"
+	apiErrors "k8s.io/apimachinery/pkg/util/errors"
```
zhiying-lin, line 10: always use the lower case for the package name. please fix it in your next PR

pkg/utils/validator/clusterresourceoverride.go
britaniar: Addressed in [#752 ](https://github.com/Azure/fleet/pull/752)

PR #748, opened 2024-04-02
feat: apply overrides before creating the work

zhiying-lin:

- added util to apply override rules before creating the work
- updated the api to make it consistent with JSON PATCH spec so that user can easily migrate from other known tools.
- used apiextensionsv1.JSON instead of string type to support any CRD, one of benefit is it allows no type field, ref link https://github.com/kubernetes-sigs/cluster-api/blob/e8b15a15d70cf94a70df4958764de9c19fb86cf2/api/v1beta1/cluster_types.go#L323

Fixes #

I have:

UT


pkg/controllers/workgenerator/override.go
```diff
@@ -0,0 +1,209 @@
...
+
+	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
+	placementv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
+	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+	"go.goms.io/fleet/pkg/utils"
+	"go.goms.io/fleet/pkg/utils/controller"
+	"go.goms.io/fleet/pkg/utils/overrider"
+)
+
+func (r *Reconciler) fetchClusterResourceOverrideSnapshots(ctx context.Context, resourceBinding *placementv1beta1.ClusterResourceBinding) (map[placementv1beta1.ResourceIdentifier][]*placementv1alpha1.ClusterResourceOverrideSnapshot, error) {
```
ryanzhang-oss, line 27: this is not used yet, right?

pkg/controllers/workgenerator/override.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func (r *Reconciler) applyOverrides(resource *placementv1beta1.ResourceContent, cluster clusterv1beta1.MemberCluster, croMap map[placementv1beta1.ResourceIdentifier][]*placementv1alpha1.ClusterResourceOverrideSnapshot, roMap map[placementv1beta1.ResourceIdentifier][]*placementv1alpha1.ResourceOverrideSnapshot) error {
+	var uResource unstructured.Unstructured
+	if err := uResource.UnmarshalJSON(resource.Raw); err != nil {
+		klog.ErrorS(err, "Work has invalid content", "selectedResource", resource.Raw)
+		return controller.NewUnexpectedBehaviorError(err)
+	}
+	gvk := uResource.GetObjectKind().GroupVersionKind()
+	key := placementv1beta1.ResourceIdentifier{
```
ryanzhang-oss, line 102: this is not used if it's clusterScoped?

pkg/controllers/workgenerator/override.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+			if !matched {
+				continue
+			}
+
+			if err := applyJSONPatchOverride(resource, rule.JSONPatchOverrides); err != nil {
+				klog.ErrorS(err, "Failed to apply JSON patch override", "clusterResourceOverrideSnapshot", klog.KObj(snapshot))
+				return controller.NewUserError(err)
+			}
+		}
```
ryanzhang-oss, line 142: any chance we can reuse this part in the resourceOverride below?

pkg/controllers/workgenerator/override.go
zhiying-lin: yeah, will be called in my next pr

pkg/controllers/workgenerator/override.go
```diff
@@ -0,0 +1,209 @@
...
+				Name:    selector.Name,
+			}
+			croMap[key] = append(croMap[key], snapshot)
+		}
+	}
+	klog.V(2).InfoS("Fetched clusterResourceOverrideSnapshots", "resourceBinding", klog.KObj(resourceBinding), "numberOfResources", len(croMap))
+	return croMap, nil
+}
+
+func (r *Reconciler) fetchResourceOverrideSnapshots(ctx context.Context, resourceBinding *placementv1beta1.ClusterResourceBinding) (map[placementv1beta1.ResourceIdentifier][]*placementv1alpha1.ResourceOverrideSnapshot, error) {
```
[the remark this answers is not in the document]
zhiying-lin: the structure type and log will be different and if combine them, need if condition in func or we need to introduce some interface. Could do the refactor later.

pkg/controllers/workgenerator/override.go
zhiying-lin: it's used in line 122 when applying the clusterResourceOverrideSnapshots.

pkg/utils/overrider/overrider_test.go
```diff
@@ -0,0 +1,194 @@
...
+		},
+		{
+			name: "rule with empty cluster selector",
+			cluster: clusterv1beta1.MemberCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "cluster-1",
+				},
+			},
+			rule: placementv1alpha1.OverrideRule{
+				ClusterSelector: &placementv1beta1.ClusterSelector{},
```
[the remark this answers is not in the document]
zhiying-lin: Now get your point after reading the next comment

// A label selector is a label query over a set of resources. The result of matchLabels and
// matchExpressions are ANDed. An empty label selector matches all objects. A null
// label selector matches no objects.
// +structType=atomic
type LabelSelector struct {

labelSelector is doing the reverse way. i can keep them consistent, otherwise, it may too confusing.

pkg/utils/overrider/overrider_test.go
```diff
@@ -0,0 +1,194 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+			want: true,
+		},
+		{
+			name: "matched overrides with multiple cluster terms",
```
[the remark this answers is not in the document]
zhiying-lin: The current test is one term matches and the other does not.

apis/placement/v1alpha1/override_types.go
```diff
@@ -67,7 +68,9 @@ type OverridePolicy struct {
 type OverrideRule struct {
 	// ClusterSelectors selects the target clusters.
 	// The resources will be overridden before applying to the matching clusters.
-	// If ClusterSelector is not set, it means selecting ALL the member clusters.
+	// An empty clusterSelector selects ALL the member clusters.
+	// A null clusterSelector selects NO member clusters.
+	// For now, only labelSelector is supported.
 	// +optional
 	ClusterSelector *placementv1beta1.ClusterSelector `json:"clusterSelector,omitempty"`
```
zhiying-lin, line 75: @ryanzhang-oss i did not mark this field as required. cx may want to set the null temporarily to ignore this rule or even override for testing purpose.

apis/placement/v1alpha1/override_types.go
```diff
@@ -68,7 +68,8 @@ type OverridePolicy struct {
 type OverrideRule struct {
 	// ClusterSelectors selects the target clusters.
 	// The resources will be overridden before applying to the matching clusters.
-	// If ClusterSelector is not set, it means selecting ALL the member clusters.
+	// An empty clusterSelector selects ALL the member clusters.
+	// A null clusterSelector selects NO member clusters.
```
ryanzhang-oss, line 72: nit, why "null" instead of "nil"?

ryanzhang-oss (approved):

michaelawyu (approved):
Re-approve to unblock the progress.

PR #749, opened 2024-04-03
fix: fix flaky e2e tests

britaniar:

Fixes #

I have: 
- change the name so of the cluster scoped resource so it doesn’t conflict with another 
- Fixed the eventually function variable 
- Ensure the correct namespace is called 

### DIFF
--- a/test/e2e/placement_selecting_resources_test.go
+++ b/test/e2e/placement_selecting_resources_test.go
@@ -483,7 +483,7 @@ var _ = Describe("validating CRP when adding resources in a matching namespace",
 			workResourcesPlacedActual := func() error {
 				return validateWorkNamespaceOnCluster(memberCluster, workNamespaceName)
 			}
-			Eventually(workResourcesPlacedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place work namespace %s on member cluster %s", workNamespaceName, memberCluster.ClusterName)
+			Expect(workResourcesPlacedActual()).Should(Succeed(), "Failed to place work namespace %s on member cluster %s", workNamespaceName, memberCluster.ClusterName)
 		}
 	})
 
@@ -855,7 +855,7 @@ var _ = Describe("validating CRP when failed to apply resources", Ordered, func(
 
 var _ = Describe("validating CRP when placing cluster scope resource (other than namespace)", Ordered, func() {
 	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
-	clusterRoleName := fmt.Sprintf("test-cluserrole-%d", GinkgoParallelProcess())
+	clusterRoleName := fmt.Sprintf("reader-%d", GinkgoParallelProcess())
 
 	BeforeAll(func() {
 		By("creating cluster role")
--- a/test/e2e/placement_selecting_resources_test.go
+++ b/test/e2e/placement_selecting_resources_test.go
@@ -428,9 +428,9 @@ var _ = Describe("validating CRP when cluster-scoped and namespace-scoped resour
 
 var _ = Describe("validating CRP when adding resources in a matching namespace", Ordered, func() {
 	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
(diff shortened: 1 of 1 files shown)


test/e2e/placement_selecting_resources_test.go
```diff
@@ -483,7 +483,7 @@ var _ = Describe("validating CRP when adding resources in a matching namespace",
 			workResourcesPlacedActual := func() error {
 				return validateWorkNamespaceOnCluster(memberCluster, workNamespaceName)
 			}
-			Expect(workResourcesPlacedActual()).Should(Succeed(), "Failed to place work namespace %s on member cluster %s", workNamespaceName, memberCluster.ClusterName)
+			Eventually(workResourcesPlacedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place work namespace %s on member cluster %s", workNamespaceName, memberCluster.ClusterName)
```
zhiying-lin, line 486: are you sure this test is flaky?

here it should use expect, cause crp has already applied the resources.
no need to use eventually.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -856,7 +855,7 @@ var _ = Describe("validating CRP when failed to apply resources", Ordered, func(
 
 var _ = Describe("validating CRP when placing cluster scope resource (other than namespace)", Ordered, func() {
 	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
-	clusterRoleName := fmt.Sprintf("reader-%d", GinkgoParallelProcess())
+	clusterRoleName := fmt.Sprintf("test-cluserrole-%d", GinkgoParallelProcess())
```
zhiying-lin, line 858: will rename help to reduce the flakiness?

test/e2e/placement_selecting_resources_test.go
```diff
@@ -1347,7 +1346,7 @@ func checkIfPlacedLargeSecretResourcesOnTargetMemberClusters(targetMemberCluster
 		memberCluster := targetMemberClusters[idx]
 
 		secretsPlacedActual := secretsPlacedOnClusterActual(memberCluster)
-		Eventually(secretsPlacedActual(), largeEventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place large secrets on member cluster %s", memberCluster.ClusterName)
+		Eventually(secretsPlacedActual, largeEventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place large secrets on member cluster %s", memberCluster.ClusterName)
```
zhiying-lin, line 1349: good catch!

test/e2e/placement_selecting_resources_test.go
britaniar: it failed in these two commits:
- [feat: property-based scheduling: enable the property-based scheduling experience #2695](https://github.com/Azure/fleet/actions/runs/8506422376/job/23296596391)
- [feat: apply overrides before creating the work #2705](https://github.com/Azure/fleet/actions/runs/8517373471/job/23327835302)

test/e2e/placement_selecting_resources_test.go
britaniar: Will check again, but without the rename it was failing locally on my machine.

test/e2e/placement_selecting_resources_test.go
zhiying-lin: let me investigate it, this one looks like a bug.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -428,10 +428,10 @@ var _ = Describe("validating CRP when cluster-scoped and namespace-scoped resour
 
 var _ = Describe("validating CRP when adding resources in a matching namespace", Ordered, func() {
 	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+	ns := appNamespace()
```
ryanzhang-oss, line 431: why moving this out of the BeforeAll? Is there any difference?

zhiying-lin (approved):

PR #751, opened 2024-04-07
feat: Handle GVK validation on CRP Resource Selector

britaniar:

Fixes #

I have: added a GVK and cluster scope check to crp validator and corresponding unit test cases.


pkg/utils/validator/clusterresourceplacement_test.go
```diff
@@ -199,9 +274,10 @@ func TestValidateClusterResourcePlacementAlpha(t *testing.T) {
```
zhiying-lin, line 274: you can revert the v1alpha1 changes and you don't need to enhance the deprecating api

test/e2e/placement_selecting_resources_test.go
```diff
@@ -609,11 +609,10 @@ var _ = Describe("validating CRP when deleting resources in a matching namespace
 	})
 })
 
-// TODO should be blocked by the validation webhook
 var _ = Describe("validating CRP when resource selector is not valid", Ordered, func() {
 	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
 
-	BeforeAll(func() {
+	It("should be denied by webhook", func() {
```
zhiying-lin, line 615: add an e2e test if the resource selector is selecting namespace scope resource.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -634,52 +633,7 @@ var _ = Describe("validating CRP when resource selector is not valid", Ordered,
...
-				Name: crpName,
-			},
-		}
-		Expect(hubClient.Delete(ctx, crp)).To(Succeed(), "Failed to delete CRP %s", crpName)
-	})
-
-	It("should remove controller finalizers from CRP", func() {
-		finalizerRemovedActual := allFinalizersExceptForCustomDeletionBlockerRemovedFromCRPActual()
-		Eventually(finalizerRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove controller finalizers from CRP %s", crpName)
+		Eventually(hubClient.Create(ctx, crp), eventuallyDuration, eventuallyInterval).ShouldNot(Succeed(), "Validation webhook did catch invalid cluster resource in crp", crpName)
```
zhiying-lin, line 636: you don't need the eventually. The create should fail immediately + please check the error string,

eg,

err := hubClient.Create(ctx, crp)
		var statusErr *k8sErrors.StatusError
		Expect(errors.As(err, &statusErr)).To(BeTrue(), fmt.Sprintf("Create CRP call produced error %s. Error type wanted is %s.", reflect.TypeOf(err), reflect.TypeOf(&k8sErrors.StatusError{})))
		Expect(statusErr.Status().Message).Should(ContainSubstring(fmt.Sprintf("the labelSelector and name fields are mutually exclusive in selector %+v", selector[0])))

can you move this test to the webhook_test.go?

zhiying-lin (commented):
please revert any v1alpha1 changes, thanks

pkg/utils/common.go
```diff
@@ -128,6 +128,18 @@ var (
 		Resource: placementv1beta1.ClusterResourcePlacementResource,
 	}
 
+	ClusterRoleGVK = schema.GroupVersionKind{
```
zhiying-lin, line 131: can you move these definition to your test file or test/util package? these are not fleet related resources.

Prefer not to mix the test code with production code.

pkg/controllers/rollout/suite_test.go
```diff
@@ -91,6 +94,20 @@ var _ = BeforeSuite(func() {
...
+	validator.ResourceInformer = &informer.FakeManager{
+		APIResources: map[schema.GroupVersionKind]bool{
+			{
+				Group:   "",
+				Version: "v1",
+				Kind:    "Namespace",
+			}: true,
+		},
+		IsClusterScopedResource: true,
+	}
```
zhiying-lin, line 109: just curious why we need these. if i member it correctly, rollout controller does not call webhook?

britaniar: The rollout controller integration test [creates a crp](https://github.com/Azure/fleet/blob/3c13e7d99cf58f6a7e34a55ec559424fceb993b0/pkg/controllers/rollout/controller_integration_test.go#L66)

pkg/controllers/rollout/manifests/test_namespace.yaml
```diff
@@ -4,3 +4,4 @@ metadata:
   name: app
   labels:
     fleet.azure.com/name: test
+    region: east
```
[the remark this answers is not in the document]
britaniar: Looked into it and tested without it, looks like it is not needed.

pkg/controllers/rollout/suite_test.go
zhiying-lin: nope, we even did not create the webhook sever.

We need something like  https://github.com/Azure/fleet/blob/main/cmd/hubagent/main.go#L111 when creating the manager.

britaniar: As discussed offline, webhook is being called in rollouts controller.go file which is why the integration test was failing previously. Need to revisit and discuss if needed here.

pkg/controllers/rollout/controller.go
```diff
@@ -123,6 +123,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 
 	// fill out all the default values for CRP just in case the mutation webhook is not enabled.
 	fleetv1beta1.SetDefaultsClusterResourcePlacement(&crp)
+	// TODO: revisit to see if we should remove logic for validation
```
zhiying-lin, line 126: @ryanzhang-oss i'm thinking of removing this validation. The current validation is very heavy and we should assume validation webhook is enabled.

zhiying-lin (commented):
please resolve the file conflicts.

pkg/controllers/rollout/controller.go
ryanzhang-oss: SGTM as the webhook is fail to close

pkg/controllers/rollout/suite_test.go
zhiying-lin: statement is not accurate, > "validation is being called", not the webhook

When you use the "webhook" terminology, it's talking about 
[k8 webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#webhook-request-and-response)

zhiying-lin (approved):

PR #752, opened 2024-04-08
feat: update validation webhook for overrides

britaniar:

Fixes #

I have: 
- validate/test the "path" value in JSONPatchOverride
- validate/test remove op does not have a value in JSONPatchOverride
- validate/test empty and nil ClusterSelector


pkg/utils/validator/resourceoverride.go
```diff
@@ -84,26 +84,47 @@ func validateResourceOverrideResourceLimit(ro fleetv1alpha1.ResourceOverride, ro
...
+	}
+	return errors.NewAggregate(allErr)
+}
+
+// validateJSONPatchOverride checks if JSON patch override is valid.
+func validateJSONPatchOverride(jsonPatchOverrides []placementv1alpha1.JSONPatchOverride) error {
+	allErr := make([]error, 0)
+	for _, patch := range jsonPatchOverrides {
+		switch {
+		case strings.Contains(patch.Path, "/kind") || strings.Contains(patch.Path, "/apiVersion"):
```
zhiying-lin, line 118: what if the path is "/abc/kind" and "/apiVersionabc"?

we should not fail the request.

zhiying-lin (commented):
did not finish the review, please fix the obvious bugs and add more tests, will review later

pkg/utils/validator/resourceoverride_test.go
```diff
@@ -580,3 +682,100 @@ func TestValidateOverridePolicy(t *testing.T) {
...
+			jsonPatchOverrides: []placementv1alpha1.JSONPatchOverride{
+				{
+					Operator: placementv1alpha1.JSONPatchOverrideOpReplace,
+					Path:     "/spec/clusterResourceSelector/kind",
+					Value:    apiextensionsv1.JSON{Raw: []byte(`"ClusterRole"`)},
+				},
+			},
+			wantErrMsg: nil,
+		},
+		"invalid resource override path - cannot override typeMeta fields (kind)": {
```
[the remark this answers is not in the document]
britaniar: so "Kind" should pass?

pkg/utils/validator/resourceoverride_test.go
```diff
@@ -580,3 +682,100 @@ func TestValidateOverridePolicy(t *testing.T) {
...
+			jsonPatchOverrides: []placementv1alpha1.JSONPatchOverride{
```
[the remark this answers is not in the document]
britaniar: For either of those scenarios, those are supposed to pass, right? Along with the scenario where it could be `spec.resourceclusterselectors", would we want to check for periods or wait for the mutation webhook to change that to be correct path? For any of these scenarios, they should be rejected when trying to apply because they are not true paths.

pkg/utils/validator/resourceoverride_test.go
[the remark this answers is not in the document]
zhiying-lin: i don't think it will pass, could you please validate it by trying the applyJSONPatchOverride UT?

pkg/utils/validator/resourceoverride_test.go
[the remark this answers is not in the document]
zhiying-lin: For these two scenarios, they're invalid. So we should reject them.

A valid path should begin with "/" based on the rfc6902.

For the "spec.resourceclusterselectors", cx could create a field itself called "spec.resourceclusterselectors", If the path is "/spec.resourceclusterselectors", we could allow and controller will check during the runtime.

pkg/utils/validator/resourceoverride_test.go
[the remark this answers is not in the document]
britaniar: Should not pass, error: Failed to apply the JSON patch to the resource" err="replace operation does not apply: doc is missing key: /Kind: missing value.

pkg/utils/validator/resourceoverride.go
```diff
@@ -112,16 +113,24 @@ func validateOverridePolicy(policy *fleetv1alpha1.OverridePolicy) error {
 func validateJSONPatchOverride(jsonPatchOverrides []fleetv1alpha1.JSONPatchOverride) error {
```
zhiying-lin, line 113:
```
// validateJSONPatchOverride checks if JSON patch override is valid.
func validateJSONPatchOverride(jsonPatchOverrides []fleetv1alpha1.JSONPatchOverride) error {
	allErr := make([]error, 0)
	for _, patch := range jsonPatchOverrides {
		if err := validateJSONPatchOverridePath(patch.Path); err != nil {
			allErr = append(allErr, fmt.Errorf("invalid JSONPatchOverride %s: %v", patch, err))
		}

		if patch.Operator == fleetv1alpha1.JSONPatchOverrideOpRemove && len(patch.Value.Raw) != 0 {
			allErr = append(allErr, fmt.Errorf("invalid JSONPatchOverride %s: remove operation cannot have value", patch))
		}
	}
	return nil
}

func validateJSONPatchOverridePath(path string) error {
	if path == "" {
		return fmt.Errorf("path cannot be empty")
	}

	if !strings.HasPrefix(path, "/") {
		return fmt.Errorf("path must start with /")
	}

	// The path begins with a slash, and at least there will be two elements.
	parts := strings.Split(path, "/")[1:]
	switch parts[0] {
	case "kind", "apiVersion":
		return fmt.Errorf("cannot override typeMeta fields")
	case "metadata":
		if len(path) == 1 {
			return fmt.Errorf("cannot override metadata fields")
		} else if parts[1] != "annotations" && parts[1] != "labels" {
			return fmt.Errorf("cannot override metadata fields except annotations and labels")
		}
	case "status":
		return fmt.Errorf("cannot override status fields")
	}

	for i := range parts {
		if len(strings.TrimSpace(parts[i])) == 0 {
			return fmt.Errorf("path cannot contain empty string")
		}
	}
	return nil
}
```

please refer to this code

pkg/utils/validator/resourceoverride.go
```diff
@@ -87,23 +88,59 @@ func validateResourceOverrideResourceLimit(ro fleetv1alpha1.ResourceOverride, ro
...
+
+		if slashPattern.MatchString(patch.Path) {
+			allErr = append(allErr, fmt.Errorf("invalid JSONPatchOverride %s: path cannot contain consecutive slashes", patch))
+		}
+
+		if patch.Path == "/kind" || patch.Path == "/apiVersion" {
+			allErr = append(allErr, fmt.Errorf("invalid JSONPatchOverride %s: cannot override typeMeta fields", patch))
+		}
+
+		if strings.HasPrefix(patch.Path, "/metadata") && !strings.HasPrefix(patch.Path, "/metadata/annotations") && !strings.HasPrefix(patch.Path, "/metadata/labels") {
```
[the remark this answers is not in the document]
zhiying-lin: add a test case?

pkg/utils/validator/resourceoverride_test.go
```diff
@@ -537,39 +621,86 @@ func TestValidateOverridePolicy(t *testing.T) {
...
+										},
+									},
+								},
+							},
+						},
+						JSONPatchOverrides: []fleetv1alpha1.JSONPatchOverride{},
+					},
+				},
+			},
+			wantErrMsg: errors.New("JSONPatchOverrides cannot be empty"),
```
zhiying-lin, line 698: add one invalid JSONPatchOverrides test?

pkg/utils/validator/resourceoverride_test.go
```diff
@@ -580,3 +711,245 @@ func TestValidateOverridePolicy(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateJSONPatchOverride(t *testing.T) {
+	tests := map[string]struct {
+		jsonPatchOverrides []fleetv1alpha1.JSONPatchOverride
+		wantErrMsg         error
+	}{
```
zhiying-lin, line 719: you don't need to repeat the tests about invalid path as you've already covered that in TestValidateJSONPatchOverridePath.

for this func, focus on
- empty jsonPatchOverrides
- nil jsonPathOverrides
- one invalid json path
- invalid remove operation
- valid jsonPathOverrides

pkg/utils/validator/resourceoverride_test.go
```diff
@@ -580,3 +711,245 @@ func TestValidateOverridePolicy(t *testing.T) {
...
+		},
+		"invalid json patch override path - slashes only": {
+			path:       "/////",
+			wantErrMsg: errors.New("path cannot contain empty string"),
+		},
+		"invalid json patch override path - path must start with /": {
+			path:       "spec.resourceSelectors/selectors/0/name",
+			wantErrMsg: errors.New("path must start with /"),
+		},
+		"valid json patch override path - apiVersion within field": {
```
zhiying-lin, line 922: duplicate test case in line 906

pkg/utils/validator/resourceoverride_test.go
```diff
@@ -580,3 +711,245 @@ func TestValidateOverridePolicy(t *testing.T) {
...
+		},
+		"invalid json patch override path - invalid metadata field": {
+			path:       "/metadata/annotationsabc",
+			wantErrMsg: errors.New("cannot override metadata fields"),
+		},
+		"invalid json patch override path - contains empty string": {
+			path:       "/spec/clusterNames///member-1",
+			wantErrMsg: errors.New("path cannot contain empty string"),
+		},
+		"valid json override patch - case sensitive check": {
```
zhiying-lin, line 938: duplicate test case in line 898

pkg/utils/validator/resourceoverride.go
```diff
@@ -80,31 +81,82 @@ func validateResourceOverrideResourceLimit(ro fleetv1alpha1.ResourceOverride, ro
...
+		return fmt.Errorf("path must start with /")
+	}
+
+	// The path begins with a slash, and at least there will be two elements.
+	parts := strings.Split(path, "/")[1:]
+	switch parts[0] {
+	case "kind", "apiVersion":
+		return fmt.Errorf("cannot override typeMeta fields")
+	case "metadata":
+		if len(path) == 1 {
```
zhiying-lin, line 147:
```suggestion
		if len(parts) == 1 {
```
it should be parts.

zhiying-lin: can you add a test? just curious about why we did not catch this

zhiying-lin (approved):

PR #754, opened 2024-04-09
feat: watch MC taint update/delete

Arvindthiru:

Fixes #

I have:

The PR is to trigger a scheduling cycle when a taint is updated/deleted on a MC so that CRPs with pickAll, pickN placement policies can pick cluster

### DIFF
--- a/pkg/scheduler/watchers/membercluster/controller_integration_test.go
+++ b/pkg/scheduler/watchers/membercluster/controller_integration_test.go
@@ -444,7 +444,7 @@ var _ = Describe("scheduler member cluster source controller", Serial, Ordered, 
 		})
 	})
 
-	Context("ready cluster has taint added & removed", func() {
+	Context("ready cluster has taint added, updated & removed", func() {
 		BeforeAll(func() {
 			Consistently(noKeyEnqueuedActual, consistentlyDuration, consistentlyInterval).Should(Succeed(), "Workqueue is not empty")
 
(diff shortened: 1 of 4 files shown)


pkg/scheduler/watchers/membercluster/controller_integration_test.go
```diff
@@ -444,6 +444,41 @@ var _ = Describe("scheduler member cluster source controller", Serial, Ordered,
...
+			// Retrieve the cluster.
+			Expect(hubClient.Get(ctx, types.NamespacedName{Name: clusterName1}, memberCluster)).To(Succeed(), "Failed to get member cluster")
+
+			// Remove taint from cluster.
+			memberCluster.Spec.Taints = nil
+			Expect(hubClient.Update(ctx, memberCluster)).Should(Succeed(), "Failed to update member cluster taints")
+		})
+
+		It("should enqueue CRPs for cluster removed taint (case 1c)", func() {
+			Eventually(qualifiedKeysEnqueuedActual(), eventuallyDuration, eventuallyInterval).Should(Succeed(), "Keys are not enqueued as expected")
```
zhiying-lin, line 473:
```suggestion
			Eventually(qualifiedKeysEnqueuedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Keys are not enqueued as expected")
```

You should not call the func, otherwise, it will populate the result. so the func will be called just one time.

You need to pass in the func instead.

pkg/scheduler/watchers/membercluster/controller_integration_test.go
```diff
@@ -444,6 +444,63 @@ var _ = Describe("scheduler member cluster source controller", Serial, Ordered,
 		})
 	})
 
+	Context("ready cluster has taint added, updated & removed", func() {
```
ryanzhang-oss, line 447: we need a case that the new taint added won't kick the existing placed resources out of the cluster

Arvindthiru: This test covers the case described https://github.com/Azure/fleet/blob/main/test/scheduler/tainttoleration_integration_test.go#L210-L281

ryanzhang-oss: interesting, we had an add taint test but didn't implement remove taint.

ryanzhang-oss (approved):

PR #755, opened 2024-04-10
feat: override manifests when creating works

zhiying-lin:

Calling the override util when creating the works

Fixes #

I have:

added integration tests


pkg/controllers/workgenerator/suite_test.go
```diff
@@ -96,11 +110,177 @@ var _ = BeforeSuite(func() {
...
 
+	appNamespace = corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: appNamespaceName,
+		},
+	}
+	Expect(k8sClient.Create(ctx, &appNamespace)).Should(Succeed(), "Failed to create the application namespace")
+	By(fmt.Sprintf("Application namespace %s created", appNamespaceName))
+
+	validClusterResourceOverrideSnapshot = placementv1alpha1.ClusterResourceOverrideSnapshot{
```
ryanzhang-oss, line 147: can this and the rest of the 2 big var definition be moved to an init function?

pkg/controllers/workgenerator/controller.go
```diff
@@ -100,40 +101,83 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+	cluster := clusterv1beta1.MemberCluster{}
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: resourceBinding.Spec.TargetCluster}, &cluster); err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(2).InfoS("Skip reconciling clusterResourceBinding when the cluster is deleted", "memberCluster", resourceBinding.Spec.TargetCluster, "clusterResourceBinding", bindingRef)
+			return controllerruntime.Result{}, nil
+		}
+		klog.ErrorS(err, "Failed to get the memberCluster", "memberCluster", resourceBinding.Spec.TargetCluster, "clusterResourceBinding", bindingRef)
+		return controllerruntime.Result{}, controller.NewAPIServerError(true, err)
+	}
+
```
ryanzhang-oss, line 117: I wonder why is this before the finalizer?

pkg/controllers/workgenerator/controller.go
```diff
@@ -173,6 +218,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 		klog.ErrorS(updateErr, "Failed to update the resourceBinding status", "resourceBinding", bindingRef)
 		return controllerruntime.Result{}, controller.NewUpdateIgnoreConflictError(updateErr)
 	}
+	if errors.Is(syncErr, controller.ErrUserError) {
```
ryanzhang-oss, line 221: what type of error is user error?  looks like we don't set the sync error in the condition when the override error !=nil? In that case, how can user know what's going on?

pkg/controllers/workgenerator/controller.go
```diff
@@ -100,40 +101,83 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+
+		if !overrideSucceeded {
+			resourceBinding.SetConditions(metav1.Condition{
+				Status:             metav1.ConditionFalse,
+				Type:               string(fleetv1beta1.ResourceBindingOverridden),
+				Reason:             condition.OverriddenFailedReason,
+				Message:            fmt.Sprintf("Failed to apply the override rules on the resources: %s", errorMessage),
+				ObservedGeneration: resourceBinding.Generation,
+			})
+		} else {
```
ryanzhang-oss, line 164: shouldn't we return all the errors?

pkg/controllers/workgenerator/controller.go
```diff
@@ -315,10 +384,10 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding *fleetv1be
 		newWork = append(newWork, work)
 
 		// issue all the create/update requests for the corresponding works for each snapshot in parallel
-		for i := range newWork {
-			work := newWork[i]
+		for ni := range newWork {
+			w := newWork[ni]
```
ryanzhang-oss, line 388: any particular reason for the local var name change?

pkg/controllers/workgenerator/controller.go
zhiying-lin: if override fails, we will stop creating/updating the work. So the workSynchronized condition will be always false.

pkg/controllers/workgenerator/controller.go
zhiying-lin: for now, we'll set user error when we fail to apply the overrides or fail to extract resources from the configmap.

they can track the details in the failedManifest.

pkg/controllers/workgenerator/controller.go
zhiying-lin: try to avoid the shadow. in case, we accidently use "=" instead of ":=" and the value will be overwritten.

ryanzhang-oss (approved):

PR #756, opened 2024-04-11
feat: check ready status by using avail condition

zhiying-lin:

Use the avail condition instead of applied

Fixes #

I have:

UT


pkg/controllers/rollout/controller_integration_test.go
```diff
@@ -158,10 +158,10 @@ var _ = Describe("Test the rollout Controller", func() {
 			}
 			return true
 		}, timeout, interval).Should(BeTrue(), "rollout controller should roll all the bindings to Bound state")
-		// simulate that some of the bindings are applied
```
ryanzhang-oss, line 161: do we still have an integration test with availabe but not trackable one?

ryanzhang-oss (approved):

PR #757, opened 2024-04-11
interface: add v1 APIs

michaelawyu:

This PR adds the v1 APIs for the GA task.

I have:

N/A.

- The v1 APIs at this moment are exactly the same with the v1beta1 APIs; per earlier agreements, we would be using the v1beta1 APIs as the storage version
- No conversion strategy has been configured, so K8s API server will simply switch the API version as appropriate.


apis/placement/v1/commons.go
```diff
@@ -0,0 +1,90 @@
...
+	ClusterSchedulingPolicySnapshotKind = "ClusterSchedulingPolicySnapshot"
+	WorkKind                            = "Work"
+	AppliedWorkKind                     = "AppliedWork"
+)
+
+const (
+	// Unprefixed labels/annotations are reserved for end-users
+	// we will add a kubernetes-fleet.io to designate these labels/annotations as official fleet labels/annotations.
+	// See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions
+	fleetPrefix = "kubernetes-fleet.io/"
```
zhiying-lin, line 22: need to think about putting them into a common package. otherwise, each version has its own consts

zhiying-lin (commented):
Can you do a manual test in the upstream to create a v1 & v1beta1 crp? and then get both v1 & v1beta1 to make sure they're working as expected.

apis/placement/v1/commons.go
michaelawyu: I guess it kind of depends on how long we plan to keep the v1beta1 version; if not long I guess duplication would be OK?

apis/cluster/v1/membercluster_types.go
```diff
@@ -0,0 +1,209 @@
...
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+
+	// Conditions is an array of current observed conditions for the member cluster.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// Properties is an array of properties observed for the member cluster.
```
ryanzhang-oss, line 88: This is a preview feature, I think we need to mark this filed as preview like k8s did

apis/cluster/v1/commons.go
```diff
@@ -0,0 +1,123 @@
...
+
+type ClusterState string
+
+const (
+	ClusterStateJoin  ClusterState = "Join"
+	ClusterStateLeave ClusterState = "Leave"
+)
+
+// ResourceUsage contains the observed resource usage of a member cluster.
+type ResourceUsage struct {
```
ryanzhang-oss, line 22: some of the fields here are also in preview

apis/placement/v1/commons.go
ryanzhang-oss: realistically, I think it will take at least 6 months

apis/placement/v1/clusterresourceplacement_types.go
```diff
@@ -0,0 +1,853 @@
...
+	// +optional
+	PropertySelector *PropertySelector `json:"propertySelector,omitempty"`
+
+	// PropertySorter sorts all matching clusters by a specific property and assigns different weights
+	// to each cluster based on their observed property values.
+	//
+	// At this moment, PropertySorter can only be used with
+	// `PreferredDuringSchedulingIgnoredDuringExecution` affinity terms.
+	// +optional
+	PropertySorter *PropertySorter `json:"propertySorter,omitempty"`
```
ryanzhang-oss, line 350: these two are preview

apis/cluster/v1/membercluster_types.go
michaelawyu: Sure, I will add the comments. These would apply for v1beta1 as well I assume? (i.e., alpha level field)

michaelawyu: Also, do we do the same for other features, e.g., taints as well?

michaelawyu: Per earlier discussion, a beta-level hint comment is added.

apis/placement/v1/clusterresourceplacement_types.go
michaelawyu: Per earlier discussions, a beta-level hint comment is added.

apis/cluster/v1/commons.go
michaelawyu: Per earlier discussions, a beta-level hint comment is added.

apis/placement/v1/clusterresourceplacement_types.go
```diff
@@ -338,6 +338,9 @@ type ClusterSelectorTerm struct {
 	//
```
zhiying-lin, line 338: missing tolerations hint

michaelawyu: Hi Zhiying! The hint for the tolerations is added in the CRP field part (line 164) rather than the struct comment part (line 667) as the struct comment is not incorporated into the CRD.

michaelawyu:
As discussed earlier, the override fields are marked as alpha; fields for other new features receive the beta tag.

ryanzhang-oss (approved):

PR #764, opened 2024-04-18
fix: reconcile the binding status

zhiying-lin:

Rollout controller needs to update the rolloutStarted condition if there is no binding to rollout

Fixes #

I have:

### DIFF
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -141,7 +141,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 
 	if !needRoll {
 		klog.V(2).InfoS("No bindings are out of date, stop rolling", "clusterResourcePlacement", crpName)
-		return runtime.Result{}, nil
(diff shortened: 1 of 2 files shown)


pkg/controllers/rollout/controller.go
```diff
@@ -141,7 +141,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 
 	if !needRoll {
 		klog.V(2).InfoS("No bindings are out of date, stop rolling", "clusterResourcePlacement", crpName)
-		return runtime.Result{}, nil
+		// There will be a corner case that rollout controller succeeds to update the binding spec to the latest one, but
+		// fail to update the binding conditions.
+		// Here it will reconcile the binding status.
```
ryanzhang-oss, line 146:
```suggestion
		// There will be a corner case that rollout controller succeeds to update the binding spec to the latest one, 
		// but fails to update the binding conditions when it reconciled it last time.
		// Here it will correct the binding status just in case this happens last time.
```

ryanzhang-oss (approved):

PR #776, opened 2024-04-24
fix: add apply strategy in work

zhiying-lin:

Fixes #

I have:

e2e tests

### DIFF
--- a/test/e2e/placement_apply_strategy_test.go
+++ b/test/e2e/placement_apply_strategy_test.go
@@ -1,0 +1,294 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package e2e
+
+import (
+	"fmt"
+
+	"github.com/google/go-cmp/cmp"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"go.goms.io/fleet/pkg/controllers/work"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
+)
+
+var _ = Describe("validating CRP when resources exists", Ordered, func() {
+	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+	annotationKey := "annotation-key"
+	annotationValue := "annotation-value"
+	workNamespaceName := fmt.Sprintf(workNamespaceNameTemplate, GinkgoParallelProcess())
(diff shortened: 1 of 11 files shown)


test/e2e/placement_apply_strategy_test.go
```diff
@@ -0,0 +1,294 @@
...
+				},
+			})
+			ns.Annotations = map[string]string{
+				annotationKey: annotationValue,
+			}
+			By(fmt.Sprintf("creating namespace %s on member cluster", ns.Name))
+			Expect(memberCluster1EastProdClient.Create(ctx, &ns)).Should(Succeed(), "Failed to create namespace %s", ns.Name)
+
+			// Create the CRP.
+			strategy := &placementv1beta1.ApplyStrategy{AllowCoOwnership: true}
```
ryanzhang-oss, line 59: can we also test the case it's not filled?

test/e2e/placement_apply_strategy_test.go
```diff
@@ -0,0 +1,294 @@
...
+		// This check will ignore the annotation of resources.
+		It("should place the selected resources on member clusters", checkIfPlacedWorkResourcesOnAllMemberClusters)
+
+		It("should have annotations on the namespace", func() {
+			want := map[string]string{annotationKey: annotationValue}
+			Expect(validateAnnotationOfWorkNamespaceOnCluster(memberCluster1EastProd, want)).Should(Succeed(), "Failed to override the annotation of work namespace on %s", memberCluster1EastProdName)
+		})
+
+		It("can delete the CRP", func() {
+			// Delete the CRP.
```
ryanzhang-oss, line 82: we delete crp in the afterAll block

test/e2e/placement_apply_strategy_test.go
zhiying-lin: we do have one test in https://github.com/Azure/fleet/blob/main/test/e2e/placement_selecting_resources_test.go#L691

test/e2e/placement_apply_strategy_test.go
zhiying-lin: each test will delete the crp in the afterAll block in line 65 to make sure we have a clean test env.

test/e2e/setup.sh
```diff
@@ -118,6 +118,7 @@ helm install hub-agent ../../charts/hub-agent/ \
     --set image.repository=$REGISTRY/$HUB_AGENT_IMAGE \
     --set image.tag=$TAG \
     --set namespace=fleet-system \
+    --set logVerbosity=5 \
```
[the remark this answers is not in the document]
zhiying-lin: i thought the default value is 2.
it's ok to set it explicitly in case we change the default value.

ryanzhang-oss (approved):

PR #798, opened 2024-05-08
docs: update troubleshooting guide with latest conditions

britaniar:

Fixes #

I have:


Arvindthiru:
Couple of suggestions,

- This TSG is huge (in terms of cognitive load) and since we will be using this as our primary TSG it would be better to split each condition into a separate .md file
- Please add a list of common error scenarios for all conditions
- And since conditions are updated in order, we can also suggest the users to looks at specific controllers when a condition is set to false for example,
-- Scheduled condition is false something is wrong within the scheduler
-- RolloutStarted failed something went wrong in the rollout controller
-- WorkApplied is faise something went wrong when applying manifest in apply controller

docs/troubleshooting/clusterResourcePlacementScheduled.md
```diff
@@ -0,0 +1,199 @@
+# How can I debug when my CRP status is ClusterResourcePlacementScheduled condition status is set to false?
+The `ClusterResourcePlacementScheduled` condition is set to `false` when the scheduler cannot find all the clusters needed as specified by the scheduling policy.
+> Note: In addition, it may be helpful to look into the logs for the scheduler controller to get more information on why the scheduling failed.
```
Arvindthiru, line 3: nit: scheduler is not a controller is a separate component maybe rename to scheduler framework

docs/troubleshooting/clusterResourcePlacementOverridden.md
```diff
@@ -0,0 +1,156 @@
...
+  selectedResources:
+  - group: rbac.authorization.k8s.io
+    kind: ClusterRole
+    name: secret-reader
+    version: v1
+```
+The CRP attempted to override a propagated resource utilizing an applicable `ClusterResourceOverrideSnapshot`.
+However, as the `ClusterResourcePlacementOverridden` condition remains false, looking at the placement status for the cluster
+where the condition `Overriden` failed will offer insights into the exact cause of the failure.
+The accompanying message highlights that the override failed due to the absence of the path `/metadata/labels/new-label` and its corresponding value.
```
Arvindthiru, line 144: Could we please spell it out, when we use metadata/labels/new-label we are trying to modify/add a label with the value we provide if the labels field is not empty. But in the case /metadata/labels we are replacing the field with the value we provide

docs/troubleshooting/clusterResourcePlacementWorkSynchronized.md
```diff
@@ -0,0 +1,114 @@
...
+  - kind: Namespace
+    name: test-ns
+    version: v1
+```
+The `ClusterResourcePlacementWorkSynchronized` condition in the CRP status is flagged as false. It is clear from the message
+that the work object `crp1-work` is prohibited from generating new content within the namespace `fleet-member-kind-cluster-1`
+as it's currently undergoing termination.
+
+### Resolution:
+- To address this specific issue, recreate the Custom Resource Placement (CRP) with a newly selected cluster.
```
Arvindthiru, line 111: - Users can also modify the CRP to select a new cluster
- Deleting the CRP, deletes work through garbage collection
- Namespace can regenerate only if cluster is re-joined so we can just ask user to re-join member cluster instead

nit: Custom Resource Placement -> Cluster Resource Placement

docs/troubleshooting/clusterResourcePlacementWorkSynchronized.md
```diff
@@ -0,0 +1,114 @@
...
+- Alternatively, delete the CRP and any work on the namespace, then wait for the namespace to regenerate
+
+In other scenarios, you might opt to wait for the work to finish propagating. If the issue persists, consider deleting the CRP and recreating it.
```
Arvindthiru, line 114: - We can mention latency as a reason to ask users to wait till work to finish propagating
- Deleting the CRP and re-creating it shouldn't fix any issue by itself in most cases we either wait or ask user to take other actions to resolve right ?

docs/troubleshooting/clusterResourcePlacementApplied.md
```diff
@@ -0,0 +1,239 @@
+# How can I debug when my CRP ClusterResourcePlacementApplied condition is set to false?
+
+### Common scenarios:
+- When the CRP is unable to propagate resources to a selected cluster due to the resource already existing on the cluster and not being managed by the fleet controller.
```
Arvindthiru, line 4: nit: We can mention co-ownership option here

docs/troubleshooting/clusterResourcePlacementApplied.md
```diff
@@ -0,0 +1,239 @@
+# How can I debug when my CRP ClusterResourcePlacementApplied condition is set to false?
+
+### Common scenarios:
+- When the CRP is unable to propagate resources to a selected cluster due to the resource already existing on the cluster and not being managed by the fleet controller.
+- When the CRP is unable to propagate resource to selected due to another CRP already managing the resource for selected cluster with a different apply strategy.
+- When the CRP is unable to propagate resource due to failing to apply manifest due to syntax errors or invalid resource configurations.
```
Arvindthiru, line 6: For syntax errors it could only happen if the resource is being propagated through an envelope object, right?

docs/troubleshooting/clusterResourcePlacementAvailable.md
```diff
@@ -0,0 +1,205 @@
+# How can I debug when my CRP ClusterResourcePlacementAvailable condition is set to false?
+The ClusterResourcePlacementAvailable condition is false when the cluster lacks the necessary resources or capabilities to accommodate new deployments or allocations.
```
Arvindthiru, line 2: "lacks the necessary resources or capabilities to accommodate new deployments or allocations" the resource cannot be available for other reasons as well Ex. like bad image name in the spec for deployment

docs/troubleshooting/clusterResourcePlacementAvailable.md
```diff
@@ -0,0 +1,205 @@
...
+    kind: Deployment
+    name: my-deployment
+    namespace: test-ns
+    ordinal: 1
+    resource: deployments
+    version: v1
+```
+Looking at the status `Available` condition for `kind-cluster-1`, we see that the deployment `my-deployment` is not available yet on the member cluster.
+Therefore, there might be something wrong with the deployment manifest.
+
```
Arvindthiru, line 199: Asking the user to take a look at the deployment in the member cluster is something worth mentioning since in this it would clearly mention the root cause is the bad image name

docs/troubleshooting/clusterResourcePlacementAvailable.md
```diff
@@ -0,0 +1,205 @@
...
+#### Resolution:
+In this scenario, a viable solution is to rectify the deployment manifest by verifying that all fields are accurately specified.
+After fixing the resource manifest and updating it, it's essential to delete the CRP and then reapply or recreate it. This step ensures that the changes made to the resource manifest are properly reflected in the CRP configuration.
```
Arvindthiru, line 202: Why do we have to delete the CRP here, the new image name changes will be automatically propagated by the existing CRP

docs/troubleshooting/clusterResourcePlacementApplied.md
britaniar: Yes

ryanzhang-oss (approved):

PR #802, opened 2024-05-08
fix: only mark a binding apply failed if there are failed manifest

ryanzhang-oss:

Currently, we mark a binding failed if not all work are applied. This causes the CRP controller to try to fill in failed placement from work. So we only mark a binding apply failed if there are failed manifest, CRP condition won't change. This is probably also useful when we start to copy failed manifest to the binding.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -173,12 +173,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 			Type:               string(fleetv1beta1.ResourceBindingWorkSynchronized),
 			Reason:             condition.AllWorkSyncedReason,
 			ObservedGeneration: resourceBinding.Generation,
-			Message:            "All of the works are synchronized to the latest",
+			Message:            "All of the works are synchronized to the latest resource snapshot",
 		})
 		if workUpdated {
 			// revert the applied condition if we made any changes to the work
 			resourceBinding.SetConditions(metav1.Condition{
-				Status:             metav1.ConditionFalse,
+				Status:             metav1.ConditionUnknown,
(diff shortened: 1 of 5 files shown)


pkg/controllers/workgenerator/controller.go
```diff
@@ -628,8 +634,17 @@ func buildAllWorkAppliedCondition(works map[string]*fleetv1beta1.Work, binding *
 			ObservedGeneration: binding.GetGeneration(),
 		}
 	}
+	if oneFailedToApply {
+		return metav1.Condition{
+			Status:             metav1.ConditionFalse,
+			Type:               string(fleetv1beta1.ResourceBindingApplied),
+			Reason:             condition.WorkNotAppliedReason,
```
zhiying-lin, line 641: should we use "WorkFailedReason"?

pkg/controllers/workgenerator/controller.go
```diff
@@ -628,8 +634,17 @@ func buildAllWorkAppliedCondition(works map[string]*fleetv1beta1.Work, binding *
...
+			Message:            fmt.Sprintf("Work object %s is not applied", notAppliedWork),
```
zhiying-lin, line 642:
```suggestion
			Message:            fmt.Sprintf("Failed to apply work object %s", notAppliedWork),
```

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -170,7 +170,7 @@ func TestBuildAllWorkAppliedCondition(t *testing.T) {
 				ObservedGeneration: 1,
 			},
 		},
-		"applied should be false if not all work applied to the latest generation": {
+		"applied should be known if not all work applied to the latest generation": {
```
zhiying-lin, line 173:
```suggestion
		"applied should be unknown if not all work applied to the latest generation": {
```

PR #803, opened 2024-05-08
fix: bump the min maxUnavailable to 1

ryanzhang-oss:

Bump the min maxUnavailable to 1 to avoid rolling out stuck.

Fixes #

I have:

### DIFF
--- a/apis/placement/v1/clusterresourceplacement_types.go
+++ b/apis/placement/v1/clusterresourceplacement_types.go
@@ -500,7 +500,7 @@ type RollingUpdateConfig struct {
 	// Absolute number is calculated from percentage by rounding up.
 	// We consider a resource unavailable when we either remove it from a cluster or in-place
 	// upgrade the resources content on the same cluster.
-	// This can not be 0 if MaxSurge is 0.
+	// The minimum of MaxUnavailable is 1 to avoid rolling out stuck during in-place resource update.
 	// Defaults to 25%.
 	// +kubebuilder:default="25%"
 	// +kubebuilder:validation:XIntOrString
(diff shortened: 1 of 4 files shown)


apis/placement/v1/clusterresourceplacement_types.go
```diff
@@ -500,7 +500,7 @@ type RollingUpdateConfig struct {
 	// Absolute number is calculated from percentage by rounding up.
 	// We consider a resource unavailable when we either remove it from a cluster or in-place
 	// upgrade the resources content on the same cluster.
-	// This can not be 0 if MaxSurge is 0.
+	// The minimum of MaxUnavailable is 1 to avoid rolling out stuck during in-place resource update.
```
circy9, line 503: // The desired number equals to the `NumberOfClusters` field when the placement type is `PickN`.
	// The desired number equals to the number of clusters scheduler selected when the placement type is `PickAll`.
	
What is the desired number? Why does it equal to max possible number?

apis/placement/v1/clusterresourceplacement_types.go
```diff
@@ -500,7 +500,7 @@ type RollingUpdateConfig struct {
...
 	// Defaults to 25%.
 	// +kubebuilder:default="25%"
 	// +kubebuilder:validation:XIntOrString
```
circy9, line 506: Does this work?

+kubebuilder:validation:ExclusiveMinimum:true
+kubebuilder:validation:Minimum:0

apis/placement/v1/clusterresourceplacement_types.go
```diff
@@ -513,7 +513,7 @@ type RollingUpdateConfig struct {
 	// The desired number equals to the number of clusters scheduler selected when the placement type is `PickAll`.
```
circy9, line 513: // The desired number equals to the `NumberOfClusters` field when the placement type is `PickN`.

What is the desired number?

apis/placement/v1/clusterresourceplacement_types.go
ryanzhang-oss: it is used to calculate the real number when the maxUnavailable is a percent like 25%

apis/placement/v1/clusterresourceplacement_types.go
ryanzhang-oss: I dobut it as this is not an int

circy9 (approved):

PR #808, opened 2024-05-10
fix: validation for property selector name

Arvindthiru:

Fixes #

I have:

This validation is to prevent users from providing random property names with prefix resources.kubernetes-fleet.io/

### DIFF
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -40,8 +40,8 @@ var (
 	// associated with all resource properties.
 	resourcePropertyNamePrefix = "resources.kubernetes-fleet.io/"
 
-	// Below are a list of supported capacity types.
-	supportedResourceCapacityTypes = []string{"total", "allocatable", "available"}
+	// Below is the map of supported capacity types.
+	supportedResourceCapacityTypes = map[string]bool{"total": true, "allocatable": true, "available": true}
 )
 
 // ValidateClusterResourcePlacementAlpha validates a ClusterResourcePlacement v1alpha1 object.
@@ -452,7 +452,7 @@ func validateName(name string) error {
 		if len(segments) != 2 {
 			return fmt.Errorf("invalid resource property name %s, expected format is [PREFIX]/[CAPACITY_TYPE]-[RESOURCE_NAME]", name)
 		}
-		if !isValidCapacityType(segments[0]) {
+		if !supportedResourceCapacityTypes[segments[0]] {
 			return fmt.Errorf("invalid capacity type in resource property name %s, supported values are %+v", name, supportedResourceCapacityTypes)
 		}
 	}
@@ -486,13 +486,4 @@ func validateValues(values []string) error {
 		}
 	}
 	return nil
-}
-
-func isValidCapacityType(ct string) bool {
-	for _, resourceCapacityType := range supportedResourceCapacityTypes {
-		if ct == resourceCapacityType {
-			return true
-		}
-	}
-	return false
 }
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -447,10 +447,15 @@ func validateName(name string) error {
 	if err := validation.IsQualifiedName(name); err != nil {
 		return fmt.Errorf("name is not a valid Kubernetes label name: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -446,3 +460,12 @@ func validateValues(values []string) error {
 	}
 	return nil
 }
+
+func isValidResourcePropertyName(resourcePropertyName string) bool {
```
michaelawyu, line 464: Hi Arvind! There's actually one more part in the property name, e.g., `resources.kubernetes-fleet.io/total-cpu`.

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -419,6 +427,17 @@ func validateName(name string) error {
 	if err := validation.IsQualifiedName(name); err != nil {
 		return fmt.Errorf("name is not a valid Kubernetes label name: %v", err)
 	}
+	// we expect the resource property names to be in this format `[PREFIX]/[CAPACITY_TYPE]-[RESOURCE_NAME]`.
+	if strings.HasPrefix(name, resourcePropertyNamePrefix) {
+		resourcePropertyName, _ := strings.CutPrefix(name, resourcePropertyNamePrefix)
+		segments := strings.Split(resourcePropertyName, "-")
+		if len(segments) != 2 || len(segments[0]) == 0 || len(segments[1]) == 0 {
```
Arvindthiru, line 434: Had to replicate some validation logic from here https://github.com/Azure/fleet/blob/6d59196736446457aa9ff7b310fb150d5bca1c67/pkg/scheduler/framework/plugins/clusteraffinity/types.go#L43-L51, but it feels like the validation belongs here. Not removing the original validation as a safety measure. But since the webhook is fail close for CRP it never should reach the origin scheduler validation logic

michaelawyu: Hi Arvind! As mentioned in Teams, considering that webhook is optional and we will be having upstream only users as well, it might not be a good idea to assume that webhook will be able catch all the validation errors for us. As for the alert issue, it might be better to update the code to mark the error as a user one I think.

Arvindthiru: Need to create a new user error object to be able to differentiate between the errors returned, so that we can bubble the error as a framework status at the top layer. Will handle it in a separate PR

Arvindthiru: Need to classify errors as user errors https://github.com/Azure/fleet/blob/28e5a083bab5d437888b281cf0cf1c76a32ee0f1/pkg/scheduler/framework/plugins/clusteraffinity/types.go#L98-L110

michaelawyu, line 434: Hi Arvind! We also have some requirements on the second argument (the resource name).

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -34,6 +36,14 @@ var (
 	invalidTolerationKeyErrFmt   = "invalid toleration key %+v: %s"
 	invalidTolerationValueErrFmt = "invalid toleration value %+v: %s"
 	uniqueTolerationErrFmt       = "toleration %+v already exists, tolerations must be unique"
+
+	// resourcePropertyNamePrefix is the prefix (also known as the subdomain) of the label name
+	// associated with all resource properties.
+	resourcePropertyNamePrefix = "resources.kubernetes-fleet.io/"
+
+	// Below is the map of supported capacity types.
+	supportedResourceCapacityTypesMap = map[string]bool{"total": true, "allocatable": true, "available": true}
```
ryanzhang-oss, line 45: why not resue the const in pkg/scheduler/framework/plugins/clusteraffinity/types.go?

ryanzhang-oss (commented):
how about remove the webhook flag in this PR too?

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -34,6 +37,14 @@ var (
 	invalidTolerationKeyErrFmt   = "invalid toleration key %+v: %s"
 	invalidTolerationValueErrFmt = "invalid toleration value %+v: %s"
 	uniqueTolerationErrFmt       = "toleration %+v already exists, tolerations must be unique"
+
+	// resourcePropertyNamePrefix is the prefix (also known as the subdomain) of the label name
```
ryanzhang-oss, line 41: why not using common one?

ryanzhang-oss (approved):

PR #813, opened 2024-05-10
fix: check the deployment  available replica count instead

ryanzhang-oss:

The "Available" condition of a deployment is not sufficient to indicate that a deployment is updated correctly so we need
to tighten up the standard.

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment

Fixes #

I have:

### DIFF
--- a/pkg/controllers/work/apply_controller.go
+++ b/pkg/controllers/work/apply_controller.go
@@ -465,16 +465,13 @@ func trackDeploymentAvailability(curObj *unstructured.Unstructured) (ApplyAction
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(curObj.Object, &deployment); err != nil {
 		return errorApplyAction, controller.NewUnexpectedBehaviorError(err)
 	}
-	// see if the DeploymentAvailable condition is true
-	var depCond *appv1.DeploymentCondition
-	for i := range deployment.Status.Conditions {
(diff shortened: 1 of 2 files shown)


pkg/controllers/work/apply_controller.go
```diff
@@ -465,16 +465,13 @@ func trackDeploymentAvailability(curObj *unstructured.Unstructured) (ApplyAction
...
 		return errorApplyAction, controller.NewUnexpectedBehaviorError(err)
 	}
-	// see if the DeploymentAvailable condition is true
-	var depCond *appv1.DeploymentCondition
-	for i := range deployment.Status.Conditions {
-		if deployment.Status.Conditions[i].Type == appv1.DeploymentAvailable {
-			depCond = &deployment.Status.Conditions[i]
-			break
-		}
+	requiredReplicas := int32(1)
```
zhiying-lin, line 468: here we're tracking the workload availability, not complete. Should we use the available condition instead?

Actually, it looks like our deamonSet and statefulset availability check is too tight. Should we respect the MaxUnavailable?

zhiying-lin: Discussed offline, Need to update the comment in a separate PR to reflect this check.

zhiying-lin (approved):

PR #819, opened 2024-05-14
docs:  rollout concept doc

ryanzhang-oss:

We need to spell out the rollout strategy more clearly

Fixes #

I have:

### DIFF
--- a/docs/concepts/SafeRollout/README.md
+++ b/docs/concepts/SafeRollout/README.md
@@ -64,7 +64,8 @@ of the following resources to determine if they are available or not. Here are t
 We only mark a `Deployment` as available when all its pods are running, ready and updated according to the latest spec. 
 
 #### DaemonSet 
-We only mark a `DaemonSet` as available when all its pods are running, ready and updated according to the latest spec.
+We only mark a `DaemonSet` as available when all its pods are available and updated according to the latest spec on all 
+desired scheduled nodes.
 
 #### StatefulSet
 We only mark a `StatefulSet` as available when all its pods are running, ready and updated according to the latest revision.
--- a/docs/concepts/SafeRollout/README.md
+++ b/docs/concepts/SafeRollout/README.md
@@ -59,7 +59,7 @@ spec:
   resourceSelectors:
     - group: ""
       kind: Namespace
-      version: v1          
+      version: v1
       name: test-ns
   policy:
     placementType: PickN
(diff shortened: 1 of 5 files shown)


docs/concepts/SafeRollout/README.md
```diff
@@ -0,0 +1,214 @@
...
+      maxSurge: 2
+```
+
+The rollout will be as follows:
+
+- We try to pick clusters (cluster-3 and cluster-4) by specifying the label selector `loc: east`.
+- But this time around since we have `maxSurge` set to 2 we are saying we can propagate resources to a maximum of 
+4 clusters but our target number of clusters specified is 2, we will rollout the namespace with deployment to both 
+cluster-3 and cluster-4 before removing the deployment from cluster-1 and cluster-2. 
+- And since `maxUnavailable` is always set to 1 by default, we will try to remove the resource from cluster-1 first and
```
ryanzhang-oss, line 166:
```suggestion
- And since `maxUnavailable` is set to 25% by default, we remove the resource from one of the existing clusters

```

docs/concepts/SafeRollout/README.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+The rollout will be as follows:
+
+- We try to pick 3 clusters out of 4, for this scenario let's say we pick cluster-1, cluster-2 & cluster-3.
+- Since we can't track the initial availability for the deployment, we rollout the namespace with deployment to 
+cluster-1, cluster-2 & cluster-3
+- Then we update the deployment with a bad image name to update the resource in place on cluster-1, cluster-2 & cluster-3
+- But since we have `maxUnavailable` set to 1, we will rollout the bad image name update for deployment to one of the clusters 
+(which cluster the resource is rolled out to first is non-deterministic)
```
zhiying-lin, line 87:
```suggestion
(which cluster the resource is rolled out to first is non-deterministic).

```

docs/concepts/SafeRollout/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+#### Job
+We only mark a `Job` as available when it has at least one succeeded pod or one ready pod.
+
+#### Service
+For `Service` based on the service type the availability is determined as follows:
+
+- For `ClusterIP` & `NodePort` service, we mark it as available when a cluster IP is assigned.
+- For `LoadBalancer` service, we mark it as available when a `LoadBalancerIngress` has been assigned along with an IP or Hostname.
+- For `ExternalName` service, we don't know how to determine the availability.
```
zhiying-lin, line 202:
```suggestion
- For `ExternalName` service, checking availability is not supported, so it will be marked as available with not trackable reason.

```

zhiying-lin (approved):

PR #820, opened 2024-05-14
fix: own the resource when the existing resource is not owned by other applier

zhiying-lin:

Allow agent to take over the resource if it has no owner reference.

Fixes #

I have:

UT and e2e

### DIFF
--- a/test/e2e/placement_apply_strategy_test.go
+++ b/test/e2e/placement_apply_strategy_test.go
@@ -8,10 +8,9 @@ package e2e
 import (
 	"fmt"
 
+	"github.com/google/go-cmp/cmp"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
-	"github.com/google/go-cmp/cmp"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


test/e2e/placement_selecting_resources_test.go
```diff
@@ -702,6 +702,14 @@ var _ = Describe("validating CRP when failed to apply resources", Ordered, func(
...
 
 		ns := appNamespace()
+		ns.SetOwnerReferences([]metav1.OwnerReference{
+			{
+				APIVersion: "another-api-version",
+				Kind:       "another-kind",
+				Name:       "another-owner",
+				UID:        "another-uid",
+			},
+		})
```
ryanzhang-oss, line 712: why put this?

test/e2e/placement_apply_strategy_test.go
```diff
@@ -34,12 +33,9 @@ var _ = Describe("validating CRP when resources exists", Ordered, func() {
...
-
-		By("deleting created work resources on member cluster")
-		cleanWorkResourcesOnCluster(allMemberClusters[0])
 	})
 
-	Context("Test a CRP place objects successfully (client-side-apply)", Ordered, func() {
+	Context("Test a CRP place objects successfully (client-side-apply and allow co-own)", Ordered, func() {
 		BeforeAll(func() {
 			ns := appNamespace()
 			ns.SetOwnerReferences([]metav1.OwnerReference{
```
ryanzhang-oss, line 41: is the  owner reference copied to the member cluster?

test/e2e/placement_selecting_resources_test.go
zhiying-lin: This test is to create a resource in the member first and then create CRP. Without the fix, the apply will fail. It validates the CRP status.

For now, only when the ns is owned by other applier, CRP will fail.

test/e2e/placement_apply_strategy_test.go
zhiying-lin: here it creates the ns directly on the member cluster. are you suggesting that we create a ns with ownerreference in the hub to see if the ownerrefernces are copied to the member?

ryanzhang-oss (approved):

PR #821, opened 2024-05-14
test: safe rollout e2e deployment

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/test/e2e/actuals_test.go
+++ b/test/e2e/actuals_test.go
@@ -568,7 +568,7 @@ func safeDeploymentCRPStatusUpdatedActual(wantSelectedResourceIdentifiers []plac
 						Type:               string(placementv1beta1.ResourcesAvailableConditionType),
 						Status:             metav1.ConditionFalse,
 						Reason:             "ManifestNotAvailableYet",
-						ObservedGeneration: crp.Generation + 1,
+						ObservedGeneration: 2,
 					},
 				},
 			},
--- a/test/e2e/actuals_test.go
+++ b/test/e2e/actuals_test.go
@@ -512,6 +512,122 @@ func customizedCRPStatusUpdatedActual(crpName string,
 	}
 }
 
+func safeDeploymentCRPStatusUpdatedActual(wantSelectedResourceIdentifiers []placementv1beta1.ResourceIdentifier, failedDeploymentName, failedDeploymentNamespace, crpName string, wantSelectedClusters []string, wantObservedResourceIndex string) func() error {
+	return func() error {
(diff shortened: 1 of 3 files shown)


test/e2e/actuals_test.go
```diff
@@ -512,6 +512,119 @@ func customizedCRPStatusUpdatedActual(crpName string,
...
+				},
+			},
+			FailedPlacements: []placementv1beta1.FailedResourcePlacement{
+				{
+					ResourceIdentifier: failedDeploymentResourceIdentifier,
+					Condition: metav1.Condition{
+						Type:               string(placementv1beta1.ResourcesAvailableConditionType),
+						Status:             metav1.ConditionFalse,
+						Reason:             "ManifestNotAvailableYet",
+						ObservedGeneration: crp.Generation + 1,
```
ryanzhang-oss, line 571: there is actually no relationship between the crp generation and this failed manifest generation. This is the generation of the object that failed to apply/available

test/e2e/setup_test.go
```diff
@@ -186,6 +186,16 @@ var (
 		ignoreConditionLTTAndMessageFields,
 		cmpopts.EquateEmpty(),
 	}
+
+	safeDeploymentCRPStatusCmpOptions = cmp.Options{
+		cmpopts.SortSlices(lessFuncCondition),
+		cmpopts.SortSlices(lessFuncPlacementStatus),
```
ryanzhang-oss, line 192: AFAIk, this sorts the resourceplacementStatus by clusterName but we are not sure which cluster gets rolled out first (in theory).  Wouldn't this be a random test case?

ryanzhang-oss: but if we don't use this function, how can they compare then?

test/e2e/setup_test.go
```diff
@@ -186,6 +186,15 @@ var (
 		ignoreConditionLTTAndMessageFields,
 		cmpopts.EquateEmpty(),
 	}
+
+	safeDeploymentCRPStatusCmpOptions = cmp.Options{
+		cmpopts.SortSlices(lessFuncCondition),
+		cmpopts.SortSlices(lessFuncResourceIdentifier),
+		cmpopts.SortSlices(lessFuncFailedResourcePlacements),
```
ryanzhang-oss, line 193: I wonder how the compare works when we don't know which cluster gets the bad deplolyment in this case?

Arvindthiru: Since we don't sort placementStatus using clusterName, we can instead sort using conditions length since there should be only one placementStatus with available condition, other placementstatus should have rollout stopped condition

PR #825, opened 2024-05-20
fix: fix wordy message

britaniar:

Fixes #

I have:
Quote the cluster names.

Fix bug bash bug:
Change 'britaniar-aks-member-3 is not selected: specified cluster is not found'
to "Specified cluster "xxx" is not found"


pkg/scheduler/framework/framework.go
```diff
@@ -41,7 +41,7 @@ const (
 	// The reasons to use for scheduling decisions.
 	pickedByPolicyReason                  = "picked by scheduling policy"
 	pickFixedInvalidClusterReasonTemplate = "cluster is not eligible for resource placement yet: %s"
-	pickFixedNotFoundClusterReason        = "specified cluster is not found"
+	pickFixedNotFoundClusterReason        = "specified cluster \"%s\" is not found"
```
ryanzhang-oss, line 44: reason is a machine read code, it should not be templated.

pkg/controllers/clusterresourceplacement/placement_status.go
```diff
@@ -142,6 +142,9 @@ func (r *Reconciler) setResourceConditions(ctx context.Context, crp *fleetv1beta
 		if unselected[i].ClusterScore != nil {
 			scheduledCondition.Message = fmt.Sprintf(resourcePlacementConditionScheduleFailedWithScoreMessageFormat, unselected[i].ClusterName, unselected[i].ClusterScore, unselected[i].Reason)
 		}
+		if unselected[i].Reason == fmt.Sprintf("specified cluster \"%s\" is not found", unselected[i].ClusterName) {
```
ryanzhang-oss, line 145: just copy the reason should be enough

pkg/scheduler/framework/framework.go
zhiying-lin: the reason here is not the one in the condition, more like a message. I think using template should be fine.

pkg/controllers/clusterresourceplacement/placement_status.go
zhiying-lin: I would suggest to copy reason directly from the cluster decision. It means we will build the whole message in the scheduler instead.

ryanzhang-oss (approved):

pkg/scheduler/framework/framework.go
```diff
@@ -40,9 +40,9 @@ const (
 
 	// The reasons to use for scheduling decisions.
 	pickedByPolicyReason                  = "picked by scheduling policy"
-	pickFixedInvalidClusterReasonTemplate = "cluster is not eligible for resource placement yet: %s"
-	pickFixedNotFoundClusterReason        = "specified cluster is not found"
-	notPickedByScoreReason                = "cluster does not score high enough"
+	pickFixedInvalidClusterReasonTemplate = "cluster \"%s\" is not eligible for resource placement yet: %s"
```
zhiying-lin, line 43: Message is usually using upper case. Can you fix that?

eg deployment condition,

status:
  conditions:
  - lastTransitionTime: "2024-05-21T08:31:21Z"
    lastUpdateTime: "2024-05-21T08:31:24Z"
    message: ReplicaSet "member-agent-76c4999765" has successfully progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
  - lastTransitionTime: "2024-05-24T14:53:37Z"
    lastUpdateTime: "2024-05-24T14:53:37Z"
    message: Deployment does not have minimum availability.
    reason: MinimumReplicasUnavailable
    status: "False"
    type: Available

pkg/scheduler/framework/framework.go
```diff
@@ -40,9 +40,9 @@ const (
...
+	pickFixedNotFoundClusterReason        = "specified cluster \"%s\" is not found"
```
zhiying-lin, line 44: should be "pickFixedNotFoundClusterReasonTemplate"

pkg/scheduler/framework/framework.go
britaniar: Addressed on #833

PR #828, opened 2024-05-22
fix: eventually create namespace in flaky E2E test

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/test/e2e/placement_selecting_resources_test.go
+++ b/test/e2e/placement_selecting_resources_test.go
@@ -711,7 +711,9 @@ var _ = Describe("validating CRP when failed to apply resources", Ordered, func(
 			},
 		})
 		By(fmt.Sprintf("creating namespace %s on member cluster", ns.Name))
-		Expect(allMemberClusters[0].KubeClient.Create(ctx, &ns)).Should(Succeed(), "Failed to create namespace %s", ns.Name)
+		Eventually(func() error {
+			return allMemberClusters[0].KubeClient.Create(ctx, &ns)
+		}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to create namespace %s", ns.Name)
 
 		// Create the CRP.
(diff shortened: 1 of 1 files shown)


test/e2e/placement_selecting_resources_test.go
```diff
@@ -711,7 +711,9 @@ var _ = Describe("validating CRP when failed to apply resources", Ordered, func(
 			},
 		})
 		By(fmt.Sprintf("creating namespace %s on member cluster", ns.Name))
-		Expect(allMemberClusters[0].KubeClient.Create(ctx, &ns)).Should(Succeed(), "Failed to create namespace %s", ns.Name)
+		Eventually(func() error {
```
zhiying-lin, line 714: you got conflict error? if so, we need to make sure the ns is deleted in other test cases before continue, other than retry.

Arvindthiru: Failed E2E https://github.com/Azure/fleet/actions/runs/9183806071/job/25255110062?pr=819

Looks like the namespace is being deleted 
image

zhiying-lin: yeah, that's the case i'm talking about. Other tests are deleting ns but we did not wait to make sure the ns is deleted. And here we create the ns again and then hit 409 issue.

We need to scan the code to find out which tests are not waiting for the ns to be deleted.

ryanzhang-oss: it seems that we see this particular e2e suite hits this ns deleting issue a lot, is there any reason this only affects this suite?

zhiying-lin: This one actually makes sure the resources are deleted from all the clusters (line 742 & line 820).

It cannot use ensureCRPAndRelatedResourcesDeletion func cause the validation is very special compared with others (validate whether the existing resource should not be deleted by CRP).

PR #832, opened 2024-05-24
fix: downgrade go mod

ryanzhang-oss:

downgrade the go mod so others can import it

Fixes #

I have:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module go.goms.io/fleet
 
-go 1.22.3
+go 1.22.2
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.9.0
@@ -14,6 +14,7 @@ require (
 	github.com/onsi/gomega v1.29.0
 	github.com/openkruise/kruise v1.4.0
 	github.com/prometheus/client_golang v1.17.0
+	github.com/prometheus/client_model v0.5.0
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.4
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/onsi/gomega v1.29.0
 	github.com/openkruise/kruise v1.4.0
 	github.com/prometheus/client_golang v1.17.0
+	github.com/prometheus/client_model v0.5.0
```
serbrech, line 17: is that voluntary?

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/onsi/gomega v1.29.0
```
serbrech, line 14: unrelated: headsup that gomega has associated CVEs on their library (been bumping it in all toolkit packages in aks repo).
I would suggest to bumpt to latest

go.mod
ryanzhang-oss: yes

go.mod
```diff
@@ -1,6 +1,6 @@
 module go.goms.io/fleet
 
-go 1.22.3
+go 1.22.2
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.9.0
```
serbrech, line 6: also bump the sdk libraries. they have many improvements to the handling of responses, pollers, auth, etc...

go.mod
ryanzhang-oss: I don't think we use gomega in the feature code

serbrech (dismissed):

go.mod
serbrech: Still shows up in the repo scans, because transitive dependencies show up in the go.sum
we have to fix them all in AKS internal repos. But up to you if you don't want to follow that guidance.

britaniar:
Namespace created on member cluster [here](https://github.com/Azure/fleet/blob/4b0f6877d4f9e7a6a7fa4c570b06b5a87da08abf/test/e2e/placement_selecting_resources_test.go#L692) needs to also be deleted.  It is not owned by CRP so it is not being deleted.

britaniar (approved):
E2E fix LGTM!

PR #833, opened 2024-05-24
fix: fix message

britaniar:

Fixes #

I have: capitalize message and label as template

Addresses comments on #825

### DIFF
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -39,10 +39,10 @@ const (
 	eventRecorderNameTemplate = "scheduler-framework-%s"
 
 	// The reasons to use for scheduling decisions.
-	pickedByPolicyReason                  = "picked by scheduling policy"
-	pickFixedInvalidClusterReasonTemplate = "cluster \"%s\" is not eligible for resource placement yet: %s"
-	pickFixedNotFoundClusterReason        = "specified cluster \"%s\" is not found"
-	notPickedByScoreReason                = "cluster \"%s\" does not score high enough (affinity score: %d, topology spread score: %d)"
+	pickedByPolicyReason                   = "picked by scheduling policy"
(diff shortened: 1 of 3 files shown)


pkg/scheduler/framework/framework.go
```diff
@@ -39,10 +39,10 @@ const (
 	eventRecorderNameTemplate = "scheduler-framework-%s"
 
 	// The reasons to use for scheduling decisions.
-	pickedByPolicyReason                  = "picked by scheduling policy"
-	pickFixedInvalidClusterReasonTemplate = "cluster \"%s\" is not eligible for resource placement yet: %s"
-	pickFixedNotFoundClusterReason        = "specified cluster \"%s\" is not found"
-	notPickedByScoreReason                = "cluster \"%s\" does not score high enough (affinity score: %d, topology spread score: %d)"
+	pickedByPolicyReason                   = "picked by scheduling policy"
```
ryanzhang-oss, line 42: why is this message not start with capital letter?

britaniar: It is used at the of [resourceScheduleSucceededMessageFormat ](https://github.com/Azure/fleet/blob/754df442ec85dbc599c22c62a1201960e02836f7/pkg/scheduler/framework/frameworkutils.go#L26C2-L26C114) and [resourceScheduleSucceededWithScoreMessageFormat](https://github.com/Azure/fleet/blob/754df442ec85dbc599c22c62a1201960e02836f7/pkg/scheduler/framework/frameworkutils.go#L27). I can delete and just add the statement to the end of these.

ryanzhang-oss: next PR then. We should group all the reason in one file, it's scattered now.

ryanzhang-oss (approved):

pkg/scheduler/framework/framework.go
britaniar: Addressed on #845

PR #834, opened 2024-05-27
test: add more UTs to the IMC controller

michaelawyu:
### Description of your changes

This PR adds more UTs to the IMC controller.

I have:

- [x] Run `make reviewable` to ensure this PR is ready for review.

### How has this code been tested

N/A

PR #835, opened 2024-05-27
test: refactor the IMC controller integration tests

michaelawyu:

This PR refactors the IMC controller integration tests to address some issues in the prev. integration and cover more functionalities (e.g., properties).

I have:

N/A

Might overlap with #834.


pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
```diff
@@ -2,220 +2,834 @@
...
-			Expect(k8sClient.Create(ctx, &internalMemberCluster)).Should(Succeed())
-		})
+			Expect(hubClient.Create(ctx, imc)).Should(Succeed())
 
-		It("should update internalMemberCluster to joined", func() {
-			result, err := r.Reconcile(ctx, ctrl.Request{
-				NamespacedName: memberClusterNamespacedName,
+			// Report properties via the property provider.
+			observationTime := metav1.Now()
+			propertyProvider1.Update(&propertyprovider.PropertyCollectionResponse{
```
ryanzhang-oss, line 70: IIRC, we setup the response here? If so, it's more like a UT then?

pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
```diff
@@ -2,220 +2,834 @@
...
+						},
+						Allocatable: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("12"),
+							corev1.ResourceMemory: resource.MustParse("24Gi"),
+						},
+						Available: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("3"),
+							corev1.ResourceMemory: resource.MustParse("6Gi"),
+						},
+					},
```
ryanzhang-oss, line 441: how should cx use those when the agent is not joined?

pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
```diff
@@ -2,220 +2,834 @@
...
+					ResourceUsage: clusterv1beta1.ResourceUsage{
+						Capacity: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("15"),
+							corev1.ResourceMemory: resource.MustParse("49Gi"),
+						},
+						Allocatable: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("13"),
+							corev1.ResourceMemory: resource.MustParse("43Gi"),
+						},
```
ryanzhang-oss, line 786: are those value tied to how we setup the integration test?

pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
michaelawyu: Hi Ryan! This case is a bit different; it uses a provider and the data reported by the provider is manually refreshed. For the IMC controller integration test it might not matter much to actually implement a provider that watches for nodes/pods, as the behavior to verify only concerns how IMC controller processes data as reported by the provider.

pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
michaelawyu: Hi Ryan! At this moment the behavior is to leave the last observed data there; would you like me to update it so that the IMC controller would clear the information out when a cluster leaves?

pkg/controllers/internalmembercluster/v1beta1/member_controller_integration_test.go
michaelawyu: Yeah, I will link them to common variables for better simplicity.

ryanzhang-oss (approved):

PR #840, opened 2024-05-29
feat: remove webhook flag

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/cmd/hubagent/options/validation.go
+++ b/cmd/hubagent/options/validation.go
@@ -38,7 +38,7 @@ func (o *Options) Validate() field.ErrorList {
 	}
 
 	if o.WebhookServiceName == "" {
-		errs = append(errs, field.Invalid(newPath.Child("WebhookServiceName"), o.WebhookServiceName, "Webhook service name is required when webhook is enabled"))
+		errs = append(errs, field.Invalid(newPath.Child("WebhookServiceName"), o.WebhookServiceName, "Webhook service name is required"))
 	}
 
 	connectionType := o.WebhookClientConnectionType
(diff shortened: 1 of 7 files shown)


Arvindthiru:
Removed validation for required webhook service name. We have a default webhook svc name **fleetwebhook** in options

test/e2e/setup.sh
```diff
@@ -45,7 +45,7 @@ AKS_NODE_REGIONS=("westus" "northeurope" "eastasia")
 AKS_NODE_SKUS=("Standard_A4_v2" "Standard_B4ms" "Standard_D8s_v5" "Standard_E16_v5" "Standard_M16ms")
 AKS_SKU_COUNT=${#AKS_NODE_SKUS[@]}
 # The number of clusters that has pre-defined configuration for testing purposes. 
-RESERVED_CLUSTER_COUNT=3
+RESERVED_CLUSTER_COUNT=${MEMBER_CLUSTER_COUNT}
```
Arvindthiru, line 48: Required change for local testing with more/less than 3 member clusters

cmd/hubagent/options/validation.go
```diff
@@ -37,10 +37,6 @@ func (o *Options) Validate() field.ErrorList {
 		errs = append(errs, field.Invalid(newPath.Child("WorkPendingGracePeriod"), o.WorkPendingGracePeriod, "Must be greater than 0"))
 	}
 
-	if o.EnableWebhook && o.WebhookServiceName == "" {
-		errs = append(errs, field.Invalid(newPath.Child("WebhookServiceName"), o.WebhookServiceName, "Webhook service name is required when webhook is enabled"))
-	}
-
 	connectionType := o.WebhookClientConnectionType
```
zhiying-lin, line 40: you cannot remove this check. eg, user sets the webhookServiceName explicitly. In this case, we need to return the error.

PR #841, opened 2024-05-31
fix: flaky UT for property name validation

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -9,7 +9,7 @@ package validator
 import (
 	"errors"
 	"fmt"
-	"reflect"
+	"sort"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -39,8 +39,8 @@ var (
 	uniqueTolerationErrFmt       = "toleration %+v already exists, tolerations must be unique"
 
 	// Below is the map of supported capacity types.
-	supportedResourceCapacityTypesMap = map[string]bool{propertyprovider.TotalCapacityName: true, propertyprovider.AllocatableCapacityName: true, propertyprovider.AvailableCapacityName: true}
-	supportedResourceCapacityTypes    = reflect.ValueOf(supportedResourceCapacityTypesMap).MapKeys()
+	supportedResourceCapacityTypesMap = map[string]bool{propertyprovider.AllocatableCapacityName: true, propertyprovider.AvailableCapacityName: true, propertyprovider.TotalCapacityName: true}
+	resourceCapacityTypes             = supportedResourceCapacityTypes()
```
ryanzhang-oss, line 43: why creating this alias?

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -452,7 +452,7 @@ func validateName(name string) error {
 			return fmt.Errorf("invalid resource property name %s, expected format is [PREFIX]/[CAPACITY_TYPE]-[RESOURCE_NAME]", name)
 		}
 		if !supportedResourceCapacityTypesMap[segments[0]] {
-			return fmt.Errorf("invalid capacity type in resource property name %s, supported values are %+v", name, supportedResourceCapacityTypes)
+			return fmt.Errorf("invalid capacity type in resource property name %s, supported values are %+v", name, resourceCapacityTypes)
```
ryanzhang-oss, line 455: would  supportedResourceCapacityTypes() work here?

Arvindthiru: Map was suggested to make the lookup faster in this comment https://github.com/Azure/fleet/pull/808#discussion_r1605199781

pkg/utils/validator/clusterresourceplacement.go
Arvindthiru: This var gets created once for this file. Since the validator function could be called many time I wanted to avoid creating the slice from the map every time the function is called.

ryanzhang-oss (approved):

PR #850, opened 2024-06-12
fix: fine tune the Azure setup

ryanzhang-oss:
Some minor fix for the Azure setup

Fixes #

I have:


hack/Azure/setup/joinMC.sh
```diff
@@ -36,7 +36,7 @@ metadata:
 type: kubernetes.io/service-account-token
 EOF
 
-echo "Creating member cluster..."
+echo "Creating member cluster CR..."
```
britaniar, line 39: what is the 'CR'?

ryanzhang-oss: the "custom resource", just to clairfy that we are not creating the cluster

britaniar (approved):

PR #852, opened 2024-06-17
feat: copy failed manifest from binding

zhiying-lin:

Build the crp status by copying failed manifest from binding

Fixes #

I have:


ryanzhang-oss (commented):
I think we need at least one integration/e2e test that talks to the real API server

ryanzhang-oss (approved):

PR #853, opened 2024-06-17
fix: address a minor issue that might cause duplicate entries in cluster decisions (status)

michaelawyu:

This PR fixes a minor issue where duplicate entries might occur in cluster decisions when a cluster is picked and then unpicked without having a scheduling policy change.

I have:

- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/pkg/scheduler/framework/frameworkutils.go
+++ b/pkg/scheduler/framework/frameworkutils.go
@@ -283,7 +283,7 @@ func newSchedulingDecisionsFromBindings(
 	//
 	// Note that for such occurrences, we will honor the previously made decisions in an attempt
 	// to reduce fluctations.
-	seenClusters := map[string]bool{}
+	seenClusters := make(map[string]bool)
 
 	// Build new scheduling decisions.
 	slotsLeft := clustersDecisionArrayLengthLimitInAPI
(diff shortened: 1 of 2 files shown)


pkg/scheduler/framework/frameworkutils.go
```diff
@@ -296,6 +308,11 @@ func newSchedulingDecisionsFromBindings(
 			break
 		}
 
+		if _, ok := seenClusters[sc.Cluster.Name]; ok {
```
ryanzhang-oss, line 311:
```suggestion
		if seenClusters[sc.Cluster.Name] {
```

pkg/scheduler/framework/frameworkutils.go
```diff
@@ -274,12 +274,24 @@ func newSchedulingDecisionsFromBindings(
...
+	// Track clusters that have been added to the decision list.
+	//
+	// This is to avoid duplicate entries in the decision list, caused by a rare occurrence
+	// where a cluster is picked before, but later loses its status as a matching (or most
+	// preferable) cluster (e.g., a label/cluster property change) under the evaluation
+	// of **the same** scheduling policy.
+	//
+	// Note that for such occurrences, we will honor the previously made decisions in an attempt
+	// to reduce fluctations.
+	seenClusters := map[string]bool{}
```
ryanzhang-oss, line 286: why not use make(map[string]bool)?

pkg/scheduler/framework/frameworkutils.go
michaelawyu: Done 👌

pkg/scheduler/framework/frameworkutils.go
michaelawyu: Done 👌

ryanzhang-oss (approved):

PR #858, opened 2024-06-21
fix: fix ut flakiness

zhiying-lin:

Fix the unit test failure
https://github.com/Azure/fleet/actions/runs/9601760461/attempts/1

Fixes #

I have:

### DIFF
--- a/pkg/controllers/workgenerator/controller_test.go
+++ b/pkg/controllers/workgenerator/controller_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 
 	fleetv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	"go.goms.io/fleet/pkg/controllers/work"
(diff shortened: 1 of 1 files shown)


pkg/controllers/workgenerator/controller_test.go
```diff
@@ -677,7 +678,6 @@ func TestSetBindingStatus(t *testing.T) {
 					},
 				},
 			},
-			maxFailedResourcePlacementLimit: 2,
```
ryanzhang-oss, line 680: looks like we don't test when it's 2 anymore?

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -971,10 +970,26 @@ func TestSetBindingStatus(t *testing.T) {
...
+			if tt.maxFailedResourcePlacementLimit != nil {
+				maxFailedResourcePlacementLimit = *tt.maxFailedResourcePlacementLimit
+			}
 			binding := &fleetv1beta1.ClusterResourceBinding{}
 			setBindingStatus(tt.works, binding)
 			got := binding.Status.FailedPlacements
+			// setBindingStatus is using map to populate the failedResourcePlacement.
+			// There is no default order in traversing the map.
+			// When the result of failedResourcePlacement exceeds the limit, the result will be truncated and cannot be
+			// guaranteed.
```
ryanzhang-oss, line 982: I was aware of that but thought it was not super random. Now that we don't test the limit anymore, can we just remove the limit from the test?

pkg/controllers/workgenerator/controller_test.go
zhiying-lin: The default is 100. if the failed number < limit, 2 or 100 does not matter.

pkg/controllers/workgenerator/controller_test.go
zhiying-lin: we're testing limit in line 821

ryanzhang-oss (approved):

PR #859, opened 2024-06-21
fix: run make manifests, generate

Arvindthiru:

Fixes #

I have:

Ran the following commands in order,
-  make manifests
- make generate
- make reviewable


config/crd/bases/cluster.kubernetes-fleet.io_memberclusters.yaml
```diff
@@ -27,7 +27,7 @@ spec:
       name: Age
       type: date
     - jsonPath: .status.properties.kubernetes-fleet\.io/node-count.value
-      name: Node-Count
+      name: NodeCount
```
ryanzhang-oss, line 30: is this not a breaking change?

zhiying-lin: The "Node-Count" is better here as usually k8 is using "-" to split the words.

Could you please fix the API itself instead of the CRD?

zhiying-lin (approved):

Arvindthiru:
The changes x-kubernetes-list-type: atomic is added for object labelSelector https://github.com/kubernetes/apimachinery/blob/eb26334eeb0f769be8f0c5665ff34713cfdec83e/pkg/apis/meta/v1/types.go#L1242 which has marker //+listType=atomic described here https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy. This doesn't seem like a breaking change for our controllers

PR #860, opened 2024-06-21
feat: remove job availability tracking

Arvindthiru:

Fixes #

I have:


ryanzhang-oss (approved):
so job will always be not trackable. Can wa add a follow up PR to test rollout job without delay as it doesn't make sense to rollout a job anymore.

PR #865, opened 2024-06-25
fix:  enhance the member cluster leave flow

ryanzhang-oss:

1.  fix the immediate bug by just listing all the work in the cluster namespace.
2. issue the work finalizer removal in parallel.
3. issue a foreground delete namespace before removing the finalizer on the MC object.
4. handle the edge case that the imc does not exist.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
+++ b/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
@@ -84,7 +84,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	// Handle deleting/leaving member cluster, garbage collect all the resources in the cluster namespace
 	if !mc.DeletionTimestamp.IsZero() {
 		klog.V(2).InfoS("the member cluster is leaving", "memberCluster", mcObjRef)
-		return r.handleDelete(ctx, mc.DeepCopy())
+		return runtime.Result{}, r.handleDelete(ctx, mc.DeepCopy())
 	}
 
 	// Add the finalizer to the member cluster
@@ -117,21 +117,34 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 
 // handleDelete handles the delete event of the member cluster, makes sure the agent has finished leaving the fleet first and
 // then garbage collects all the resources in the cluster namespace.
-func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.MemberCluster) (runtime.Result, error) {
+func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.MemberCluster) error {
 	mcObjRef := klog.KObj(mc)
 	if !controllerutil.ContainsFinalizer(mc, placementv1beta1.MemberClusterFinalizer) {
 		klog.V(2).InfoS("No need to do anything for the deleting member cluster without a finalizer", "memberCluster", mcObjRef)
-		return runtime.Result{}, nil
+		return nil
 	}
-	currentImc, err := r.getInternalMemberCluster(ctx, mc.GetName())
-	if err != nil {
-		return runtime.Result{}, err
+	// check if the namespace still exist
+	var currentNS corev1.Namespace
+	namespaceName := fmt.Sprintf(utils.NamespaceNameFormat, mc.Name)
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: namespaceName}, &currentNS); err != nil {
+		if !apierrors.IsNotFound(err) {
+			klog.ErrorS(err, "failed to get the member cluster namespace", "memberCluster", mcObjRef)
+			return controller.NewAPIServerError(true, err)
+		}
+		klog.V(2).InfoS("the member cluster namespace is not found, remove the finalizer", "memberCluster", mcObjRef)
+		controllerutil.RemoveFinalizer(mc, placementv1beta1.MemberClusterFinalizer)
+		return controller.NewUpdateIgnoreConflictError(r.Update(ctx, mc, &client.UpdateOptions{}))
 	}
-	// we don't need to wait for the agent to leave if the internal member cluster is not found
-	if currentImc == nil {
-		controller.NewUnexpectedBehaviorError(fmt.Errorf("internal member cluster %s not found", mcObjRef))
-		klog.V(2).InfoS("InternalMemberCluster not found, start garbage collecting", "memberCluster", mcObjRef)
-		return runtime.Result{}, r.garbageCollect(ctx, mc)
+	currentImc := &clusterv1beta1.InternalMemberCluster{}
+	imcNamespacedName := types.NamespacedName{Namespace: namespaceName, Name: mc.Name}
+	if err := r.Client.Get(ctx, imcNamespacedName, currentImc); err != nil {
+		if !apierrors.IsNotFound(err) {
+			klog.ErrorS(err, "failed to get internal member cluster", "internalMemberCluster", imcNamespacedName)
+			return controller.NewAPIServerError(true, err)
+		}
+		// we don't need to wait for the agent to leave if the internal member cluster is not found
+		klog.ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("internal member cluster %s not found", mcObjRef)), "InternalMemberCluster not found, start garbage collecting", "memberCluster", mcObjRef)
+		return r.garbageCollect(ctx, mc)
 	}
 	// calculate the current status of the member cluster from imc status
 	r.syncInternalMemberClusterStatus(currentImc, mc)
@@ -141,17 +154,16 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
 	//       beyond a pre-agreed threshold (which should be in the order of hours instead of minutes) and proceed with garbage collection
 	if condition.IsConditionStatusFalse(cond, mc.GetGeneration()) {
 		klog.V(2).InfoS("Agent already left, start garbage collecting", "memberCluster", mcObjRef)
-		return runtime.Result{}, r.garbageCollect(ctx, mc)
+		return r.garbageCollect(ctx, mc)
 	}
-	klog.V(2).InfoS("Need to wait for agent to leave", "memberCluster", mcObjRef, "agentJoinedCondition", cond)
-	// mark the imc as left again to make sure the agent is leaving the fleet
-	if err = r.leave(ctx, mc, currentImc); err != nil {
+	klog.V(2).InfoS("Need to wait for the agent to leave", "memberCluster", mcObjRef, "agentJoinedCondition", cond)
+	// mark the imc as left to make sure the agent is leaving the fleet
+	if err := r.leave(ctx, mc, currentImc); err != nil {
 		klog.ErrorS(err, "failed to mark the imc as leave", "memberCluster", mcObjRef)
-		return runtime.Result{}, err
+		return err
 	}
 	// update the mc status to track the leaving status while we wait for all the agents to leave
-	err = r.updateMemberClusterStatus(ctx, mc)
-	return runtime.Result{}, controller.NewUpdateIgnoreConflictError(err)
+	return controller.NewUpdateIgnoreConflictError(r.updateMemberClusterStatus(ctx, mc))
 }
 
 func (r *Reconciler) getInternalMemberCluster(ctx context.Context, name string) (*clusterv1beta1.InternalMemberCluster, error) {
@@ -197,7 +209,7 @@ func (r *Reconciler) garbageCollectWork(ctx context.Context, mc *clusterv1beta1.
 	}
 	klog.V(2).InfoS("try to remove all the work finalizers in the cluster namespace",
 		"memberCluster", klog.KObj(mc), "number of work", len(works.Items))
-	return errs.Wait()
+	return controller.NewAPIServerError(false, errs.Wait())
 }
 
 // garbageCollect is used to garbage collect all the resources in the cluster namespace associated with the member cluster.
@@ -212,15 +224,11 @@ func (r *Reconciler) garbageCollect(ctx context.Context, mc *clusterv1beta1.Memb
 	if err := r.garbageCollectWork(ctx, mc, namespaceName); err != nil {
 		return err
 	}
-	// delete the namespace in the foreground so that we are sure it is deleted before we mark the member cluster as left
-	deletePolicy := metav1.DeletePropagationForeground
-	if err := r.Delete(ctx, &clusterNS, &client.DeleteOptions{PropagationPolicy: &deletePolicy}); err != nil {
-		klog.ErrorS(err, "failed to remove the cluster namespace",
-			"memberCluster", klog.KObj(mc), "namespace", namespaceName)
-		return err
+	if err := r.Delete(ctx, &clusterNS); err != nil {
+		klog.ErrorS(err, "failed to remove the cluster namespace", "memberCluster", klog.KObj(mc), "namespace", namespaceName)
 	}
-	controllerutil.RemoveFinalizer(mc, placementv1beta1.MemberClusterFinalizer)
-	return r.Update(ctx, mc, &client.UpdateOptions{})
+	klog.V(2).InfoS("deleted the member cluster namespace", "memberCluster", klog.KObj(mc))
+	return nil
 }
 
 // ensureFinalizer makes sure that the member cluster CR has a finalizer on it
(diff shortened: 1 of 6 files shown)


pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -162,37 +171,56 @@ func (r *Reconciler) getInternalMemberCluster(ctx context.Context, name string)
...
+	var clusterNS corev1.Namespace
+	namespaceName := fmt.Sprintf(utils.NamespaceNameFormat, mc.Name)
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: namespaceName}, &clusterNS); apierrors.IsNotFound(err) {
+		klog.V(2).InfoS("the member cluster namespace is successfully deleted", "memberCluster", klog.KObj(mc))
+		return nil
+	}
+	if err := r.garbageCollectWork(ctx, mc, namespaceName); err != nil {
+		return err
+	}
+	// delete the namespace in the foreground so that we are sure it is deleted before we mark the member cluster as left
```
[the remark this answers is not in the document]
ryanzhang-oss: kss delete won't be blocked because some resources have owner not deleted (owner deleted will trigger gc but not the other way around). Here I remove the finalizer from MC after the NS is deleted so the MC won't be removed until the NS is deleted.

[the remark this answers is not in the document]
zhiying-lin: hmm, ```Here I remove the finalizer from MC after the NS is deleted so the MC won't be removed until the NS is deleted.```

but you did not wait here. line 217 is not a blocking call to make sure the ns is deleted before return.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -162,37 +171,56 @@ func (r *Reconciler) getInternalMemberCluster(ctx context.Context, name string)
...
 		"memberCluster", klog.KObj(mc), "number of work", len(works.Items))
+	return errs.Wait()
+}
+
+// garbageCollect is used to garbage collect all the resources in the cluster namespace associated with the member cluster.
+func (r *Reconciler) garbageCollect(ctx context.Context, mc *clusterv1beta1.MemberCluster) error {
+	// check if the namespace still exist
+	var clusterNS corev1.Namespace
+	namespaceName := fmt.Sprintf(utils.NamespaceNameFormat, mc.Name)
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: namespaceName}, &clusterNS); apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
zhiying-lin: since we've already validated if ns exists, we don't need to GET again?

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -162,37 +183,52 @@ func (r *Reconciler) getInternalMemberCluster(ctx context.Context, name string)
...
+		klog.V(2).InfoS("the member cluster namespace is successfully deleted", "memberCluster", klog.KObj(mc))
+		return nil
+	}
+	if err := r.garbageCollectWork(ctx, mc, namespaceName); err != nil {
+		return err
+	}
+	if err := r.Delete(ctx, &clusterNS); err != nil {
+		klog.ErrorS(err, "failed to remove the cluster namespace", "memberCluster", klog.KObj(mc), "namespace", namespaceName)
+	}
+	klog.V(2).InfoS("deleted the member cluster namespace", "memberCluster", klog.KObj(mc))
```
zhiying-lin, line 230: all the roles, role bindings are owned by mc instead of namespace. If we don't remove the finalizer from the mc, will gc delete the roles and role bindings etc? Need to test it.

Otherwise, the namespace will stuck because of namespace scoped resources.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -115,42 +117,61 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
+	}
+	currentImc := &clusterv1beta1.InternalMemberCluster{}
+	imcNamespacedName := types.NamespacedName{Namespace: namespaceName, Name: mc.Name}
+	if err := r.Client.Get(ctx, imcNamespacedName, currentImc); err != nil {
+		if !apierrors.IsNotFound(err) {
+			klog.ErrorS(err, "failed to get internal member cluster", "internalMemberCluster", imcNamespacedName)
+			return controller.NewAPIServerError(true, err)
+		}
+		// we don't need to wait for the agent to leave if the internal member cluster is not found
+		klog.ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("internal member cluster %s not found", mcObjRef)), "InternalMemberCluster not found, start garbage collecting", "memberCluster", mcObjRef)
```
zhiying-lin, line 146: now it becomes expected error? cause mc will be there until we delete all the namespace scoped resources and imc may be gone.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -489,7 +525,7 @@ func (r *Reconciler) aggregateJoinedCondition(mc *clusterv1beta1.MemberCluster)
 	reportedAgents := make(map[clusterv1beta1.AgentType]bool)
 	for _, agentStatus := range mc.Status.AgentStatus {
 		if !r.agents[agentStatus.Type] {
-			klog.V(2).InfoS("Ignoring unexpected agent type status", "agentStatus", agentStatus)
+			_ = controller.NewUnexpectedBehaviorError(fmt.Errorf("find an unexpected agent type %s for member cluster %s", agentStatus.Type, mc.Name))
```
zhiying-lin, line 528: prefer to add a error log here so that it will print out the error location and line number and it's easier for debugging though we'll log the error in the NewUnexpectedBehaviorError func.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -162,37 +183,52 @@ func (r *Reconciler) getInternalMemberCluster(ctx context.Context, name string)
...
+			if updateErr := r.Update(cctx, staleWork, &client.UpdateOptions{}); updateErr != nil {
+				klog.ErrorS(updateErr, "failed to remove the finalizer from the work",
+					"memberCluster", klog.KObj(mc), "work", klog.KObj(staleWork))
+				return updateErr
+			}
+			return nil
+		})
 	}
-	klog.V(2).InfoS("successfully removed all the work finalizers in the cluster namespace",
+	klog.V(2).InfoS("try to remove all the work finalizers in the cluster namespace",
```
zhiying-lin, line 210: need to update the log

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
ryanzhang-oss: yes, namespace controller removes all the resources in the namespace regardless of if they are orphaned.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
ryanzhang-oss: so we have two back to back error pting here? I was wondering if the functions in the controller package can print out the call stack.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
zhiying-lin: nvm, i may look at the wrong line. sorry for the confusion.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
zhiying-lin: looks like this func https://pkg.go.dev/runtime/debug#Stack could help to print the stack trace. But i find that k8 does not use this func a lot.

https://github.com/search?q=org%3Akubernetes+++%22debug.Stack%28%29%22+language%3AGo&type=code

not sure if it's ok for production code, eg, will it consume too much memory.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -121,86 +123,136 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
...
+}
+
+// garbageCollect is used to garbage collect all the resources in the cluster namespace associated with the member cluster.
+func (r *Reconciler) garbageCollect(ctx context.Context, mc *clusterv1beta1.MemberCluster) error {
+	// check if the namespace still exist
+	var clusterNS corev1.Namespace
+	namespaceName := fmt.Sprintf(utils.NamespaceNameFormat, mc.Name)
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: namespaceName}, &clusterNS); err != nil {
+		klog.ErrorS(err, "Failed to get the member cluster namespace", "memberCluster", klog.KObj(mc))
+		return err
```
zhiying-lin, line 237:
```suggestion
		return controller.NewAPIServerError(true, err)
```

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
ryanzhang-oss: perf shouldn't be a problem as this is supposed to be very rare.

test/e2e/join_and_leave_test.go
```diff
@@ -0,0 +1,113 @@
...
+			{
+				Kind:      "ConfigMap",
+				Name:      testEnvelopConfigMap.Name,
+				Version:   "v1",
+				Namespace: workNamespaceName,
+			},
+		}
+	})
+
+	Context("Test cluster join and leave flow with CRP not deleted", Ordered, Serial, func() {
```
zhiying-lin, line 48:
```suggestion
	Context("Test cluster join and leave flow with CRP not deleted", func() {
```
nit, ordered and serial is redundant here as they're in parent already.

test/e2e/join_and_leave_test.go
```diff
@@ -0,0 +1,113 @@
...
+			// resourceQuota is enveloped so it's not trackable yet
+			crpStatusUpdatedActual := customizedCRPStatusUpdatedActual(crpName, wantSelectedResources, nil, nil, "0", false)
+			Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
+		})
+
+		It("Should be able to rejoin the cluster", func() {
+			By("rejoin all the clusters without deleting the CRP")
+			setAllMemberClustersToJoin()
+			checkIfAllMemberClustersHaveJoined()
+			checkIfAzurePropertyProviderIsWorking()
```
zhiying-lin, line 102: validate the crp status again? the selected resources should be placed to all members again.

test/e2e/join_and_leave_test.go
```diff
@@ -0,0 +1,113 @@
...
+		})
+	})
+
+	AfterAll(func() {
+		// Remove the custom deletion blocker finalizer from the CRP.
+		cleanupCRP(crpName)
+
+		// Delete the created resources.
+		cleanupWorkResources()
```
zhiying-lin, line 111: need to use ensureCRPAndRelatedResourcesDeletion func otherwise, the resources could be persistent in the members and it may fail other tests.

pkg/controllers/membercluster/v1beta1/membercluster_controller_test.go
```diff
@@ -1540,3 +1542,132 @@ func TestUpdateMemberClusterStatus(t *testing.T) {
...
+		"return success when the mc has no finalizer": {
+			r: &Reconciler{Client: &test.MockClient{},
+				recorder: utils.NewFakeRecorder(1),
+			},
+			memberCluster: &clusterv1beta1.MemberCluster{},
+			wantResult:    ctrl.Result{},
+			wantErr:       nil,
+		},
+		"return success when the namespace does not exit": {
+			r: &Reconciler{Client: &test.MockClient{
```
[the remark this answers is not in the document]
ryanzhang-oss: I wonder why a fake client is easier to maintain? How does it return the error we want to test?

pkg/controllers/membercluster/v1beta1/membercluster_controller_test.go
```diff
@@ -1540,3 +1542,132 @@ func TestUpdateMemberClusterStatus(t *testing.T) {
...
+					}
+					return nil
+				}},
+				recorder: utils.NewFakeRecorder(1),
+			},
+			memberCluster: memberClusterWithFinalizer.DeepCopy(),
+			wantResult:    ctrl.Result{RequeueAfter: time.Second},
+			wantErr:       controller.ErrUnexpectedBehavior,
+		},
+		"requeue with no error when the imc does not exist": {
```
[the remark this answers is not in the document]
ryanzhang-oss: that scenario is covered in the integration test

pkg/controllers/membercluster/v1beta1/membercluster_controller_test.go
```diff
@@ -1540,3 +1542,132 @@ func TestUpdateMemberClusterStatus(t *testing.T) {
...
+			},
+			memberCluster: memberClusterWithFinalizer.DeepCopy(),
+			wantResult:    ctrl.Result{Requeue: true},
+			wantErr:       nil,
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			result, err := tt.r.handleDelete(context.Background(), tt.memberCluster)
```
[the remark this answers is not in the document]
ryanzhang-oss: The unit test mostly just tested the edge cases, none modified the MC status. The MC status is now checked by the integration test.

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -164,14 +164,18 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
...
+	if condition.IsConditionStatusFalse(mcJoinedCondition, mc.GetGeneration()) {
 		klog.V(2).InfoS("Agent already left, start garbage collecting", "memberCluster", mcObjRef)
-		return runtime.Result{Requeue: true}, r.garbageCollect(ctx, mc)
+		if gcErr := r.garbageCollect(ctx, mc); gcErr != nil {
+			return runtime.Result{}, gcErr
+		}
+		return runtime.Result{Requeue: true}, controller.NewUpdateIgnoreConflictError(r.updateMemberClusterStatus(ctx, mc))
 	}
-	klog.V(2).InfoS("Need to wait for the agent to leave", "memberCluster", mcObjRef, "agentJoinedCondition", cond)
+	klog.V(2).InfoS("Need to wait for the agent to leave", "memberCluster", mcObjRef, "agentJoinedCondition", mcJoinedCondition)
```
zhiying-lin, line 178:
```suggestion
	klog.V(2).InfoS("Need to wait for the agent to leave", "memberCluster", mcObjRef, "joinedCondition", mcJoinedCondition)
```

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -154,7 +154,7 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
 			return runtime.Result{}, controller.NewAPIServerError(true, err)
 		}
 		// we don't need to wait for the agent to leave if the internal member cluster is not found
-		if condition.IsConditionStatusTrue(cond, mc.GetGeneration()) {
+		if condition.IsConditionStatusTrue(mcJoinedCondition, mc.GetGeneration()) {
```
zhiying-lin, line 157: now it becomes expected, we garbage collect in line 173 and the update mc call could fail in line 176.

i'm thinking of skipping updating the mc condition to simplify the implementation here?

pkg/controllers/membercluster/v1beta1/membercluster_controller_test.go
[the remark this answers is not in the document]
zhiying-lin: Using the mock needs to maintain the CRUD function manually. That's true for the non 404 error is hard to produce using fake client.

zhiying-lin (approved):

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -121,86 +123,135 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
...
+	currentImc := &clusterv1beta1.InternalMemberCluster{}
+	imcNamespacedName := types.NamespacedName{Namespace: namespaceName, Name: mc.Name}
+	if err := r.Client.Get(ctx, imcNamespacedName, currentImc); err != nil {
+		if !apierrors.IsNotFound(err) {
+			klog.ErrorS(err, "Failed to get internal member cluster", "internalMemberCluster", imcNamespacedName)
+			return runtime.Result{}, controller.NewAPIServerError(true, err)
+		}
+		// this is possible since we garbage collect the internal member cluster first before deleting the MC
+		klog.V(2).Info("InternalMemberCluster not found, start garbage collecting", "memberCluster", mcObjRef)
+		return runtime.Result{Requeue: true}, r.garbageCollect(ctx, mc)
```
zhiying-lin, line 157: we need to update the mc status here in case last time update mc call fails. Otherwise, the mc status is not reliable.

ryanzhang-oss: We can't as the MC status is stale since there is no IMC.

zhiying-lin: Should we treat the IMC not found as MC joined as false in this case?

PR #873, opened 2024-07-08
fix: fix flaky ro webhook e2e test

britaniar:

Fixes #

I have:
- For [webhook failure](https://github.com/Azure/fleet/actions/runs/9805690990/job/27075899440), ensure both are selecting the same resource.
- For [already exists failure](https://github.com/Azure/fleet/actions/runs/9756434642/job/26928197484), there are only two places that use "test-ro" so differentiated the names so that they don't overlap. In one instance it should get denied when trying to create, while in the other it gets created and then deleted.

### DIFF
--- a/test/e2e/webhook_test.go
+++ b/test/e2e/webhook_test.go
@@ -1076,10 +1076,10 @@ var _ = Describe("webhook tests for ResourceOverride CREATE operations resource 
 
 	It("should deny create RO with invalid resource override", func() {
 		Consistently(func(g Gomega) error {
-			By("create 2nd resourceOverride with same resource selection")
+			By("deny create 2nd resourceOverride with same resource selection")
 			ro1 := &placementv1alpha1.ResourceOverride{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      fmt.Sprintf("test-ro-%d", GinkgoParallelProcess()),
+					Name:      fmt.Sprintf("test-create-ro-%d", GinkgoParallelProcess()),
 					Namespace: roNamespace,
 				},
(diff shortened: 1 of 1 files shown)


test/e2e/webhook_test.go
```diff
@@ -1201,7 +1201,7 @@ var _ = Describe("webhook tests for ResourceOverride UPDATE operations", Ordered
 			By("creating a new resource override")
 			ro1 := &placementv1alpha1.ResourceOverride{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      fmt.Sprintf("test-ro-%d", GinkgoParallelProcess()),
+					Name:      ro1Name,
```
zhiying-lin, line 1204: have you figured out which test is creating the RO at the same time? Could you please explain why you believe the PR will fix the issue?

test/e2e/webhook_test.go
```diff
@@ -1076,10 +1076,10 @@ var _ = Describe("webhook tests for ResourceOverride CREATE operations resource
 
 	It("should deny create RO with invalid resource override", func() {
 		Consistently(func(g Gomega) error {
-			By("create 2nd resourceOverride with same resource selection")
+			By("deny create 2nd resourceOverride with same resource selection")
 			ro1 := &placementv1alpha1.ResourceOverride{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      fmt.Sprintf("test-ro-%d", GinkgoParallelProcess()),
+					Name:      fmt.Sprintf("test-create-ro-%d", GinkgoParallelProcess()),
```
zhiying-lin, line 1082: i don't think this's the root cause. this create should always fail and ro cannot be created in this case.

PR #879, opened 2024-07-11
fix: failed placement change need to trigger work generator copy to binding

ryanzhang-oss:

- fix the case that the failed placement in the work has changed but it's not propagated to the binding by the work generator. 
- Also remove the periodically check in the CRP controller as the binding will be updated which means the binding watcher also needs to check the failed placement
- Move some common string to common

I have:

Verified that both the new e2e and integration test failed without the fix.


pkg/controllers/workgenerator/controller.go
```diff
@@ -874,19 +876,30 @@ func (r *Reconciler) SetupWithManager(mgr controllerruntime.Manager) error {
...
+				newAvailableCondition := meta.FindStatusCondition(newWork.Status.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
+
+				// we try to filter out events, we only need to handle the updated event if the applied or available condition flip between true and false
+				// or the failed placements are changed.
+				if condition.EqualCondition(oldAppliedCondition, newAppliedCondition) && condition.EqualCondition(oldAvailableCondition, newAvailableCondition) {
+					if !condition.IsConditionStatusTrue(newAppliedCondition, newWork.Generation) || !condition.IsConditionStatusTrue(newAvailableCondition, newWork.Generation) {
+						// we need to compare the failed placement if the work is not applied or available
+						oldFailedPlacements := extractFailedResourcePlacementsFromWork(oldWork)
+						newFailedPlacements := extractFailedResourcePlacementsFromWork(newWork)
+						if cmp.Equal(oldFailedPlacements, newFailedPlacements, cmp.Options{cmpopts.SortSlices(condition.LessFuncResourceIdentifier),
```
zhiying-lin, line 891: we cannot use the "cmp" package and this package is only used for testing purpose. :(

You can use "equality.Semantic.DeepEqual" instead. If i remember it correctly, it will ignore the slice order.

pkg/utils/condition/condition.go
```diff
@@ -97,6 +99,21 @@ const (
 	AllWorkAvailableReason = "AllWorkAreAvailable"
 )
 
+// IgnoreConditionLTTAndMessageFields is a cmpopts.IgnoreFields that ignores the LastTransitionTime and Message fields
```
zhiying-lin, line 102: we cannot put the test utils into a production package. You can move to test/utils folder.

test/e2e/enveloped_object_placement_test.go
```diff
@@ -36,33 +39,34 @@ var (
 // Note that this container will run in parallel with other containers.
 var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
 	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
-	workNamespaceName := fmt.Sprintf(workNamespaceNameTemplate, GinkgoParallelProcess())
```
zhiying-lin, line 39: need to include the GinkgoParallelProcess(). The test can run in parallel with others. Otherwise, it may conflict with others.

pkg/utils/common.go
```diff
@@ -81,6 +81,9 @@ const (
 
 	// LastWorkUpdateTimeAnnotationKey is used to mark the last update time on a work object.
 	LastWorkUpdateTimeAnnotationKey = "work.fleet.azure.com/last-update-time"
+
+	// ResourceIdentifierStringFormat is the format of the resource identifier string.
+	ResourceIdentifierStringFormat = "%s/%s/%s/%s/%s"
```
zhiying-lin, line 86: can be moved to the test util package? I don't think we'll use it in the production code.

pkg/controllers/workgenerator/controller.go
zhiying-lin: It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that its unsuitable for production environments where a spurious panic may be fatal.

https://pkg.go.dev/github.com/google/go-cmp/cmp

test/e2e/enveloped_object_placement_test.go
Arvindthiru: appNamespace().Name is used which uses GinkgoParallelProcess()

pkg/controllers/workgenerator/controller.go
```diff
@@ -874,23 +875,56 @@ func (r *Reconciler) SetupWithManager(mgr controllerruntime.Manager) error {
...
+	sort.Slice(newFailedResourcePlacements, func(i, j int) bool {
+		return condition.LessFuncFailedResourcePlacements(newFailedResourcePlacements[i], newFailedResourcePlacements[j])
+	})
+	for i := range oldFailedResourcePlacements {
+		oldFailedResourcePlacement := oldFailedResourcePlacements[i]
+		newFailedResourcePlacement := newFailedResourcePlacements[i]
+		if !apiequality.Semantic.DeepEqual(oldFailedResourcePlacement.ResourceIdentifier, newFailedResourcePlacement.ResourceIdentifier) {
+			return false
+		}
+		if !condition.EqualCondition(&oldFailedResourcePlacement.Condition, &newFailedResourcePlacement.Condition) {
```
ryanzhang-oss, line 925: the old observed generation will be less than or equal to the new generation but the equalCondition code is flipped.

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -1484,3 +1484,518 @@ func TestExtractFailedResourcePlacementsFromWork(t *testing.T) {
...
+						Envelope: &fleetv1beta1.EnvelopeIdentifier{
+							Name:      "test-envelope-object",
+							Namespace: "default",
+							Type:      fleetv1beta1.ConfigMapEnvelopeType,
+						},
+					},
+					Condition: metav1.Condition{
+						Type:               fleetv1beta1.WorkConditionTypeApplied,
+						Status:             metav1.ConditionFalse,
+						ObservedGeneration: 0,
```
ryanzhang-oss, line 1622: need to test different generation

test/e2e/resources/statefulset-with-volume.yaml
```diff
@@ -0,0 +1,35 @@
...
+              mountPath: /usr/share/nginx/html
+  volumeClaimTemplates:
+    - metadata:
+        name: www
+      spec:
+        accessModes: [ "ReadWriteOnce" ]
+        storageClassName: "my-storage-class"
+        resources:
+          requests:
+            storage: 1Gi
```
michaelawyu, line 35: Just a nit: we might need to add an empty line here.

pkg/utils/common_test.go
```diff
@@ -0,0 +1,622 @@
...
+				},
+			},
+			new:  []fleetv1beta1.FailedResourcePlacement{},
+			want: false,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			got := IsFailedResourcePlacementsEqual(tc.old, tc.new)
+			if diff := cmp.Diff(tc.want, got); diff != "" {
```
michaelawyu, line 617: Also a nit, but since the diff is just comparing two bool values, it might be better to just use a equality check here

test/e2e/rollout_test.go
```diff
@@ -542,61 +544,11 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
-}
-
-func readEnvelopeConfigMapTestManifest(testEnvelopeObj *corev1.ConfigMap) {
-	By("Read testEnvelopConfigMap resource")
-	err := utils.GetObjectFromManifest("resources/test-envelope-object.yaml", testEnvelopeObj)
-	Expect(err).Should(Succeed())
-}
-
 // createWrappedResourcesForRollout creates an enveloped resource on the hub cluster with a workload object for testing purposes.
 func createWrappedResourcesForRollout(testEnvelopeObj *corev1.ConfigMap, obj metav1.Object, kind string, namespace corev1.Namespace) {
```
michaelawyu, line 548: A nit: after the latest refactoring this function feels somewhat redundant to me. Not a blocking issue at all though; feel free to ignore.

michaelawyu (approved):
LGTM ;) Added some nits.

test/e2e/rollout_test.go
ryanzhang-oss: I tried but found out that the diff makes the code a little bit cleaner.

Arvindthiru (approved):

PR #888, opened 2024-07-18
feat: calculate the time we need to wait for the first applied but not ready binding to be ready.

britaniar:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -133,7 +133,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 
 	// pick the bindings to be updated according to the rollout plan
 	// staleBoundBindings is a list of "Bound" bindings and are not selected in this round because of the rollout strategy.
-	toBeUpdatedBindings, staleBoundBindings, needRoll, err := r.pickBindingsToRoll(ctx, allBindings, latestResourceSnapshot, &crp, matchedCRO, matchedRO)
+	toBeUpdatedBindings, staleBoundBindings, needRoll, waitTime, err := r.pickBindingsToRoll(ctx, allBindings, latestResourceSnapshot, &crp, matchedCRO, matchedRO)
 	if err != nil {
 		klog.ErrorS(err, "Failed to pick the bindings to roll", "clusterResourcePlacement", crpName)
 		return runtime.Result{}, err
@@ -159,10 +159,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	// Update all the bindings in parallel according to the rollout plan.
 	// We need to requeue the request regardless if the binding updates succeed or not
 	// to avoid the case that the rollout process stalling because the time based binding readiness does not trigger any event.
-	// We wait for 1/5 of the UnavailablePeriodSeconds so we can catch the next ready one early.
-	// TODO: only wait the time we need to wait for the first applied but not ready binding to be ready
-	return runtime.Result{RequeueAfter: time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds) * time.Second / 5},
-		r.updateBindings(ctx, toBeUpdatedBindings)
+	// Wait the time we need to wait for the first applied but not ready binding to be ready
+	return runtime.Result{Requeue: true, RequeueAfter: waitTime}, r.updateBindings(ctx, toBeUpdatedBindings)
 }
 
 func (r *Reconciler) checkAndUpdateStaleBindingsStatus(ctx context.Context, bindings []*fleetv1beta1.ClusterResourceBinding) error {
@@ -299,7 +297,7 @@ func createUpdateInfo(binding *fleetv1beta1.ClusterResourceBinding, crp *fleetv1
 // Thus, it also returns a bool indicating whether there are out of sync bindings to be rolled to differentiate those
 // two cases.
 func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*fleetv1beta1.ClusterResourceBinding, latestResourceSnapshot *fleetv1beta1.ClusterResourceSnapshot, crp *fleetv1beta1.ClusterResourcePlacement,
-	matchedCROs []*fleetv1alpha1.ClusterResourceOverrideSnapshot, matchedROs []*fleetv1alpha1.ResourceOverrideSnapshot) ([]toBeUpdatedBinding, []toBeUpdatedBinding, bool, error) {
+	matchedCROs []*fleetv1alpha1.ClusterResourceOverrideSnapshot, matchedROs []*fleetv1alpha1.ResourceOverrideSnapshot) ([]toBeUpdatedBinding, []toBeUpdatedBinding, bool, time.Duration, error) {
 	// Those are the bindings that are chosen by the scheduler to be applied to selected clusters.
 	// They include the bindings that are already applied to the clusters and the bindings that are newly selected by the scheduler.
 	schedulerTargetedBinds := make([]*fleetv1beta1.ClusterResourceBinding, 0)
@@ -333,25 +331,30 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
 	readyTimeCutOff := time.Now().Add(-time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds) * time.Second)
 
 	// classify the bindings into different categories
-	// TODO: calculate the time we need to wait for the first applied but not ready binding to be ready.
+	// Wait for the first applied but not ready binding to be ready.
 	// return wait time longer if the rollout is stuck on failed apply/available bindings
+	minWaitTime := time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds) * time.Second
+	allReady := true
 	crpKObj := klog.KObj(crp)
-	for idx := range allBindings {
-		binding := allBindings[idx]
+	for _, binding := range allBindings {
 		bindingKObj := klog.KObj(binding)
 		switch binding.Spec.State {
 		case fleetv1beta1.BindingStateUnscheduled:
-			appliedCondition := binding.GetCondition(string(fleetv1beta1.ResourceBindingApplied))
-			availableCondition := binding.GetCondition(string(fleetv1beta1.ResourceBindingAvailable))
-			if condition.IsConditionStatusFalse(appliedCondition, binding.Generation) || condition.IsConditionStatusFalse(availableCondition, binding.Generation) {
+			if !isBindingAvailable(binding) {
 				klog.V(3).InfoS("Found a failed to be ready unscheduled binding", "clusterResourcePlacement", crpKObj, "binding", bindingKObj)
 			} else {
 				canBeReadyBindings = append(canBeReadyBindings, binding)
 			}
-			_, bindingReady := isBindingReady(binding, readyTimeCutOff)
+			waitTime, bindingReady := isBindingReady(binding, readyTimeCutOff)
 			if bindingReady {
 				klog.V(3).InfoS("Found a ready unscheduled binding", "clusterResourcePlacement", crpKObj, "binding", bindingKObj)
 				readyBindings = append(readyBindings, binding)
+			} else {
+				allReady = false
+				klog.Info("Binding WaitTime:", waitTime)
+				if waitTime >= 0 && waitTime < minWaitTime {
+					minWaitTime = waitTime
+				}
 			}
 			if binding.DeletionTimestamp.IsZero() {
 				// it's not been deleted yet, so it is a removal candidate
@@ -362,7 +365,6 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
 				// it is being deleted, it can be removed from the cluster at any time, so it can be unavailable at any time
 				canBeUnavailableBindings = append(canBeUnavailableBindings, binding)
 			}
-
 		case fleetv1beta1.BindingStateScheduled:
 			// the scheduler has picked a cluster for this binding
 			schedulerTargetedBinds = append(schedulerTargetedBinds, binding)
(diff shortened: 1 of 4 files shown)


pkg/controllers/rollout/controller_test.go
```diff
@@ -726,6 +726,181 @@ func TestIsBindingReady(t *testing.T) {
...
 }
 
+func TestWaitForFirstAppliedButNotReadyBinding(t *testing.T) {
+	tests := map[string]struct {
+		bindings          []*fleetv1beta1.ClusterResourceBinding
+		readyTimeCutOff   time.Time
+		wantWaitDuration  time.Duration
+		unavailablePeriod time.Duration
+	}{
+		"all bindings ready": {
```
zhiying-lin, line 736: add a test when all of the bindings are not applied

pkg/controllers/rollout/controller.go
```diff
@@ -507,6 +507,29 @@ func (r *Reconciler) calculateRealTarget(crp *fleetv1beta1.ClusterResourcePlacem
 	return targetNumber
 }
 
+// waitForFirstAppliedButNotReadyBinding returns the minimum amount of time that needs to elapse before the first
+// binding that has been applied but not yet marked as ready becomes ready.
+func waitForFirstAppliedButNotReadyBinding(bindings []*fleetv1beta1.ClusterResourceBinding, readyTimeCutOff time.Time, unavailablePeriod time.Duration) time.Duration {
+	minWaitTime := unavailablePeriod
```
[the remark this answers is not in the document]
britaniar: My assumption for "if none of the binding is applied" is that we wait for the `UnavailablePeriodSeconds` based on the crp, which is why it is set as the minimum wait time at first. Or, the option to use the orginal value (1/5 UnavailablePeriodseconds).

pkg/controllers/rollout/controller_test.go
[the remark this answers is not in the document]
britaniar: I thought about that, but when I return the `minWaitTime` as zero, it led to some e2e failures.

ryanzhang-oss (commented):
we should piggy back the min wait time in the pickBindingsToRoll function as it goes through all bindings already

pkg/controllers/rollout/controller.go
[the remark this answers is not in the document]
zhiying-lin: keeping the original value is better. If the resource is built-in, we don't need to wait after UnavailablePeriodseconds. The built-in resource may become ready faster.

pkg/controllers/rollout/controller.go
```diff
@@ -333,82 +331,34 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
 	readyTimeCutOff := time.Now().Add(-time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds) * time.Second)
 
 	// classify the bindings into different categories
-	// TODO: calculate the time we need to wait for the first applied but not ready binding to be ready.
+	// Wait for the first applied but not ready binding to be ready.
 	// return wait time longer if the rollout is stuck on failed apply/available bindings
+	allReady := true
+	minWaitTime := time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds) * time.Second / 5
```
ryanzhang-oss, line 337: the minWaitTime should be 0 regardless of the UnavailablePeriodSeconds

pkg/controllers/rollout/controller.go
```diff
@@ -333,82 +331,34 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
 	readyTimeCutOff := time.Now().Add(-time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds) * time.Second)
 
 	// classify the bindings into different categories
-	// TODO: calculate the time we need to wait for the first applied but not ready binding to be ready.
+	// Wait for the first applied but not ready binding to be ready.
 	// return wait time longer if the rollout is stuck on failed apply/available bindings
+	allReady := true
```
ryanzhang-oss, line 336: is this still needed?

pkg/controllers/rollout/controller.go
```diff
@@ -418,20 +368,146 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
-		"lowerBoundAvailableBindings", lowerBoundAvailableNumber, "maxNumberOfBindingsToRemove", maxNumberToRemove)
+	toBeUpdatedBindingList, staleUnselectedBinding := determineBindingsToUpdate(crp, removeCandidates, updateCandidates, boundingCandidates, applyFailedUpdateCandidates, targetNumber,
+		readyBindings, canBeReadyBindings, canBeUnavailableBindings, crpKObj)
+
+	return toBeUpdatedBindingList, staleUnselectedBinding, true, minWaitTime, nil
+}
+
+// Process unscheduled bindings
+func (r *Reconciler) processUnscheduledBinding(
+	crpKObj, bindingKObj klog.ObjectRef,
```
ryanzhang-oss, line 382: I thought about refactoring the func but the parameters are so huge so I gave up. I guess it's still worth it.

However,  we can get rid of ObjectRef. 

We should also make sure what are the input vs output

pkg/controllers/rollout/controller.go
```diff
@@ -418,20 +368,146 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
+	minWaitTime *time.Duration,
+	readyTimeCutOff time.Time,
+	binding *fleetv1beta1.ClusterResourceBinding,
+	readyBindings, canBeReadyBindings, canBeUnavailableBindings *[]*fleetv1beta1.ClusterResourceBinding,
+	removeCandidates *[]toBeUpdatedBinding,
```
ryanzhang-oss, line 387: these are output. I am not sure if there is any clean way to distinguish them

pkg/controllers/rollout/controller.go
```diff
@@ -362,40 +366,39 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
 			schedulerTargetedBinds = append(schedulerTargetedBinds, binding)
-			// this binding has not been bound yet, so it is an update candidate
-			// pickFromResourceMatchedOverridesForTargetCluster always returns the ordered list of the overrides.
-			cro, ro, err := r.pickFromResourceMatchedOverridesForTargetCluster(ctx, binding, matchedCROs, matchedROs)
+			boundingCandidate, err := r.processScheduledBindings(ctx, binding, crp, latestResourceSnapshot, matchedCROs, matchedROs)
 			if err != nil {
-				return nil, nil, false, err
+				return nil, nil, false, 0, err
+			}
+			if boundingCandidate.currentBinding != nil {
```
ryanzhang-oss, line 376: when will this be true?

ryanzhang-oss, line 376: I don't see much benefit of this refactor, the function is not reused and it didn't shorten the lines of code either.

pkg/controllers/rollout/controller.go
```diff
@@ -408,30 +411,73 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
 		}
 	}
+	minWaitTime := time.Duration(0)
+	if len(waitTimes) > 0 {
+		// Sort the wait times in ascending order
+		sort.Slice(waitTimes, func(i, j int) bool {
+			return waitTimes[i] < waitTimes[j]
+		})
+		minWaitTime = waitTimes[0]
+	}
```
ryanzhang-oss, line 421: find the min in an array is an O(N) time, O(1) space problem, this implementation is O(NlogN) time O(N) space.

pkg/controllers/rollout/controller.go
```diff
@@ -408,30 +411,73 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
 
+// Determine bindings to update
+func determineBindingsToUpdate(
+	crp *fleetv1beta1.ClusterResourcePlacement,
+	removeCandidates, updateCandidates, boundingCandidates, applyFailedUpdateCandidates []toBeUpdatedBinding,
+	targetNumber int,
+	readyBindings, canBeReadyBindings, canBeUnavailableBindings []*fleetv1beta1.ClusterResourceBinding,
+) (
+	[]toBeUpdatedBinding, []toBeUpdatedBinding,
+) {
```
ryanzhang-oss, line 477: this format is unusual

pkg/controllers/rollout/controller.go
```diff
@@ -478,12 +515,36 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
+	// This is the upper bound of the number of bindings that can be ready during the rolling update
+	// We count anything that still has work object on the hub cluster as can be ready since the member agent may have connection issue with the hub cluster
+	upperBoundReadyNumber := len(canBeReadyBindings)
+	maxNumberToAdd := maxReadyNumber - upperBoundReadyNumber
+
+	klog.V(2).InfoS("Calculated the max number of bindings to add", "clusterResourcePlacement", klog.KObj(crp),
+		"maxSurgeNumber", maxSurgeNumber, "maxReadyNumber", maxReadyNumber, "upperBoundReadyBindings",
+		upperBoundReadyNumber, "maxNumberOfBindingsToAdd", maxNumberToAdd)
+	return maxNumberToAdd
+}
```
ryanzhang-oss, line 546: I am not sure if there is benefit of this type of refactor if we don't add UT for those functions.

ryanzhang-oss: and looking at the logic, they seem to be too straightforward to warrantee UT

ryanzhang-oss (commented):
it would good to have an integration test to test this

pkg/controllers/rollout/controller.go
```diff
@@ -408,30 +411,73 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
+	return toBeUpdatedBindingList, staleUnselectedBinding, true, minWaitTime, nil
+}
+
+func checkReadyConditions(binding *fleetv1beta1.ClusterResourceBinding) bool {
+	appliedCondition := binding.GetCondition(string(fleetv1beta1.ResourceBindingApplied))
+	availableCondition := binding.GetCondition(string(fleetv1beta1.ResourceBindingAvailable))
+	if condition.IsConditionStatusFalse(appliedCondition, binding.Generation) || condition.IsConditionStatusFalse(availableCondition, binding.Generation) {
+		return false
+	}
+	return true
```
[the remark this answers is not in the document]
britaniar: Is it not the same line as this https://github.com/Azure/fleet/blob/d48b6b18618677dab2f9e5cf4cc4a615b51210c8/pkg/controllers/rollout/controller.go#L346C4-L346C153?

pkg/controllers/rollout/controller.go
britaniar: I took them out of the function because of this error https://github.com/Azure/fleet/actions/runs/10154135990/job/28078753503. Most of the "refactoring" was just breaking up the function into smaller, more manageable functions.

pkg/controllers/rollout/controller.go
britaniar: yes, checks if they are all ready to be able to set the minimum waitTime to 0.

pkg/controllers/rollout/controller.go
```diff
@@ -333,25 +331,30 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
 	// classify the bindings into different categories
-	// TODO: calculate the time we need to wait for the first applied but not ready binding to be ready.
+	// Wait for the first applied but not ready binding to be ready.
 	// return wait time longer if the rollout is stuck on failed apply/available bindings
+	minWaitTime := time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds) * time.Second
+	allReady := true
 	crpKObj := klog.KObj(crp)
-	for idx := range allBindings {
-		binding := allBindings[idx]
+	for _, binding := range allBindings {
```
ryanzhang-oss, line 339: do you know the difference between this and the original?

pkg/controllers/rollout/controller.go
```diff
@@ -333,25 +331,30 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
 				canBeReadyBindings = append(canBeReadyBindings, binding)
 			}
-			_, bindingReady := isBindingReady(binding, readyTimeCutOff)
+			waitTime, bindingReady := isBindingReady(binding, readyTimeCutOff)
 			if bindingReady {
 				klog.V(3).InfoS("Found a ready unscheduled binding", "clusterResourcePlacement", crpKObj, "binding", bindingKObj)
 				readyBindings = append(readyBindings, binding)
+			} else {
+				allReady = false
+				klog.Info("Binding WaitTime:", waitTime)
```
ryanzhang-oss, line 354: please clean up the log code, don't leave debugging logs in the PR.

pkg/controllers/rollout/controller.go
```diff
@@ -362,40 +365,40 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
-			if condition.IsConditionStatusFalse(appliedCondition, binding.Generation) || condition.IsConditionStatusFalse(availableCondition, binding.Generation) {
-				klog.V(3).InfoS("Found a failed to be ready bound binding", "clusterResourcePlacement", crpKObj, "binding", bindingKObj)
-				bindingFailed = true
-			} else {
+			bindingFailed := false
+			if isBindingAvailable(binding) {
 				canBeReadyBindings = append(canBeReadyBindings, binding)
+			} else {
+				bindingFailed = true
+				klog.V(3).InfoS("Found a failed to be ready bound binding", "clusterResourcePlacement", crpKObj, "binding", bindingKObj)
```
ryanzhang-oss, line 395: this is not the same logic as before (it's flipped) and not correct.

pkg/controllers/rollout/controller.go
```diff
@@ -408,30 +411,49 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
 			}
 		}
 	}
+	if minWaitTime == time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds)*time.Second && allReady {
+		minWaitTime = 0
+	}
+	klog.Info("MinWaitTime:", minWaitTime)
```
ryanzhang-oss, line 417: log

pkg/controllers/rollout/controller.go
```diff
@@ -408,30 +411,49 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
-	klog.V(2).InfoS("Calculated the max number of bindings to remove", "clusterResourcePlacement", crpKObj,
-		"maxUnavailableNumber", maxUnavailableNumber, "minAvailableNumber", minAvailableNumber,
-		"lowerBoundAvailableBindings", lowerBoundAvailableNumber, "maxNumberOfBindingsToRemove", maxNumberToRemove)
+	toBeUpdatedBindingList, staleUnselectedBinding := determineBindingsToUpdate(crp, removeCandidates, updateCandidates, boundingCandidates, applyFailedUpdateCandidates, targetNumber,
+		readyBindings, canBeReadyBindings, canBeUnavailableBindings)
+
+	return toBeUpdatedBindingList, staleUnselectedBinding, true, minWaitTime, nil
+}
+
+func isBindingAvailable(binding *fleetv1beta1.ClusterResourceBinding) bool {
```
ryanzhang-oss, line 438: the name of the function is not what its content says, it's very mis-leading.

pkg/controllers/rollout/controller.go
britaniar: Yes, did not mean to change it. It was part of a different refactoring approach. Restored to the original.

pkg/controllers/rollout/controller.go
```diff
@@ -408,30 +411,49 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
 			}
 		}
 	}
+	if minWaitTime == time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds)*time.Second && allReady {
```
ryanzhang-oss, line 414: when will minWaitTime != time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds)*time.Second && allReady

pkg/controllers/rollout/controller.go
```diff
@@ -408,30 +411,49 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
...
+		minWaitTime = 0
+	}
 
+	// Calculate target number
 	targetNumber := r.calculateRealTarget(crp, schedulerTargetedBinds)
-	klog.V(2).InfoS("Calculated the targetNumber", "clusterResourcePlacement", crpKObj,
+	klog.V(2).InfoS("Calculated the targetNumber", "clusterResourcePlacement", klog.KObj(crp),
```
ryanzhang-oss, line 420: is crpKObj changed?

ryanzhang-oss (approved):

PR #890, opened 2024-07-24
fix: fix the kind join script

ryanzhang-oss:

Fix the kind join script as kind cluster can't pull images directly from MCR

Fixes #

I have:

### DIFF
--- a/docs/getting-started/on-prem.md
+++ b/docs/getting-started/on-prem.md
@@ -69,7 +69,7 @@ helm install hub-agent fleet/charts/hub-agent/ \
     --set image.pullPolicy=Always \
     --set image.repository=$REGISTRY/$HUB_AGENT_IMAGE \
     --set image.tag=$FLEET_VERSION \
-    --set logVerbosity=2 \
+    --set logVerbosity=5 \
     --set namespace=fleet-system \
     --set enableWebhook=true \
     --set webhookClientConnectionType=service \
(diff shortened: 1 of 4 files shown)


test/e2e/setup.sh
```diff
@@ -6,7 +6,7 @@ set -o pipefail
 
 # Before updating the default kind image to use, verify that the version is supported
 # by the current kind release.
-KIND_IMAGE="${KIND_IMAGE:-kindest/node:v1.28.0}"
```
britaniar, line 9: If updating the version here, we could update the version in the [Makefile](https://github.com/Azure/fleet/blob/main/Makefile) and [ReadMe](https://github.com/Azure/fleet/blob/main/README.md)

docs/getting-started/on-prem.md
```diff
@@ -107,8 +107,8 @@ export MEMBER_CLUSTER=YOUR-MEMBER-CLUSTER
 export MEMBER_CLUSTER_CONTEXT=YOUR-MEMBER-CLUSTER-CONTEXT
 
 # Run the script.
-chmod +x fleet/hack/membership/join.sh
-./fleet/hack/membership/join.sh
+chmod +x fleet/hack/hack/Azure/setup/joinMC.sh
```
britaniar, line 110: Maybe we should try to keep these consistent across the docs update the previous steps as well to match kind.md
```suggestion
chmod +x ./hack/Azure/setup/joinMC.sh
```

test/e2e/setup.sh
ryanzhang-oss: I will leave this to another PR

britaniar (approved):

PR #897, opened 2024-08-05
feat: add rollout API spec

ryanzhang-oss:

An overall idea of rollout API

Fixes #

I have:


apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope="Cluster",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// StageRollout defines a group of override policies about how to override the selected cluster scope resources
+// to target clusters.
+type StageRollout struct {
```
michaelawyu, line 21: Hi Ryan! This will be a separate GVK, how would the users be able to use it?

michaelawyu: I assume that since there's a separate `RolloutStrategyType`, we would also need a config field like `RollingUpdateConfig` in the CRP API, with the field itself being a selector that selects a GVK instance; is my understanding correct?

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope="Cluster",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// StageRollout defines a group of override policies about how to override the selected cluster scope resources
```
michaelawyu, line 19: Hi Ryan! This comment does not look right to me; is it correct?

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+	// +required
+	StageRollout []StageRolloutConfig `json:"stages"`
+
+	// The wait time between each stage is completed.
+	// +optional
+	WaitTimeBetweenStage *metav1.Duration `json:"waitTimeBetweenStage,omitempty"`
+
+	// TODO: Add alerting configuration.
+
+	// TODO: Add health-check configuration.
```
michaelawyu, line 46: Hi Ryan! For the P0 we would also need at least a manual gate between stages.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+	// LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected.
+	// We don't check if there are overlap between clusters selected by different stages.
+	LabelSelector *metav1.LabelSelector `json:"labelSelector"`
+
+	// The label key used to sort the selected clusters.
+	// The clusters within the stage are updated following the rule below:
+	//   - primary: Ascending order based on the value of the label key, interpreted as integers if present.
+	//   - secondary: Ascending order based on the name of the cluster if the label key is absent.
+	// +optional
+	SortingLabelKey *string `json:"sortingLabelKey,omitempty"`
```
michaelawyu, line 64: Hi Ryan! Do you think it would be a concern that if there are multiple CRPs targeting the same cluster, this pattern of weight management by label would be a bit cumbersome for users?

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// The maximum number of clusters that can be updated simultaneously within this stage
+	// default is 1.
+	// +optional
+	MaxParallel *intstr.IntOrString `json:"maxParallel,omitempty"`
```
michaelawyu, line 69: Just for clarification, for now do we still proceed within a stage if the first few clusters fail? I know that there's a TO-DO field about liveness checks; if I understand it correctly, we would allow users to fine-tune this behavior, right?

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -488,6 +488,10 @@ const (
 	// RollingUpdateRolloutStrategyType replaces the old placed resource using rolling update
 	// i.e. gradually create the new one while replace the old ones.
 	RollingUpdateRolloutStrategyType RolloutStrategyType = "RollingUpdate"
+
+	// StageRollingUpdateRolloutStrategyType replaces the old placed resource stage by stage
```
michaelawyu, line 492: Hi Ryan! Infosys mentioned that this would apply to first-time placement as well.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+	// the clusters that are actively updating.
+	// +optional
+	CurrentUpdatingClusters []string `json:"currentUpdatingClusters,omitempty"`
+
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	//
+	// Conditions is an array of current observed conditions for ClusterResourceBinding.
```
michaelawyu, line 94: Hi Ryan! The comment needs an update I think.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+// StageRolloutConfig describes a single rollout stage group configuration.
+type StageRolloutConfig struct {
+	// The name of the stage. This should be unique within the StageRollout.
+	// +required
+	Name string `json:"name"`
+
+	// LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected.
+	// We don't check if there are overlap between clusters selected by different stages.
+	LabelSelector *metav1.LabelSelector `json:"labelSelector"`
```
michaelawyu, line 57: Hi Ryan! If a cluster appears in a prev. stage, would the controller go back and apply resources to that controller first, or would the controller re-start only after all stages are completed? This is just for clarification of course.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+	//
+	// Conditions is an array of current observed conditions for ClusterResourceBinding.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+}
+
+// StageRolloutConditionType identifies a specific condition of the StageRollout.
+type StageRolloutConditionType string
+
+const (
```
michaelawyu, line 102: Hi Ryan! Just for clarification, from the comments it seems that `StageRolloutWaiting` is just a special case of `StageRollingOut` being true?

michaelawyu (commented):
Added some comments, PTAL 🙏

apis/placement/v1alpha1/stagerollout_types.go
ryanzhang-oss: just to be clear, it's a "hook" to inject "go ahead" signal to our rollout between stages.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// The wait time after all the clusters in this stage are updated.
+	// This will override the global waitTimeBetweenStage for this specific stage
+	// +optional
+	WaitTime *metav1.Duration `json:"waitTime,omitempty"`
+}
+
+// StageRolloutStatus defines the observed state of the StageRollout.
+type StageRolloutStatus struct {
+	// The stage that is in the middle of the rollout.
```
ryanzhang-oss, line 79: // Need resourceSnapshotIndex

apis/placement/v1alpha1/stagerollout_types.go
ryanzhang-oss: TODO: clarify if we should restart a previous stage if there are changes in the previous stage.

apis/placement/v1alpha1/stagerollout_types.go
ryanzhang-oss: // Need policySnapshot too

ryanzhang-oss: restart the rollout if any of these two changes

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,135 @@
...
+	CurrentStage int `json:"stage"`
+
+	// the clusters that are in the current stage.
+	ClustersInCurrentStage []string `json:"clustersInCurrentStage"`
```
ryanzhang-oss, line 83: probably need to list all clusters in each stage [][]string, []stage

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,299 @@
...
+	// +required
+	Stages []StageRolloutConfig `json:"stages"`
+
+	// The common post rollout tasks that ALL stages needs to be completed before moving to the next stage.
+	// +optional
+	PostRolloutTasks PostRolloutTask `json:"postRolloutTasks,omitempty"`
+}
+
+// StageRolloutConfig describes a single rollout stage group configuration.
+type StageRolloutConfig struct {
```
zhiying-lin, line 56: how do we rollout the selected clusters within a stage?

sequentially or concurrently?

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,299 @@
...
+	// TODO: Add alerting configuration.
+
+}
+
+// PostRolloutTask is the collection of post rollout tasks that ALL need to be completed before we can move to the next stage.
+type PostRolloutTask struct {
+	// The name of the RolloutApprovalRequest we need to create to get the approval for the next stage.
+	// An RolloutApprovalRequest object will be created if this is not empty.
+	// +optional
+	ApprovalRequestName string `json:"approvalTask,omitempty"`
```
zhiying-lin, line 87: keptn uses an generic data type to indicate the state and triggers the task by the state,
https://v1.keptn.sh/docs/1.0.x/define/triggers/


```
sequences:
  - name: "rollback"
    triggeredOn:
      - event: "production.delivery.finished"
        selector:
          match:
            result: failed
```

 https://github.com/keptn/spec/blob/0.2.2/cloudevents.md#evaluation-triggered

The pros is we don't need to create a CRD for each task.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// The time to wait after all the clusters in the current stage complete the update
+	// before we move to the next stage.
+	// +optional
+	WaitTime *metav1.Duration `json:"waitTime,omitempty"`
+}
```
zhiying-lin, line 93:
```suggestion
type PostRolloutTask struct {
	type TaskType // approval, wait

	// The time to wait after all the clusters in the current stage complete the update
	// before we move to the next stage.
	// +optional
	WaitTime *metav1.Duration `json:"waitTime,omitempty"`
}
```
PostRolloutTasks []PostRolloutTask

Then user can define their own sequence. Otherwise, it's hard to tell which one runs first.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,299 @@
...
+	Status StageRolloutStatus `json:"status,omitempty"`
+}
+
+// StageRolloutSpec defines the desired the rollout sequence stage by stage.
+type StageRolloutSpec struct {
+	// The name of the CRP that this rollout is applied to.
+	// There can only be one active rollout for each ClusterResourcePlacement,
+	// we will fail the rollout if there is already an active rollout.
+	// +required
+	ParentCRPName string `json:"parentCRPName"`
```
zhiying-lin, line 39: i'm thinking of the other way. CRP refers to the StageRollout name.
StageRollout will be a common rollout configurations for multiple CRP.


All the rollout status will be part of the CRP status, like today.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// The resource snapshot index that this rollout is based on.
+	// The rollout will use the latest resource snapshot if this is empty.
+	// +optional
+	ResourceSnapShotIndex string `json:"resourceSnapShotIndex,omitempty"`
```
zhiying-lin, line 44: we also need to know the override snapshot index.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,299 @@
...
+
+	// Stage rollout configurations for each rollout stage.
+	// +required
+	Stages []StageRolloutConfig `json:"stages"`
```
ryanzhang-oss, line 48: what happened to the "de" selected clusters?

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,299 @@
...
+	//   - primary: Ascending order based on the value of the label key, interpreted as integers if present.
+	//   - secondary: Ascending order based on the name of the cluster if the label key is absent.
+	// +optional
+	SortingLabelKey *string `json:"sortingLabelKey,omitempty"`
+
+	// The post rollout tasks that this stage needs to complete before moving to the next stage in-addition to
+	// the common post rollout tasks. This setting will override the common post rollout tasks if there are
+	// duplicate type of tasks with different settings.
+	// +optional
+	PostRolloutTasks PostRolloutTask `json:"postRolloutTasks,omitempty"`
```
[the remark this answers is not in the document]
ryanzhang-oss: removed the common one for now to simplify

apis/placement/v1alpha1/stagerollout_types.go
ryanzhang-oss: My current idea is that they run at the same time. For example, we don't wait before or after the approval, the wait time is the minimum time needed between stages.

ryanzhang-oss: Also, I thought about the type pattern. The problem with that is each type requires a distinct set of parameters which pretty much makes the `taskType` field redundant.

apis/placement/v1alpha1/stagerollout_types.go
ryanzhang-oss: who will send the cloudEvent? How is it send and received?

zhiying-lin: my bad, i thought it was a CR.

apis/placement/v1alpha1/stagerollout_types.go
```diff
@@ -0,0 +1,344 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Namespaced",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// RolloutApprovalRequest defines a stage by stage rollout policy that is applied to the ClusterResourcePlacement of the same name.
+// Each snapshot MUST have the following labels:
+//   - `ParentStageRollout` which points to its owner stage rollout.
+//   - `TargetStage` which is the name of the stage that this approval request is for.
+//   - `IsLatestRolloutApproval` which indicates whether this approval request is the latest one related to this rollout
+type RolloutApprovalRequest struct {
```
circy9, line 267: Which namespace should the rollout approve request be in? Will it be in the same one as rollout update run? How do we manage access control of approve requests vs runs vs strategies?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+	// The observed status of StagedUpdateStrategy.
+	// +optional
+	Status StagedUpdateStrategyStatus `json:"status,omitempty"`
+}
+
+// StagedUpdateStrategySpec defines the desired the update strategy.
+type StagedUpdateStrategySpec struct {
+	// Staged update configurations for each update stage.
+	// +required
+	Stages []StagedUpdateConfig `json:"stages"`
```
circy9, line 95: stageconfig

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// TODO: Add alerting configuration.
+}
+
+// StagedUpdateStrategyStatus defines the observed state of the StagedUpdateStrategy.
+type StagedUpdateStrategyStatus struct {
+	// ComputedStages list the name of the clusters in each stage.
+	// The clusters in each stage are ordered following its order to be rolled out.
+	// +required
+	ComputedStages [][]string `json:"currentStages"`
```
circy9, line 105: We shouldn't compute computed stages for strategy.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+	// The clusters within the stage are updated sequentially following the rule below:
+	//   - primary: Ascending order based on the value of the label key, interpreted as integers if present.
+	//   - secondary: Ascending order based on the name of the cluster if the label key is absent.
+	// +optional
+	SortingLabelKey *string `json:"sortingLabelKey,omitempty"`
+
+	// The collection of tasks that each stage needs to complete successfully before moving to the next stage.
+	// Each task is executed in parallel.
+	// +optional
+	PostStageTasks []PostStageTask `json:"postRolloutTasks,omitempty"`
```
circy9, line 139: AfterStageTasks

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+// PostStageTask is the collection of post stage tasks that ALL need to be completed before we can move to the next stage.
+type PostStageTask struct {
+	// The type of the post stage task.
+	// +kubebuilder:validation:Enum=TimedWait;ManualApproval
```
circy9, line 145: Approval
TimedWait

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+
+	// The label key used to sort the selected clusters.
+	// The clusters within the stage are updated sequentially following the rule below:
+	//   - primary: Ascending order based on the value of the label key, interpreted as integers if present.
+	//   - secondary: Ascending order based on the name of the cluster if the label key is absent.
+	// +optional
+	SortingLabelKey *string `json:"sortingLabelKey,omitempty"`
+
+	// The collection of tasks that each stage needs to complete successfully before moving to the next stage.
+	// Each task is executed in parallel.
```
circy9, line 137: Each type can appear at most once.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+type PostStageTask struct {
+	// The type of the post stage task.
+	// +kubebuilder:validation:Enum=TimedWait;ManualApproval
+	// +required
+	Type PostStageTaskType `json:"type"`
+
+	// The name of the StagedUpdateApprovalRequest we need to create to get the approval for the next stage.
+	// We will generate a name with a prefix of the StagedUpdateRun name if it's empty.
+	// +optional
+	ApprovalRequestName string `json:"approvalTask,omitempty"`
```
circy9, line 152: ApprovalRequestPrefix: We will add prefix-runName-stageName as post fix.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+	// +kubebuilder:default:="1h"
+	// +kubebuilder:validation:Pattern="^0|([0-9]+(\\.[0-9]+)?(s|m|h))+$"
+	// +kubebuilder:validation:Type:=string
+	// +optional
+	WaitTime metav1.Duration `json:"waitTime,omitempty"`
+}
+
+// StagedUpdateRunStatus defines the observed state of the StagedUpdateRun.
+type StagedUpdateRunStatus struct {
+	// ComputedStages list the name of the clusters in each stage when the updateRun started.
```
circy9, line 164: Add: CurrentStrategy to take a snapshot of the strategy when the run is created. It is immutable.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+// unselected after all the stages explicitly defined in the updateStrategy complete.
+// Each StagedUpdateRun object corresponds to a single "release" of a certain version of the resources.
+// The release is abandoned if the StagedUpdateRun object is deleted or the scheduling decision changes.
+type StagedUpdateRun struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// The desired state of StagedUpdateRun.
+	// +required
+	Spec StagedUpdateRunSpec `json:"spec"`
```
circy9, line 30: We should make run spec immutable such that we don't go through the stages N times.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+	WaitTime metav1.Duration `json:"waitTime,omitempty"`
+}
+
+// StagedUpdateRunStatus defines the observed state of the StagedUpdateRun.
+type StagedUpdateRunStatus struct {
+	// ComputedStages list the name of the clusters in each stage when the updateRun started.
+	// The clusters in each stage are ordered following its order to be rolled out.
+	// The updateRun will stop if the cluster stages change during the update.
+	// +required
+	ComputedStages [][]string `json:"currentStages"`
```
circy9, line 168: Use []StageStatus which contains each cluster's status. See this one: https://learn.microsoft.com/en-us/rest/api/fleet/update-runs/get?view=rest-fleet-2023-10-15&tabs=HTTP

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+	// We will NOT update the strategy in the update even if the appliedStrategy changes in the CRP.
+	// +required
+	CurrentAppliedStrategy string `json:"currentAppliedStrategy"`
+
+	// The collection of clusters on which we are going to remove the application from.
+	// We restart the update if CRP changes the selected clusters during our update.
+	// Therefore, we need to keep track of the clusters that we have applied the resources to but no long is selected.
+	// We remove the application from these clusters after all the stages are completed and remove them from this list.
+	// +required
+	ToBeRevertedClusters []string `json:"toBeRevertedClusters"`
```
circy9, line 187: DeletionStageStatus StageStatus
contains a list of clusters whose work will be deleted ordered by the given sort label key.

apis/placement/v1alpha1/stagedupdate_types.go
circy9, line 168: We should add UpdateStatus at each level: cluster, stage, run.

https://learn.microsoft.com/en-us/rest/api/fleet/update-runs/get?view=rest-fleet-2023-10-15&tabs=HTTP#updatestatus

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Namespaced",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// StagedUpdateRun defines a stage by stage update run that applies the selected resources by the
+// corresponding ClusterResourcePlacement to its selected clusters. We remove the resources from the clusters that are
+// unselected after all the stages explicitly defined in the updateStrategy complete.
+// Each StagedUpdateRun object corresponds to a single "release" of a certain version of the resources.
+// The release is abandoned if the StagedUpdateRun object is deleted or the scheduling decision changes.
+type StagedUpdateRun struct {
```
circy9, line 24: If binding changes, we will abort the run.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,382 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Namespaced",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// StagedUpdateRun defines a stage by stage update run that applies the selected resources by the
+// corresponding ClusterResourcePlacement to its selected clusters. We remove the resources from the clusters that are
+// unselected after all the stages explicitly defined in the updateStrategy complete.
+// Each StagedUpdateRun object corresponds to a single "release" of a certain version of the resources.
+// The release is abandoned if the StagedUpdateRun object is deleted or the scheduling decision changes.
```
circy9, line 23: If the run failed, how can the user resume the run?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+	// in which the application will be updated on the member clusters.
+	// +required
+	StagedUpdateStrategyRef StagedUpdateStrategyReference `json:"stagedRolloutStrategyRef"`
+}
+
+// PlacementReference is a reference to a placement object.
+type PlacementReference struct {
+	// Name is the name of the referenced StagedUpdateStrategy.
+	Name string `json:"name"`
+	// Namespace is the namespace where the referenced StagedUpdateStrategy is located.
```
circy9, line 67: "the referenced placement"

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
```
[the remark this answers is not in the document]
circy9: Since we don't support resource placement today. Shall we remove namespace for now?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+// StageConfig describes a single update stage group configuration.
+// The clusters in each stage are updated sequentially for now.
+// We will stop the update if any of the updates fail.
+type StageConfig struct {
+	// The name of the stage. This MUST be unique within the same StagedUpdateStrategy.
+	// +required
+	Name string `json:"name"`
+
+	// LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected.
+	// We don't check if there are overlap between clusters selected by different stages.
```
circy9, line 119: We will check when a run is created for this strategy, right?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+type AfterStageTask struct {
+	// The type of the after stage task.
+	// +kubebuilder:validation:Enum=TimedWait;Approval
+	// +required
+	Type AfterStageTaskType `json:"type"`
+
+	// The prefix of the name of the StageApprovalRequest we need to create to get the approval for the next stage.
+	// We create the full name of the approval request by concatenating the prefix to the stagedUpdateRunName and the stageName.
+	// +optional
+	ApprovalRequestPrefix string `json:"approvalTask,omitempty"`
```
circy9, line 145: should use *string for optional?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+
+// AfterStageTask is the collection of post stage tasks that ALL need to be completed before we can move to the next stage.
+type AfterStageTask struct {
+	// The type of the after stage task.
+	// +kubebuilder:validation:Enum=TimedWait;Approval
+	// +required
+	Type AfterStageTaskType `json:"type"`
+
+	// The prefix of the name of the StageApprovalRequest we need to create to get the approval for the next stage.
+	// We create the full name of the approval request by concatenating the prefix to the stagedUpdateRunName and the stageName.
```
circy9, line 143: Could you provide the approval request name template and example?
<prefix>-<runName>-<stageName>, e.g., "myPrefix-run20240908-stage1"

BTW, do we really need the prefix? Isn't "run20240908-stage1" good enough? When is the prefix useful?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+
+// StagedUpdateStrategy defines a reusable strategy that specifies the stages and the sequence
+// in which the application will be updated on the member clusters.
+type StagedUpdateStrategy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Stage configurations for each update stage.
+	// +required
+	Stages []StageConfig `json:"stages"`
```
circy9, line 96: Shall we have a type for the strategy as it may have more fields beyond stages? You can use the same type in UPdateRunStatus, rather than copying N fields to status. 

type XXX {
    maxUnavailable *int64
    Stages []StageConfig
}

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+type StagedUpdateRunStatus struct {
+	// UpdateRunStrategy is the collection of the stageConfig in the StagedUpdateStrategy that we are going to use for the update.
+	// +required
+	UpdateRunStrategy []StageConfig `json:"updateRunStrategy"`
+
+	// AppliedStrategy is applied strategy of the CRP that we are currently using.
+	// It is the same as the strategy in the CRP when we first start the update.
+	// We will NOT update the strategy in the update even if the appliedStrategy changes in the CRP.
+	// +required
+	AppliedStrategy v1beta1.ApplyStrategy `json:"appliedStrategy"`
```
circy9, line 165: ApplyStrategy instead of AppliedSrategy? AppliedStrategy is too generic. L159 is also an applied strategy.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+	// +required
+	StageName string `json:"stageName"`
+
+	// The list of clusters in the stage.
+	// +required
+	Clusters []string `json:"clusters"`
+
+	// The status of the post update tasks that are associated with current stage.
+	// +optional
+	CurrentPostStageTaskStatus []AfterStageTaskStatus `json:"currentPostStageTaskStatus,omitempty"`
```
circy9, line 202: CurrentPostStageTaskStatus -> AfterStateTaskStatus

Don't keep introducing new names?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+type StageStatus struct {
+	// The name of the stage.
+	// +required
+	StageName string `json:"stageName"`
+
+	// The list of clusters in the stage.
+	// +required
+	Clusters []string `json:"clusters"`
```
circy9, line 198: ClusterStatus []ClusterStatus

Could you provide a status for each cluster?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+	// +listType=map
+	// +listMapKey=type
+	//
+	// Conditions is an array of current observed conditions for StagedUpdateRun.
+	// Known conditions are "RollingOut".
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+}
+
+type StageStatus struct {
```
circy9, line 191: Could you add conditions for the run, each stage, and each cluster? Currently only after stage tasks have conditions.

The condition type could be the same, like pending -> updating (as this is a staged update) -> succeeded/failed.

"CurrentUpdatingStage" is unnecessary if you have conditions per stage.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+	// +required
+	PlacementRef PlacementReference `json:"placementRef"`
+
+	// The resource snapshot index that this update run is based on.
+	// +required
+	ResourceSnapShotIndex string `json:"resourceSnapShotIndex"`
+
+	// The override snapshot index that this update run is based on.
+	// +required
+	OverrideSnapShotIndex string `json:"overrideSnapShotIndex"`
```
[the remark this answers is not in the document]
ryanzhang-oss: This has to go to the status field instead.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+	Items           []StagedUpdateStrategy `json:"items"`
+}
+
+// StageConfig describes a single update stage group configuration.
+// The clusters in each stage are updated sequentially for now.
+// We will stop the update if any of the updates fail.
+type StageConfig struct {
+	// The name of the stage. This MUST be unique within the same StagedUpdateStrategy.
+	// +required
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
ryanzhang-oss: not at the moment since the rest of the fields don't really apply to that stage. We were thinking of mix the to be deleted clusters with the new/updated clusters but want to keep it simple for now.

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: string has a "zero" value so it's okay.

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: I am not sure if we need conditions inside each stage anymore since the updateStatus  has all the info we need (and  we can add more if needed). Only overall condition missing is in the updateRun itself.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="The spec field is immutable"
+	Spec StagedUpdateRunSpec `json:"spec"`
+
+	// The observed status of StagedUpdateRun.
+	// +kubebuilder:validation:Optional
+	Status StagedUpdateRunStatus `json:"status,omitempty"`
+}
+
+// StagedUpdateRunSpec defines the desired the update strategy and the snapshot indices
+// on which the update is based.
```
circy9, line 41: // StagedUpdateRunSpec contains the snapshot index of the selected resources to be updated, and the stage-by-stage strategy to update them across selected cluster, for the specified ResourcePlacement object.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+type StagedUpdateRunSpec struct {
+	// A reference to the placement that this update run is applied to.
+	// There can be multiple active update runs for each placement but
+	// it's up to the devOps to make sure they don't conflict with each other.
+	// +kubebuilder:validation:Required
+	PlacementRef PlacementReference `json:"placementRef"`
+
+	// The resource snapshot index that this update run is based on.
```
circy9, line 49:
```suggestion
	// The snapshot index of the selected resources to be updated across clusters.
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+	// +kubebuilder:validation:Required
+	ResourceSnapShotIndex string `json:"resourceSnapShotIndex"`
+
+	// The reference to the update strategy that specifies the stages and the sequence
+	// in which the application will be updated on the member clusters. We will compute
```
circy9, line 54:
```suggestion
	// in which the selected resources will be updated on the member clusters. We will compute
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+
+// +genclient
+// +genclient:namespaced
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Namespaced",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// StagedUpdateStrategy defines a reusable strategy that specifies the stages and the sequence
+// in which the application will be updated on the member clusters.
```
circy9, line 76:
```suggestion
// in which the selected resources will be updated on the member clusters.
```

apis/placement/v1alpha1/stagedupdate_types.go
circy9: Not resolved?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+// StagedUpdateStrategyList contains a list of StagedUpdateStrategy.
+// +kubebuilder:resource:scope="Namespaced"
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type StagedUpdateStrategyList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []StagedUpdateStrategy `json:"items"`
+}
+
+// StageConfig describes a single update stage group configuration.
```
circy9, line 96:
```suggestion
// StageConfig describes a single update stage.
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+// The clusters in each stage are updated sequentially for now.
+// We will stop the update if any of the updates fail.
+type StageConfig struct {
+	// The name of the stage. This MUST be unique within the same StagedUpdateStrategy.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected
+	// for this stage. There cannot be overlapping clusters between stages when the stagedUpdateRun is created.
+	// If the label selector is absent, all the selected clusters are selected.
```
circy9, line 106:
```suggestion
	// If the label selector is absent, the stage includes all the selected clusters.
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+// StageConfig describes a single update stage group configuration.
+// The clusters in each stage are updated sequentially for now.
+// We will stop the update if any of the updates fail.
+type StageConfig struct {
+	// The name of the stage. This MUST be unique within the same StagedUpdateStrategy.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected
+	// for this stage. There cannot be overlapping clusters between stages when the stagedUpdateRun is created.
```
circy9, line 105: What happens if some selected cluster aren't included by any stage?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+	// If the label selector is absent, all the selected clusters are selected.
+	// +kubebuilder:validation:Optional
+	LabelSelector *metav1.LabelSelector `json:"labelSelector"`
+
+	// The label key used to sort the selected clusters.
+	// The clusters within the stage are updated sequentially following the rule below:
+	//   - primary: Ascending order based on the value of the label key, interpreted as integers if present.
+	//   - secondary: Ascending order based on the name of the cluster if the label key is absent.
```
circy9, line 113:
```suggestion
	//   - secondary: Ascending order based on the name of the cluster if the label key is absent or the label value is the same.
```

apis/placement/v1alpha1/stagedupdate_types.go
circy9: Unresovled?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+type AfterStageTask struct {
+	// The type of the after stage task.
+	// +kubebuilder:validation:Enum=TimedWait;Approval
+	// +kubebuilder:validation:Required
+	Type AfterStageTaskType `json:"type"`
+
+	// The prefix of the name of the StageApprovalRequest we need to create to get the approval for the next stage.
+	// We create the full name of the approval request by concatenating the prefix to the stagedUpdateRunName and the stageName.
+	// For example, if the approvalRequestPrefix is "myApproval", the name of the approval request for the stage "stage1"
+	// in the update run "updateRun1" is "myApproval-updateRun1-stage1".
```
circy9, line 133: Shall we add a timestamp based or a random postfix to avoid name collisions?

Since spec is immutable, customers might need to delete and recreate the same upddaterun when the updaterun fails, which will lead to the same approval request name.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+	// CurrentStageIndex is the name of the stage that is in the middle of the update.
+	// If the update run is not started, the value is an empty string.
+	// If we are in the DeletionStage, the name is called "deletionStage".
+	// +kubebuilder:validation:Optional
+	CurrentUpdatingStage string `json:"currentUpdatingStage,omitempty"`
+
+	// Stages list the current status of each stage.
+	// The list is empty if the update run is not started.
+	// +kubebuilder:validation:Optional
+	StagesStatus []StageStatus `json:"stagesStatus,omitempty"`
```
circy9, line 183: Just "Stages" as everything is part of "XXXStatus {}"?

Sometimes you use Foos []FooStatus, other times you use FoosStatus []FooStatus. Be consistent?

See below:
Clusters []ClusterUpdateStatus

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+}
+
+// UpdateStatus defines the status of a resource in an update run.
+type UpdateStatus struct {
+	// The status of the update on the cluster.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=NotStarted;Running;Completed;Failed;WaitingForApproval
+	State UpdateState `json:"state"`
+
+	// message is a human readable message indicating details about the update status.
```
circy9, line 237:
```suggestion
	// Message is a human readable message containing details about the update status.
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	//
+	// Conditions is an array of current observed conditions for the specific type of post update task.
+	// Known conditions are "ApprovalRequestCreated", "WaitTimeElapsed", and "ApprovalRequestApproved".
+	// +kubebuilder:validation:Optional
+	Conditions []metav1.Condition `json:"conditions"`
```
circy9, line 291: Can you reuse UpdateStatus instead of Conditions? Maybe rename to StepStatus? Everything: a stage, a cluster, a after stage task is a step in the staged update run.

You also need to record start time and complete time of the step. This is especially useful for TimedWait so the user knows when the wait started and when it ended.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+	// The status of the update run in the stage.
+	// +kubebuilder:validation:Required
+	Status UpdateStatus `json:"status"`
+}
+
+// UpdateStatus defines the status of a resource in an update run.
+type UpdateStatus struct {
+	// The status of the update on the cluster.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=NotStarted;Running;Completed;Failed;WaitingForApproval
```
circy9, line 234: Waiting instead of WaitingForApproval? You could be waiting for timedwait or for approval

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	//
+	// Conditions is an array of current observed conditions for the specific type of post update task.
+	// Known conditions are "ApprovalRequestCreated", "WaitTimeElapsed", and "ApprovalRequestApproved".
+	// +kubebuilder:validation:Optional
+	Conditions []metav1.Condition `json:"conditions"`
+}
```
circy9, line 292: Add ApprovalRequest which is the actual name of the approval request? Not sure whether you need namespace too.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+type StageConfig struct {
+	// The name of the stage. This MUST be unique within the same StagedUpdateStrategy.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected
+	// for this stage. There cannot be overlapping clusters between stages when the stagedUpdateRun is created.
+	// If the label selector is absent, all the selected clusters are selected.
+	// +kubebuilder:validation:Optional
+	LabelSelector *metav1.LabelSelector `json:"labelSelector"`
```
jwtty, line 108: We are going to do an intersection between the clusters selected by scheduler and the clusters selected here right?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+
+	// The label key used to sort the selected clusters.
+	// The clusters within the stage are updated sequentially following the rule below:
+	//   - primary: Ascending order based on the value of the label key, interpreted as integers if present.
+	//   - secondary: Ascending order based on the name of the cluster if the label key is absent.
+	// +kubebuilder:validation:Optional
+	SortingLabelKey *string `json:"sortingLabelKey,omitempty"`
+
+	// The collection of tasks that each stage needs to complete successfully before moving to the next stage.
+	// Each task is executed in parallel and they can not be more than one task of the same type.
```
jwtty, line 118:
```suggestion
	// Each task is executed in parallel and there can not be more than one task of the same type.
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+	ResourceOverrideSnapshots []v1beta1.NamespacedName `json:"resourceOverrideSnapshots,omitempty"`
+
+	// ClusterResourceOverrides contains a list of applicable ClusterResourceOverride snapshot names associated with the
+	// selected resources.
+	// The list computed as the beginning of the update run and not updated during the update run.
+	// The list is empty if there are no cluster overrides associated with the selected resources.
+	// +kubebuilder:validation:Optional
+	ClusterResourceOverrideSnapshots []string `json:"clusterResourceOverrideSnapshots,omitempty"`
+
+	// CurrentStageIndex is the name of the stage that is in the middle of the update.
```
jwtty, line 174:
```suggestion
	// CurrentUpdatingStage is the name of the stage that is in the middle of the update.
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Namespaced",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// StageApprovalRequest defines a request for the approval of the stage in the update.
+// Each snapshot MUST have the following labels:
+//   - `ParentStagedUpdateRun` which points to its owner stage update.
+//   - `TargetStage` which is the name of the stage that this approval request is for.
+//   - `IsLatestUpdateRunApproval` which indicates whether this approval request is the latest one related to this update run.
+type StageApprovalRequest struct {
```
jwtty, line 372: A little bit confused: how will this request get approved?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,441 @@
...
+	// It is the same as the apply strategy in the CRP when we first start the staged update run.
+	// We will NOT update the apply strategy during the update run even if the apply strategy changes in the CRP.
+	// +kubebuilder:validation:Required
+	ApplyStrategy v1beta1.ApplyStrategy `json:"appliedStrategy"`
+
+	// ResourceOverrideSnapshots is a list of ResourceOverride snapshots associated with the selected resources.
+	// The list computed as the beginning of the update run and not updated during the update run.
+	// The list is empty if there are no resource overrides associated with the selected resources.
+	// +kubebuilder:validation:Optional
+	ResourceOverrideSnapshots []v1beta1.NamespacedName `json:"resourceOverrideSnapshots,omitempty"`
```
zhiying-lin, line 165: overrides snapshot per cluster will be more friendly to the user.

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: user needs to patch the "Approved" condition of the request as true

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: The recommended in k8s is to use condition to record information for each state (running, wait, completed). I think it's better to at least keep the overall state in condition instead of a single status. We can convert all the state back into conditions too but it's little bit too much.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -346,6 +368,12 @@ const (
 	// - "False" means the staged update run is not waiting between stages.
 	// - "Unknown" means it is unknown.
 	StagedUpdateRunConditionWaitingBetweenStage StagedUpdateRunConditionType = "WaitingBetweenStage"
```
zhiying-lin, line 370: do we need the "Running" and "WaitingBetweenStage" conditions? they could be "reason" when the completed is "false".

"Running", "WaitingBetweenStage" and "Completed" all these conditions are delivering some duplicate information.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,378 @@
...
+	// The time to wait after all the clusters in the current stage complete the update before we move to the next stage.
+	// +kubebuilder:default:="1h"
+	// +kubebuilder:validation:Pattern="^0|([0-9]+(\\.[0-9]+)?(s|m|h))+$"
+	// +kubebuilder:validation:Type:=string
+	// +optional
+	WaitTime metav1.Duration `json:"waitTime,omitempty"`
+}
+
+// StagedUpdateRunStatus defines the observed state of the StagedUpdateRun.
+type StagedUpdateRunStatus struct {
```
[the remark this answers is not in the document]
zhiying-lin: not resolved?

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: I am not sure what other type of strategy we will add, I think we can just add new fields in the StagedUpdateStrategy

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: With "running" and "waittingBetweenStage" conditions, we will get the information like last time the stage has completed. If we collapse them into one condition then we lost a bit of the history.

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: removed this field

apis/placement/v1alpha1/stagedupdate_types.go
circy9: Do we really need both conditions and state? State serves the same purpose as conditions. State contains starttime and endtime.

RunStatus {
  runState State
  stages []StageStatus
  deletionStage StageStatus
}

StageStatus {
  stageName string
  stageState State
  clusters []ClusterStatus
  afterStageTasks []TaskStatus
}

ClusterStatus {
  clusterName string
  clusterState State
}

TastkStatus {
  taskType 
  taskState State
}

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -37,21 +38,21 @@ type StagedUpdateRun struct {
...
+// It specifies a stage-by-stage update process across selected clusters for the given ResourcePlacement object.
 type StagedUpdateRunSpec struct {
 	// A reference to the placement that this update run is applied to.
 	// There can be multiple active update runs for each placement but
 	// it's up to the devOps to make sure they don't conflict with each other.
 	// +kubebuilder:validation:Required
 	PlacementRef PlacementReference `json:"placementRef"`
 
-	// The resource snapshot index that this update run is based on.
+	// The snapshot index of the selected resources to be updated across clusters.
```
circy9, line 50: Could you give a concrete example on what you mean by the snaptshot index?

Something like:

e.g., for resource snapshots "xxx-123-1" and "xxx-123-2", "123" is the index.

circy9: I noticed that you used policysnaptshot name instead of index below. Is that because there are N resource snapshot for a single version of selected resources?

ryanzhang-oss: yes, for each resource snapshot index, there are N resource snapshots. The index is "123" in the example you gave. There is only 1 policy snapshot object for a snapshot of the CRP placement policy.

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: It's [k8s convention](https://github.com/zecke/Kubernetes/blob/master/docs/devel/api-conventions.md#typical-status-properties) in the status to use condition to show the state transition which is pretty much equivalent to the runState. I think using state is more efficient and acceptable as an internal field but we better to keep condition as the first layer fields status. If we really want to unify them then they all should be conditions like in CPR.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+	// the stage that removes all the resources from the clusters that are not selected by the
+	// current policy after all the update stages are completed.
+	// +kubebuilder:validation:Optional
+	DeletionStageStatus StageStatus `json:"deletionStageStatus,omitempty"`
+
+	// CurrentUpdatingStage is the name of the stage that is in the middle of the update.
+	// If the update run is not started, the value is an empty string.
+	// If we are in the DeletionStage, the name is called "deletionStage".
+	// +kubebuilder:validation:Optional
+	CurrentUpdatingStage string `json:"currentUpdatingStage,omitempty"`
```
circy9, line 190: Can we remove this as this information is already part of stagesstatus?

What about current cluster (that's in stagesstatus too)?

apis/placement/v1alpha1/stagedupdate_types.go
circy9: How about this? We use conditions at the updaterun level, but use updatestatus for stages and below.

Better to have a start time and end time for these AfterStageTasks too. So it is better to use UpdateStatus instead conditions for AfterStageTasks.

ryanzhang-oss: The condition states (ApprovalRequestCreated, ApprovalRequestApproved, WaitTimeElapsed) in the afterStageTasks are different from the update status so it's hard to reuse it.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+	// +kubebuilder:validation:Optional
+	ApplyStrategy v1beta1.ApplyStrategy `json:"appliedStrategy"`
+
+	// StagedUpdateStrategySnapshot is the snapshot of the StagedUpdateStrategy that we are going to use for the update run.
+	// The snapshot is immutable during the update run.
+	// We will apply the strategy to the the list of clusters scheduled by the CRP according to the current policy.
+	// The update run will fail to initialize if the strategy fails to produce a valid list of stages in which each selected
+	// cluster is included in exactly one stage.
+	// +kubebuilder:validation:Optional
+	StagedUpdateStrategySnapshot StagedUpdateStrategySpec `json:"StagedUpdateStrategySnapshot"`
```
zhiying-lin, line 173:
```suggestion
	StagedUpdateStrategySnapshot StagedUpdateStrategySpec `json:"stagedUpdateStrategySnapshot,omitempty"`
```

apis/placement/v1alpha1/stagedupdate_types.go
zhiying-lin: agree with Liqian about the detailed stage information.  They are already part of the StageStatus. 

 For the conditions in the "StagedUpdateRunStatus", can we treat it as overall condition/status? so that user can query this condition to know about whether the updateRun has completed or not.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	//
+	// Conditions is an array of current observed conditions for StagedUpdateRun.
+	// Known conditions are "Initialized", "Running", "WaitingBetweenStage", "Completed".
+	// +kubebuilder:validation:Optional
+	Conditions []metav1.Condition `json:"conditions"`
```
zhiying-lin, line 200:
```suggestion
	Conditions []metav1.Condition `json:"conditions,omitempty"`
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+	Status UpdateStatus `json:"status"`
+
+	// The list of each cluster's update status in this stage.
+	// +kubebuilder:validation:Required
+	Clusters []ClusterUpdateStatus `json:"clusters"`
+
+	// The status of the post update tasks that are associated with current stage.
+	// +kubebuilder:validation:MaxItems=2
+	// +kubebuilder:validation:Optional
+	AfterStageTaskStatus []AfterStageTaskStatus `json:"currentPostStageTaskStatus,omitempty"`
```
zhiying-lin, line 220:
```suggestion
	AfterStageTaskStatus []AfterStageTaskStatus `json:"afterStageTaskStatus ,omitempty"`
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+type StageApprovalRequestStatus struct {
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	//
+	// Conditions is an array of current observed conditions for the specific type of post update task.
+	// Known conditions are "Approved", and "ApprovalAccepted".
+	// +kubebuilder:validation:Optional
+	Conditions []metav1.Condition `json:"conditions"`
```
zhiying-lin, line 447:
```suggestion
	Conditions []metav1.Condition `json:"conditions,omitempty"`
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern="[A-Za-z0-9]+$"
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// LabelSelector is a label query over all the joined member clusters. Clusters matching the query are selected
+	// for this stage. There cannot be overlapping clusters between stages when the stagedUpdateRun is created.
+	// If the label selector is absent, the stage includes all the selected clusters.
+	// +kubebuilder:validation:Optional
+	LabelSelector *metav1.LabelSelector `json:"labelSelector"`
```
zhiying-lin, line 119:
```suggestion
	LabelSelector *metav1.LabelSelector `json:"labelSelector,omitempty"`
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=type
+	//
+	// Conditions is an array of current observed conditions for the specific type of post update task.
+	// Known conditions are "ApprovalRequestCreated", "WaitTimeElapsed", and "ApprovalRequestApproved".
+	// +kubebuilder:validation:Optional
+	Conditions []metav1.Condition `json:"conditions"`
```
zhiying-lin, line 318:
```suggestion
	Conditions []metav1.Condition `json:"conditions,omitempty"`
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// The desired state of StageApprovalRequest.
+	// +kubebuilder:validation:Required
+	Spec StageApprovalRequestSpec `json:"spec"`
+
+	// The desired state of StageApprovalRequest.
+	// +kubebuilder:validation:Optional
+	Status StageApprovalRequestStatus `json:"status"`
```
zhiying-lin, line 422:
```suggestion
	Status StageApprovalRequestStatus `json:"status,omitempty"`
```

apis/placement/v1alpha1/stagedupdate_types.go
circy9: NotStarted -> (created approval request) -> waiting -> (approve request) -> completed
NotStarted -> (start wait time) -> waiting -> (wait time has passed) -> completed

UpdateRun is a tree of steps: a step can be a run, a stage, a cluster, a after-stage task.

It is better to have the same state model for each step so customers don't need to learn new state/condition model for each step.

That's why we should have the same state for the run and each after-stage task as well.

circy9: Fleet RP also have the wait time step and use the same state data model.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Namespaced",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// StageApprovalRequest defines a request for the approval of the stage in the update.
+// Each snapshot MUST have the following labels:
+//   - `ParentStagedUpdateRun` which points to its owner stage update.
+//   - `TargetStage` which is the name of the stage that this approval request is for.
+//   - `IsLatestUpdateRunApproval` which indicates whether this approval request is the latest one related to this update run.
+type StageApprovalRequest struct {
```
circy9, line 412: I would recommend to use ApprovalRequest instead of StageApprovalRequest.

Add TargetStep in the spec. A step can be a stage, a group or a cluster.

```
StepReference {
  RunName
  StageName
  (future) GroupName
  (future) ClusterName
}
```

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,481 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Namespaced",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// StagedUpdateRun defines a stage by stage update run that applies the selected resources by the
+// corresponding ClusterResourcePlacement to its selected clusters. We remove the resources from the clusters that are
+// unselected after all the stages explicitly defined in the updateStrategy complete.
+// Each StagedUpdateRun object corresponds to a single "release" of a certain version of the resources.
+// The release is abandoned if the StagedUpdateRun object is deleted or the scheduling decision (i.e., the selected clusters) changes.
+// The name of the StagedUpdateRun needs to conform to [RFC 1123](https://tools.ietf.org/html/rfc1123).
```
circy9, line 26: Can we make sure the names are shorter so that in the future if we need to have an approval name like <runname>-<stagename>-<group-name>-<cluster-name>, it is shorter than 253 characters. We can keep each name to be shorter than 32 (we can relax the limit in the future if needed).

ryanzhang-oss: rfc 1123 limit is 63 so one can concatenate 4 parts into the approval name.

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,478 @@
...
+	// Empty if the stage has not finished updating all the clusters.
+	// +kubebuilder:validation:MaxItems=2
+	// +kubebuilder:validation:Optional
+	AfterStageTaskStatus []AfterStageTaskStatus `json:"afterStageTaskStatus ,omitempty"`
+
+	// The time when the update started on the stage. Empty if the stage has not started updating.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Format=date-time
+	StartTime *metav1.Time `json:"startTime,omitempty"`
```
circy9, line 249: Can we add endtime?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,478 @@
...
+	// The list is computed at the beginning of the update run and not updated during the update run.
+	// The list is empty if there are no cluster overrides associated with the cluster.
+	// +kubebuilder:validation:Optional
+	ClusterResourceOverrideSnapshots []string `json:"clusterResourceOverrideSnapshots,omitempty"`
+
+	// The time when the update started on the cluster. Empty if the cluster has not started updating.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Format=date-time
+	StartTime *metav1.Time `json:"startTime,omitempty"`
```
circy9, line 303: Can we add endtime?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,478 @@
...
+	// +kubebuilder:validation:Enum=TimedWait;Approval
+	// +kubebuilder:validation:Required
+	Type AfterStageTaskType `json:"type"`
+
+	// The name of the approval request object that is created for this stage.
+	// Only valid if the AfterStageTaskType is Approval.
+	// +kubebuilder:validation:Optional
+	ApprovalRequestName string `json:"approvalRequestName,omitempty"`
+
+	// +patchMergeKey=type
```
circy9, line 339: Can we add startime and endtime?

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -0,0 +1,478 @@
...
+
+// +genclient
+// +genclient:namespaced
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope="Namespaced",categories={fleet,fleet-placement}
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// ApprovalRequest defines a request for the approval from the user.
+// Each snapshot MUST have the following labels:
```
circy9, line 404:
```suggestion
// It MUST have the following labels:
```

apis/placement/v1alpha1/stagedupdate_types.go
ryanzhang-oss: the end time is the lastTransitionTime of the "succeded" condition

apis/placement/v1alpha1/stagedupdate_types.go
```diff
@@ -296,19 +296,13 @@ type ClusterUpdatingStatus struct {
...
-	StartTime *metav1.Time `json:"startTime,omitempty"`
-
 	// +patchMergeKey=type
 	// +patchStrategy=merge
 	// +listType=map
 	// +listMapKey=type
 	//
 	// Conditions is an array of current observed conditions for clusters. Empty if the cluster has not started updating.
-	// Known condition is "Succeeded".
+	// Known conditions are "Started,"Succeeded".
```
circy9, line 305:
```suggestion
	// Known conditions are "Started", "Succeeded".
```

circy9 (dismissed):

circy9 (approved):

PR #903, opened 2024-08-13
feat: add take over and drift detection options in CRP API

ryanzhang-oss:

Fixes #

I have:

### DIFF
--- a/apis/placement/v1beta1/clusterresourceplacement_types.go
+++ b/apis/placement/v1beta1/clusterresourceplacement_types.go
@@ -420,22 +420,25 @@ type RolloutStrategy struct {
 	// +optional
 	RollingUpdate *RollingUpdateConfig `json:"rollingUpdate,omitempty"`
 
-	// ApplyStrategy describes how to resolve the conflict if the resource to be placed already exists in the target cluster
-	// and is owned by other appliers.
+	// ApplyStrategy describes when and how to apply the selected resources to the target cluster.
 	// +optional
 	ApplyStrategy *ApplyStrategy `json:"applyStrategy,omitempty"`
 }
 
-// ApplyStrategy describes how to resolve the conflict if the resource to be placed already exists in the target cluster
-// and whether it's allowed to be co-owned by other non-fleet appliers.
+// ApplyStrategy describes when and how to apply the selected resource to the target cluster.
 // Note: If multiple CRPs try to place the same resource with different apply strategy, the later ones will fail with the
 // reason ApplyConflictBetweenPlacements.
 type ApplyStrategy struct {
 	// Type defines the type of strategy to use. Default to ClientSideApply.
-	// Server-side apply is a safer choice. Read more about the differences between server-side apply and client-side
-	// apply: https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply.
+	// Server-side apply is more powerful and flexible than client-side apply.
+	// You SHOULD use server-side apply to safely resolve any potential drift between the
+	// original applied resource version and the current resource on the member cluster.
+	// Read more about the differences between server-side apply and client-side apply:
+	// https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply.
+	// You can also use ReportDiff to only report the difference between the resource on the member cluster
+	// and the resource to be applied from the hub on all the selected clusters.
 	// +kubebuilder:default=ClientSideApply
-	// +kubebuilder:validation:Enum=ClientSideApply;ServerSideApply
+	// +kubebuilder:validation:Enum=ClientSideApply;ServerSideApply;ReportDiff
 	// +optional
 	Type ApplyStrategyType `json:"type,omitempty"`
 
@@ -448,10 +451,15 @@ type ApplyStrategy struct {
 	// ServerSideApplyConfig defines the configuration for server side apply. It is honored only when type is ServerSideApply.
 	// +optional
 	ServerSideApplyConfig *ServerSideApplyConfig `json:"serverSideApplyConfig,omitempty"`
+
+	// TakeoverAction describes the action to take when we place the selected resources on the target cluster the first time.
+	// +kubebuilder:default=AlwaysApply
+	// +kubebuilder:validation:Enum=AlwaysApply;ApplyIfNoDiff
+	// +optional
+	TakeoverAction TakeOverActionType `json:"actionType,omitempty"`
 }
 
-// ApplyStrategyType describes the type of the strategy used to resolve the conflict if the resource to be placed already
-// exists in the target cluster and is owned by other appliers.
+// ApplyStrategyType describes the type of the strategy used to apply the resource to the target cluster.
 // +enum
 type ApplyStrategyType string
 
@@ -465,6 +473,10 @@ const (
 	// and the existing resource in the target cluster.
 	// Details: https://kubernetes.io/docs/reference/using-api/server-side-apply
 	ApplyStrategyTypeServerSideApply ApplyStrategyType = "ServerSideApply"
+
+	// ApplyStrategyTypeReportDiff will generate a report of the difference between
+	// the resource on the member cluster and to be placed resource snapshot from the hub periodically.
+	ApplyStrategyTypeReportDiff ApplyStrategyType = "ReportDiff"
 )
 
 // ServerSideApplyConfig defines the configuration for server side apply.
@@ -480,6 +492,21 @@ type ServerSideApplyConfig struct {
 	// +optional
 	ForceConflicts bool `json:"force"`
 }
+
+// TakeOverActionType describes the type of the action to take when we first apply the resources to the member cluster.
+// +enum
+type TakeOverActionType string
+
+const (
+	// TakeOverActionTypeApplyIfNoDiff will apply the yaml from hub cluster only if there is no difference
+	// between the current resource snapshot version on the hub cluster and the existing resources on the member cluster.
+	// Otherwise, we will report the difference.
+	TakeOverActionTypeApplyIfNoDiff TakeOverActionType = "ApplyIfNoDiff"
+
+	// TakeOverActionTypeAlwaysApply will always apply the resource to the member cluster regardless
+	// if there are differences between the resource on the hub cluster and the existing resources on the member cluster.
+	TakeOverActionTypeAlwaysApply TakeOverActionType = "AlwaysApply"
+)
 
 // +enum
 type RolloutStrategyType string
--- a/config/crd/bases/placement.kubernetes-fleet.io_clusterresourcebindings.yaml
+++ b/config/crd/bases/placement.kubernetes-fleet.io_clusterresourcebindings.yaml
@@ -428,6 +428,15 @@ spec:
                   ApplyStrategy describes how to resolve the conflict if the resource to be placed already exists in the target cluster
                   and is owned by other appliers.
                 properties:
+                  actionType:
+                    default: AlwaysApply
+                    description: TakeoverAction describes the action to take when
+                      we place the selected resources on the target cluster the first
+                      time.
+                    enum:
+                    - AlwaysApply
+                    - ApplyIfNoDiff
+                    type: string
                   allowCoOwnership:
                     description: |-
                       AllowCoOwnership defines whether to apply the resource if it already exists in the target cluster and is not
@@ -455,11 +464,17 @@ spec:
                     default: ClientSideApply
                     description: |-
                       Type defines the type of strategy to use. Default to ClientSideApply.
-                      Server-side apply is a safer choice. Read more about the differences between server-side apply and client-side
-                      apply: https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply.
+                      Server-side apply is more powerful and flexible than client-side apply.
+                      You SHOULD use server-side apply to safely resolve any potential drift between the
+                      original applied resource version and the current resource on the member cluster.
+                      Read more about the differences between server-side apply and client-side apply:
+                      https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply.
+                      You can also use ReportDiff to only report the difference between the resource on the member cluster
+                      and the resource to be applied from the hub on all the selected clusters.
                     enum:
                     - ClientSideApply
                     - ServerSideApply
+                    - ReportDiff
                     type: string
                 type: object
               clusterDecision:
--- a/config/crd/bases/placement.kubernetes-fleet.io_clusterresourceplacements.yaml
+++ b/config/crd/bases/placement.kubernetes-fleet.io_clusterresourceplacements.yaml
@@ -1753,10 +1753,18 @@ spec:
                   with new ones.
                 properties:
                   applyStrategy:
-                    description: |-
-                      ApplyStrategy describes how to resolve the conflict if the resource to be placed already exists in the target cluster
-                      and is owned by other appliers.
+                    description: ApplyStrategy describes when and how to apply the
+                      selected resources to the target cluster.
                     properties:
+                      actionType:
+                        default: AlwaysApply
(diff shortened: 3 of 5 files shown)


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -420,12 +420,81 @@ type RolloutStrategy struct {
 	// +optional
 	RollingUpdate *RollingUpdateConfig `json:"rollingUpdate,omitempty"`
 
-	// ApplyStrategy describes how to resolve the conflict if the resource to be placed already exists in the target cluster
-	// and is owned by other appliers.
+	// ConflictResolutionStrategy describes how to resolve conflicts if the resource to be placed already exists in the target cluster.
+	// +optional
+	ConflictResolutionStrategy *ConflictResolutionStrategy `json:"conflictResolutionStrategy,omitempty"`
```
zhiying-lin, line 425: seems that ConflictResolutionStrategy and ApplyStrategy have some duplicate functionalities.

They're all about how to apply when the resource exists in the target cluster.

how about combining this new field with applyStrategy? treat them as new applyStrategy type?

ryanzhang-oss: I actually found it hard to explain if they are combined. For example, they both have a type.  I think the main reason is exactly because previously the "applyStrategy" also touches some conflict resolution. Unfortunately, we can't change that anymore.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -450,6 +452,72 @@ type ApplyStrategy struct {
...
+	// +optional
+	DiffCalculateStrategy DiffCalculateStrategy `json:"diffCalculateStrategy,omitempty"`
+}
+
+// ConflictResolutionActionType describes the type of the action to resolve the conflict
+// if the resource to be placed already exists in the target cluster.
+// +enum
+type ConflictResolutionActionType string
+
+const (
```
michaelawyu, line 477: The combos are:

* ReportIfDiff + HubFieldsOnly: complain if the apply would cause changes in the object, ignore changes in fields that are not mentioned by hub cluster resource
* ReportIfDiff + AllFields: complain if any field becomes inconsistent with the hub cluster resource
* ApplyIfNoDiff + HubFieldsOnly: complain if the apply would cause changes in the object, ignore changes in fields that are not mentioned by hub cluster resource (**duplicate**?)
* ApplyIfNoDiff + AllFieldsOnly: complain if any field becomes inconsistent with the hub cluster resource (**duplicate**?)
* AlwaysApply: overwrite any change in the fields mentioned by hub cluster resource
* Recreate: re-create if there's any change in the resource (*does this require diff option considerations?*)

zhiying-lin: echo Chen's comment, 

what's the diff between ReportIfDiff and ApplyIfNoDiff?

They're same, if there is difference, we'll report the diff and stop applying?

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -450,6 +452,72 @@ type ApplyStrategy struct {
 	ServerSideApplyConfig *ServerSideApplyConfig `json:"serverSideApplyConfig,omitempty"`
 }
 
+// ConflictResolutionConfig describes the action used to resolve conflicts if the resource
+// to be placed already exists in the target cluster.
+type ConflictResolutionConfig struct {
```
zhiying-lin, line 457: ConflictResolutionConfig describes "when to apply" while ApplyStrategyType describes "how to apply".

"Recreate" should be treated as new "ApplyStrategyType"? it always delete the existing one before creating.

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: good point, changed the "ReportIfDiff" to "Report"

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: good point

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -450,6 +452,69 @@ type ApplyStrategy struct {
...
+
+// ConflictResolutionActionType describes the type of the action to resolve the conflict
+// if the resource to be placed already exists in the target cluster.
+// +enum
+type ConflictResolutionActionType string
+
+const (
+	// ConflictResolutionActionTypeReport will generate a report of the diff between
+	// the yaml from the hub and the resource on the member cluster based on the `diffCalculateStrategy`.
+	ConflictResolutionActionTypeReport ConflictResolutionActionType = "Report"
```
zhiying-lin, line 480: hi ryan, what's the diff between  "report" and "applyIfNoDiff"?

do you mean for the "report", even when there's no diff, we won't apply? like a dry-run?

Another question is when the user finds there's a diff and we stop applying, how can the user tells us to continue the apply?

ryanzhang-oss: yes, it's a "dryrun".  

For your second question, they need to change the "ConflictResolutionActionType" to be a different type

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -420,22 +420,24 @@ type RolloutStrategy struct {
 	// +optional
 	RollingUpdate *RollingUpdateConfig `json:"rollingUpdate,omitempty"`
 
-	// ApplyStrategy describes how to resolve the conflict if the resource to be placed already exists in the target cluster
-	// and is owned by other appliers.
+	// ApplyStrategy describes when and how to apply the hub resource manifest to the target cluster.
```
circy9, line 423: s/the hub resource manifest/the selected resources/

Don't use new terms.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -420,22 +420,24 @@ type RolloutStrategy struct {
...
 	// +optional
 	ApplyStrategy *ApplyStrategy `json:"applyStrategy,omitempty"`
 }
 
-// ApplyStrategy describes how to resolve the conflict if the resource to be placed already exists in the target cluster
-// and whether it's allowed to be co-owned by other non-fleet appliers.
+// ApplyStrategy describes when and how to apply the resource to the target cluster.
```
circy9, line 428: s/the resource/the selected resources/

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -420,22 +420,24 @@ type RolloutStrategy struct {
...
 // Note: If multiple CRPs try to place the same resource with different apply strategy, the later ones will fail with the
 // reason ApplyConflictBetweenPlacements.
 type ApplyStrategy struct {
+	// ConflictResolutionConfig describes what to do when the resource to be placed already exists in the target cluster.
+	// +optional
+	ConflictResolutionConfig *ConflictResolutionConfig `json:"conflictResolutionStrategy,omitempty"`
```
circy9, line 434: With ConflictResolutionConfig + ApplyStrategyType + AllowCoOwership, it is very hard to understand how to use the API.

I assume ApplyStrategyType is about how to resolve field-level conflicts, AllowCoOwership is related to resource-level conflicts (when the resource already exists). Then ConflictResolutionConfig is also about resource-level conflicts.

I suggest to rethink about all the fields as a whole and not try to patch the API by making it even more complicated to think about all the old/new fields in applystrategy and what their impact is.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -450,6 +452,69 @@ type ApplyStrategy struct {
...
+	ConflictResolutionActionTypeApplyIfNoDiff ConflictResolutionActionType = "ApplyIfNoDiff"
+
+	// ConflictResolutionActionTypeAlwaysApply will always apply the resource to the member cluster regardless
+	// if there are differences between the resource on the hub cluster and the existing resources on the member cluster.
+	ConflictResolutionActionTypeAlwaysApply ConflictResolutionActionType = "AlwaysApply"
+)
+
+// DiffCalculateStrategy describes how to calculate the diff between the resource on the
+// hub cluster and the existing resources on the member cluster.
+type DiffCalculateStrategy struct {
```
circy9, line 495: I would recommend to remove this field and just do two-way delta to stay simple. Unless later customers want to have both options.

circy9, line 495: Where are we going to report the diff?

ryanzhang-oss: in the apply condition's message

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: yes, we can organize those fields better if we start from scratch but the API is V1 now so we technically cannot remove/change existing fields.

ryanzhang-oss: the current thinking is that `ConflictResolutionConfig` describes WHAT to do when the resource already exists on the member cluster.   `ApplyStrategyType` describes HOW to apply if "apply" is the action. The `AllowCoOwnership` technically belongs to ConflictResolutionConfig but we can't move it anymore.

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: what is a "two-way" delta?

apis/placement/v1beta1/clusterresourceplacement_types.go
zhiying-lin: AllowCoOwnership is not related to the resolve conflict directly. It only indicates whether it allows co-own even when the object has no conflicts.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -420,22 +420,25 @@ type RolloutStrategy struct {
...
-// and whether it's allowed to be co-owned by other non-fleet appliers.
+// ApplyStrategy describes when and how to apply the selected resource to the target cluster.
 // Note: If multiple CRPs try to place the same resource with different apply strategy, the later ones will fail with the
 // reason ApplyConflictBetweenPlacements.
 type ApplyStrategy struct {
 	// Type defines the type of strategy to use. Default to ClientSideApply.
-	// Server-side apply is a safer choice. Read more about the differences between server-side apply and client-side
-	// apply: https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply.
+	// Server-side apply is more powerful and flexible than client-side apply.
+	// You SHOULD use server-side apply to safely resolve any potential drift between the
```
zhiying-lin, line 434: could you please define what's the drift? do you mean the drift between current resources in the member and resources to be applied on the hub?

Just curious why cx cares about the drift before the applied.

I thought they would like to know the drift between (current resource in the member + resources to be applied on the hub) vs desired.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -420,22 +420,25 @@ type RolloutStrategy struct {
...
+	// original applied resource version and the current resource on the member cluster.
+	// Read more about the differences between server-side apply and client-side apply:
+	// https://kubernetes.io/docs/reference/using-api/server-side-apply/#comparison-with-client-side-apply.
+	// You can also use ReportDiff to only report the difference between the resource on the member cluster
+	// and the resource to be applied from the hub on all the selected clusters.
 	// +kubebuilder:default=ClientSideApply
-	// +kubebuilder:validation:Enum=ClientSideApply;ServerSideApply
+	// +kubebuilder:validation:Enum=ClientSideApply;ServerSideApply;ReportDiff
```
michaelawyu, line 441: Hi! My understanding is that this option is for apply diff, right? i.e., diff is found if a) a change has been made on the member cluster end w/o Fleet's acknowledgement (we need to track hashes of last successful apply results); and b) further apply op would lead to object changes. Shall we make this explicit in the comment?

michaelawyu: And this brings out the question of whether we'd like to support full diff as prev. explored in the UX doc, i.e., we also factor in fields that are not managed by Fleet -> I'll check with Infosys just to be certain.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -481,6 +493,21 @@ type ServerSideApplyConfig struct {
...
 
+// TakeOverActionType describes the type of the action to take when we first apply the resources to the member cluster.
+// +enum
+type TakeOverActionType string
+
+const (
+	// TakeOverActionTypeApplyIfNoDiff will apply the yaml from hub cluster only if there is no difference
+	// between the current resource snapshot version on the hub cluster and the existing resources on the member cluster.
+	// Otherwise, we will report the difference.
+	TakeOverActionTypeApplyIfNoDiff TakeOverActionType = "ApplyIfNoDiff"
```
michaelawyu, line 504: And the diff variants apply here as well. Though I guess for this one it's less of an issue considering that using apply diff would have no side effect.

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: drift = diff between "previous version in the hub" and "current resource in the member"

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: diff = diff between "the current version on the member cluster" and "the resource on the member cluster after the apply"

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -448,10 +451,15 @@ type ApplyStrategy struct {
 	// ServerSideApplyConfig defines the configuration for server side apply. It is honored only when type is ServerSideApply.
 	// +optional
 	ServerSideApplyConfig *ServerSideApplyConfig `json:"serverSideApplyConfig,omitempty"`
+
+	// TakeoverAction describes the action to take when we place the selected resources on the target cluster the first time.
+	// +kubebuilder:default=AlwaysApply
+	// +kubebuilder:validation:Enum=AlwaysApply;ApplyIfNoDiff
+	// +optional
+	TakeoverAction TakeOverActionType `json:"actionType,omitempty"`
```
seth-capistron, line 459: Does TakeoverAction have no impact if ApplyStrategyType is set to ReportDiff?

michaelawyu (approved):
LGTM ;)

apis/placement/v1beta1/clusterresourceplacement_types.go
ryanzhang-oss: yes, the ApplyStrategyType determines the action

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -448,10 +451,15 @@ type ApplyStrategy struct {
 	// ServerSideApplyConfig defines the configuration for server side apply. It is honored only when type is ServerSideApply.
 	// +optional
 	ServerSideApplyConfig *ServerSideApplyConfig `json:"serverSideApplyConfig,omitempty"`
+
+	// TakeoverAction describes the action to take when we place the selected resources on the target cluster the first time.
```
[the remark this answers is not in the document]
ryanzhang-oss: the first time means there is no previous hub version so whatever is on the member is different from "hub". The recreated case is different, in that case, there is an original "hub" version so there might or might not be any "drift" even if the resource is recreated/updated.

michaelawyu:
A side note: will reach to the customer asking for this feature to confirm on their use cases

PR #905, opened 2024-08-15
feat: force delete member cluster

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
+++ b/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
@@ -169,7 +169,8 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
 		}
 		return runtime.Result{Requeue: true}, controller.NewUpdateIgnoreConflictError(r.updateMemberClusterStatus(ctx, mc))
 	}
-	if r.canForceDelete(mc, currentImc) {
+	// check to see if we can force delete member cluster.
+	if currentImc.Spec.State == clusterv1beta1.ClusterStateLeave && time.Since(mc.DeletionTimestamp.Time) >= r.ForceDeleteWaitTime {
 		klog.V(2).InfoS("Force delete the member cluster, by garbage collecting owned resources", "memberCluster", mcObjRef)
 		return runtime.Result{Requeue: true}, r.garbageCollect(ctx, mc)
 	}
@@ -183,18 +184,6 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
 	// once the imc is updated, the mc controller will reconcile again ,or we reconcile to force delete
 	// the member cluster after force delete wait time.
 	return runtime.Result{RequeueAfter: r.ForceDeleteWaitTime}, controller.NewUpdateIgnoreConflictError(r.updateMemberClusterStatus(ctx, mc))
-}
-
-func (r *Reconciler) canForceDelete(mc *clusterv1beta1.MemberCluster, imc *clusterv1beta1.InternalMemberCluster) bool {
-	// we only force delete if member agent cannot reconcile an InternalMemberCluster in Leave state
-	// before the force delete wait time.
-	if imc.Spec.State != clusterv1beta1.ClusterStateLeave {
-		return false
-	}
-	if time.Since(mc.DeletionTimestamp.Time) < r.ForceDeleteWaitTime {
-		return false
-	}
-	return true
 }
 
 func (r *Reconciler) getInternalMemberCluster(ctx context.Context, name string) (*clusterv1beta1.InternalMemberCluster, error) {
--- a/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
+++ b/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
@@ -187,16 +187,13 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
 }
 
 func (r *Reconciler) canForceDelete(mc *clusterv1beta1.MemberCluster, imc *clusterv1beta1.InternalMemberCluster) bool {
-	// we only force delete if member agents cannot reconcile an InternalMemberCluster in Leave state.
+	// we only force delete if member agents cannot reconcile an InternalMemberCluster in Leave state
+	// before the force delete wait time.
 	if imc.Spec.State != clusterv1beta1.ClusterStateLeave {
 		return false
 	}
-	for i := range mc.Status.AgentStatus {
-		lastReceivedHB := mc.Status.AgentStatus[i].LastReceivedHeartbeat
-		// we don't set the LastReceivedHeartbeat field in the AgentStatus for networking agents.
-		if !reflect.DeepEqual(lastReceivedHB, metav1.Time{}) && time.Since(lastReceivedHB.Time) < r.ForceDeleteWaitTime {
-			return false
-		}
+	if time.Since(mc.DeletionTimestamp.Time) < r.ForceDeleteWaitTime {
+		return false
 	}
 	return true
 }
--- a/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
+++ b/pkg/controllers/membercluster/v1beta1/membercluster_controller.go
@@ -169,7 +169,6 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
 		}
 		return runtime.Result{Requeue: true}, controller.NewUpdateIgnoreConflictError(r.updateMemberClusterStatus(ctx, mc))
 	}
-	// check to see if we can force delete the member cluster.
 	if r.canForceDelete(mc, currentImc) {
 		klog.V(2).InfoS("Force delete the member cluster, by garbage collecting owned resources", "memberCluster", mcObjRef)
 		return runtime.Result{Requeue: true}, r.garbageCollect(ctx, mc)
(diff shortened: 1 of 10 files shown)


pkg/controllers/membercluster/v1beta1/membercluster_controller_integration_test.go
```diff
@@ -842,4 +855,215 @@ var _ = Describe("Test MemberCluster Controller", func() {
...
+			Expect(k8sClient.Delete(ctx, &mc)).Should(Succeed())
+
+			By("trigger reconcile to initiate leave workflow")
+			result, err := r.Reconcile(ctx, ctrl.Request{
+				NamespacedName: memberClusterNamespacedName,
+			})
+			Expect(result).Should(Equal(ctrl.Result{RequeueAfter: r.ForceDeleteWaitTime}))
+			Expect(err).Should(Succeed())
+
+			By("wait until force delete wait time has crossed")
```
Arvindthiru, line 1044: Just a fancy way to sleep, please lmk if there is a better way to do this

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -169,15 +169,32 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
...
+	// once the imc is updated, the mc controller will reconcile again ,or we reconcile to force delete
+	// the member cluster after force delete wait time.
+	return runtime.Result{RequeueAfter: r.ForceDeleteWaitTime}, controller.NewUpdateIgnoreConflictError(r.updateMemberClusterStatus(ctx, mc))
+}
+
+func (r *Reconciler) canForceDelete(mc *clusterv1beta1.MemberCluster, imc *clusterv1beta1.InternalMemberCluster) bool {
+	// we only force delete if member agent cannot reconcile an InternalMemberCluster in Leave state
+	// before the force delete wait time.
+	if imc.Spec.State != clusterv1beta1.ClusterStateLeave {
+		return false
```
ryanzhang-oss, line 192: when can this happen? Isn't the IMC state set by this controller?

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -169,15 +169,32 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
...
+	}
+	if mc.DeletionTimestamp != nil && time.Since(mc.DeletionTimestamp.Time) < r.ForceDeleteWaitTime {
```
ryanzhang-oss, line 194: when will the "mc.deletionTimeStamp != nil"

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
```diff
@@ -169,15 +169,32 @@ func (r *Reconciler) handleDelete(ctx context.Context, mc *clusterv1beta1.Member
...
 	if err := r.leave(ctx, mc, currentImc); err != nil {
 		klog.ErrorS(err, "Failed to mark the imc as leave", "memberCluster", mcObjRef)
 		return runtime.Result{}, err
 	}
 	// update the mc status to track the leaving status while we wait for all the agents to leave.
-	// once the imc is updated, the mc controller will reconcile again.
-	return runtime.Result{}, controller.NewUpdateIgnoreConflictError(r.updateMemberClusterStatus(ctx, mc))
+	// once the imc is updated, the mc controller will reconcile again ,or we reconcile to force delete
+	// the member cluster after force delete wait time.
+	return runtime.Result{RequeueAfter: r.ForceDeleteWaitTime}, controller.NewUpdateIgnoreConflictError(r.updateMemberClusterStatus(ctx, mc))
```
ryanzhang-oss, line 185: why do we want to wait for so long?

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
Arvindthiru: We don't want to force delete before we set IMC's spec to leave

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
Arvindthiru: That's just a nil check just to safe since we are accessing a pointer but as mentioned this is in the leave which meaning mc.DeletionTimestamp is non nil

pkg/controllers/membercluster/v1beta1/membercluster_controller.go
Arvindthiru: We didn't have to wait at all before, since we expect member agent to update IMC,

- If that occurs the logic remains the same
- if not, we wait until force delete wait time before requeue so that on the next reconcile we would have crossed the force delete wait time threshold and garbage collect

pkg/controllers/membercluster/v1beta1/membercluster_controller_integration_test.go
```diff
@@ -842,4 +845,215 @@ var _ = Describe("Test MemberCluster Controller", func() {
 			Expect(ns.DeletionTimestamp != nil).Should(BeTrue())
 		})
 	})
+
+	Context("Test membercluster controller force delete with enabling networking agents", func() {
+		BeforeEach(func() {
```
ryanzhang-oss, line 850: most of them are very similiar to the other contexts, any way not to copy paste so many code?

pkg/controllers/membercluster/v1beta1/membercluster_controller_integration_test.go
```diff
@@ -842,4 +845,215 @@ var _ = Describe("Test MemberCluster Controller", func() {
...
+				Message: propertyProviderConditionMessage1,
+			})
+			meta.SetStatusCondition(&imc.Status.Conditions, metav1.Condition{
+				Type:    propertyProviderConditionType2,
+				Status:  propertyProviderConditionStatus2,
+				Reason:  propertyProviderConditionReason2,
+				Message: propertyProviderConditionMessage2,
+			})
+			Expect(k8sClient.Status().Update(ctx, &imc)).Should(Succeed())
+
```
ryanzhang-oss, line 962: are those needed?

pkg/controllers/membercluster/v1beta1/membercluster_controller_integration_test.go
```diff
@@ -842,4 +845,215 @@ var _ = Describe("Test MemberCluster Controller", func() {
...
+			result, err = r.Reconcile(ctx, ctrl.Request{
+				NamespacedName: memberClusterNamespacedName,
+			})
+			Expect(result).Should(Equal(ctrl.Result{Requeue: true}))
+			Expect(err).Should(Succeed())
+
+			// check the member cluster namespace is being deleted. There is no namespace controller, so it won't be removed
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: namespaceName}, &ns)).Should(Succeed())
+			Expect(ns.DeletionTimestamp != nil).Should(BeTrue())
+		})
```
ryanzhang-oss, line 1057: this means we. need e2e test to really verify that

pkg/controllers/membercluster/v1beta1/membercluster_controller_integration_test.go
ryanzhang-oss: why wait before the check? The expected behavior is the namespace should be deleting after the delete wait time passes so the wait should be after.

Arvindthiru: That can be done for E2Es but since it's an we need to manually trigger the reconcile to simulate the Requeue After the force delete wait time

pkg/controllers/membercluster/v1beta1/membercluster_controller_integration_test.go
ryanzhang-oss: we can't check this in without e2e, that's like check in a bug.

test/e2e/join_and_leave_test.go
```diff
@@ -114,3 +121,51 @@ var _ = Describe("Test member cluster join and leave flow", Ordered, Serial, fun
...
+		Eventually(func() error {
+			return updateMemberAgentDeploymentReplicas(1)
+		}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to simulate member agent coming back up")
+
+		createMemberCluster(memberCluster3WestProd.ClusterName, memberCluster3WestProd.PresentingServiceAccountInHubClusterName, labelsByClusterName[memberCluster3WestProd.ClusterName], annotationsByClusterName[memberCluster3WestProd.ClusterName])
+		checkIfMemberClusterHasJoined(memberCluster3WestProd)
+	})
+})
+
+func updateMemberAgentDeploymentReplicas(replicas int32) error {
```
ryanzhang-oss, line 163: this function name doesn't mention that it updated a certain cluster agent replica to 0. Better pass the cluster in.

test/e2e/setup.sh
```diff
@@ -121,6 +121,7 @@ helm install hub-agent ../../charts/hub-agent/ \
     --set logVerbosity=5 \
     --set enableWebhook=true \
     --set webhookClientConnectionType=service \
+    --set forceDeleteWaitTime="2m0s" \
```
ryanzhang-oss, line 124: nit: can this duration be shorter? The longer this is, the more wait the forcedelete e2e needs

ryanzhang-oss (approved):

test/e2e/join_and_leave_test.go
```diff
@@ -114,3 +121,51 @@ var _ = Describe("Test member cluster join and leave flow", Ordered, Serial, fun
...
 })
+
+var _ = Describe("Test member cluster force delete flow", Ordered, Serial, func() {
+	Context("Test cluster join and leave flow with member agent down and force delete member cluster", Ordered, Serial, func() {
+		It("Simulate the member agent going down in member cluster", func() {
+			Eventually(func() error {
+				return updateMemberAgentDeploymentReplicas(memberCluster3WestProdClient, 0)
+			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to simulate member agent going down")
+		})
+
```
ryanzhang-oss, line 132: it just occurred to me that we didn't verity that there are no more pods in the deployment

PR #906, opened 2024-08-16
fix: updates binding transition times

britaniar:

Fixes #

I have: implemented some fixes so that the binding transition times updates for available condition, as well as add a retry for when there is a conflict when updating the binding.

### DIFF
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -120,6 +120,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 
 	workUpdated := false
 	overrideSucceeded := false
+	// Reset the conditions and failed placements.
+	for i := condition.OverriddenCondition; i < condition.TotalCondition; i++ {
+		resourceBinding.RemoveCondition(string(i.ResourceBindingConditionType()))
+	}
+	resourceBinding.Status.FailedPlacements = nil
 	// list all the corresponding works
 	works, syncErr := r.listAllWorksAssociated(ctx, &resourceBinding)
 	if syncErr == nil {
@@ -152,8 +157,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 		if err := errors.Unwrap(syncErr); err != nil && len(err.Error()) > 2 {
 			errorMessage = errorMessage[len(err.Error())+2:]
 		}
-		// remove all the failedPlacement as it does not reflect the latest status
-		resourceBinding.Status.FailedPlacements = nil
 		if !overrideSucceeded {
 			resourceBinding.SetConditions(metav1.Condition{
 				Status:             metav1.ConditionFalse,
@@ -180,8 +183,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 			Message:            "All of the works are synchronized to the latest",
 		})
 		if workUpdated {
-			// revert the applied condition and failedPlacement if we made any changes to the work
-			resourceBinding.Status.FailedPlacements = nil
+			// revert the applied condition if we made any changes to the work
 			resourceBinding.SetConditions(metav1.Condition{
 				Status:             metav1.ConditionFalse,
 				Type:               string(fleetv1beta1.ResourceBindingApplied),
@@ -195,9 +197,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 	}
 
 	// update the resource binding status
-	if updateErr := r.Client.Status().Update(ctx, &resourceBinding); updateErr != nil {
-		klog.ErrorS(updateErr, "Failed to update the resourceBinding status", "resourceBinding", bindingRef)
-		return controllerruntime.Result{}, controller.NewUpdateIgnoreConflictError(updateErr)
+	if updateErr := r.updateBindingStatusWithRetry(ctx, &resourceBinding); updateErr != nil {
+		return controllerruntime.Result{}, updateErr
 	}
 	if errors.Is(syncErr, controller.ErrUserError) {
 		// Stop retry when the error is caused by user error
@@ -218,6 +219,50 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 	// requeue if we failed to sync the work
 	// If we update the works, their status will be changed and will be detected by the watch event.
 	return controllerruntime.Result{}, syncErr
+}
+
+// updateBindingStatusWIthRetry sends the update request to API server with retry.
+func (r *Reconciler) updateBindingStatusWithRetry(ctx context.Context, resourceBinding *fleetv1beta1.ClusterResourceBinding) error {
+	// Retry only for specific errors or conditions
+	err := r.Client.Status().Update(ctx, resourceBinding)
+	if err != nil {
+		klog.ErrorS(err, "Failed to update the resourceBinding status, will retry", "resourceBinding", klog.KObj(resourceBinding), "resourceBindingStatus", resourceBinding.Status)
+		errAfterRetries := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			var latestBinding fleetv1beta1.ClusterResourceBinding
+			if err := r.Client.Get(ctx, client.ObjectKeyFromObject(resourceBinding), &latestBinding); err != nil {
+				return err
+			}
+			// Work generator is the only controller that updates conditions excluding rollout started which is updated by rollout controller.
+			if rolloutCond := latestBinding.GetCondition(string(fleetv1beta1.ResourceBindingRolloutStarted)); rolloutCond != nil {
+				found := false
+				for i := range resourceBinding.Status.Conditions {
+					if resourceBinding.Status.Conditions[i].Type == rolloutCond.Type {
+						// Replace the existing condition
+						resourceBinding.Status.Conditions[i] = *rolloutCond
+						found = true
+						break
+					}
+				}
+				if !found {
+					// Prepend the new condition if it wasn't found
+					resourceBinding.Status.Conditions = append([]metav1.Condition{*rolloutCond}, resourceBinding.Status.Conditions...)
+				}
+			}
+
+			if err := r.Client.Status().Update(ctx, resourceBinding); err != nil {
+				klog.ErrorS(err, "Failed to update the resourceBinding status on retry", "resourceBinding", klog.KObj(resourceBinding), "resourceBindingStatus", resourceBinding.Status)
+				return err
+			}
+			klog.V(2).InfoS("Successfully updated the resourceBinding status", "resourceBinding", klog.KObj(resourceBinding), "resourceBindingStatus", resourceBinding.Status)
+			return nil
+		})
+		if errAfterRetries != nil {
+			klog.ErrorS(errAfterRetries, "Failed to update binding status after retries", "resourceBinding", klog.KObj(resourceBinding))
+			return errAfterRetries
+		}
+		return nil
+	}
+	return err
 }
 
 // handleDelete handle a deleting binding
--- a/pkg/controllers/workgenerator/controller_integration_test.go
+++ b/pkg/controllers/workgenerator/controller_integration_test.go
@@ -47,7 +47,8 @@ var (
 	validResourceOverrideSnapshot          placementv1alpha1.ResourceOverrideSnapshot
 	invalidClusterResourceOverrideSnapshot placementv1alpha1.ClusterResourceOverrideSnapshot
 
-	cmpConditionOption = cmp.Options{cmpopts.SortSlices(utils.LessFuncFailedResourcePlacements), utils.IgnoreConditionLTTAndMessageFields, cmpopts.EquateEmpty()}
+	cmpConditionOption        = cmp.Options{cmpopts.SortSlices(utils.LessFuncFailedResourcePlacements), utils.IgnoreConditionLTTAndMessageFields, cmpopts.EquateEmpty()}
+	cmpConditionOptionWithLTT = cmp.Options{cmpopts.SortSlices(utils.LessFuncFailedResourcePlacements), cmpopts.EquateEmpty()}
 
 	fakeFailedAppliedReason  = "fakeApplyFailureReason"
 	fakeFailedAppliedMessage = "fake apply failure message"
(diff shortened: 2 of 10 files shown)


pkg/utils/condition/condition.go
```diff
@@ -109,7 +109,8 @@ func EqualCondition(current, desired *metav1.Condition) bool {
 		current.Type == desired.Type &&
 		current.Status == desired.Status &&
 		current.Reason == desired.Reason &&
-		current.ObservedGeneration >= desired.ObservedGeneration
+		current.ObservedGeneration >= desired.ObservedGeneration &&
+		current.LastTransitionTime == desired.LastTransitionTime
```
ryanzhang-oss, line 113: we cannot change this as this is used by many other functions

pkg/controllers/workgenerator/controller.go
```diff
@@ -197,7 +204,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 	// update the resource binding status
 	if updateErr := r.Client.Status().Update(ctx, &resourceBinding); updateErr != nil {
 		klog.ErrorS(updateErr, "Failed to update the resourceBinding status", "resourceBinding", bindingRef)
-		return controllerruntime.Result{}, controller.NewUpdateIgnoreConflictError(updateErr)
+		return controllerruntime.Result{Requeue: true}, controller.NewUpdateIgnoreConflictError(updateErr)
```
ryanzhang-oss, line 207: why requeue when there is an error?

pkg/controllers/workgenerator/controller.go
```diff
@@ -618,6 +625,8 @@ func setBindingStatus(works map[string]*fleetv1beta1.Work, resourceBinding *flee
 	if appliedCond.Status == metav1.ConditionTrue {
 		availableCond = buildAllWorkAvailableCondition(works, resourceBinding)
 		resourceBinding.SetConditions(availableCond)
+	} else if appliedCond.Status == metav1.ConditionFalse {
+		resourceBinding.RemoveConditions(resourceBinding.Status.Conditions, string(fleetv1beta1.ResourceBindingAvailable))
```
ryanzhang-oss, line 629: why do we need this? We already set the availableCondition to false when the resource changes.
Also, this checks the appliedCondition but remove the AvailableCondition, is that by design?

pkg/controllers/workgenerator/controller.go
britaniar: A requeue is need when there is an error. ASAIK, the error is expected and we continue on. Without this requeue, I do not see the available condition updated.

pkg/controllers/workgenerator/controller.go
britaniar: When the work is updated, we change the applied condition to false with the message that the work is synchronizing. Without this, the condition will stay as that when the applied is false.

pkg/controllers/workgenerator/controller.go
britaniar: Tested it locally. Works without it.

pkg/controllers/workgenerator/controller.go
britaniar: Will update test cases for the integration test to follow instead.

test/manifests/test-resource.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: random-test-resource
   namespace: app
 spec:
-  foo: foo1
+  foo: foo134
```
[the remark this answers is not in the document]
britaniar: I accidentally merged it. I was changing it when I was testing.

test/e2e/setup_test.go
```diff
@@ -226,8 +227,8 @@ func TestMain(m *testing.M) {
 	if err := fleetnetworkingv1alpha1.AddToScheme(scheme); err != nil {
 		log.Fatalf("failed to add custom APIs (networking) to the runtime scheme: %v", err)
 	}
-	if err := placementv1alpha1.AddToScheme(scheme); err != nil {
-		log.Fatalf("failed to add custom APIs (placement) to the runtime scheme: %v", err)
+	if err := testv1alpha1.AddToScheme(scheme); err != nil {
```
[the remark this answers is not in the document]
britaniar: add the test resource to the scheme

pkg/controllers/workgenerator/controller.go
```diff
@@ -189,35 +223,47 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
-		// This error usually indicates that the resource snapshot is deleted since the rollout controller which fills
-		// the resource snapshot share the same informer cache with this controller. We don't need to retry in this case
-		// since the resource snapshot will not come back. We will get another event if the binding is pointing to a new resource.
-		// However, this error can happen when the resource snapshot exists during the IT test when the client that creates
-		// the resource snapshot is not the same as the controller client so that we need to retry in this case.
-		// This error can also happen if the user uses a customized rollout controller that does not share the same informer cache with this controller.
-		return controllerruntime.Result{Requeue: true}, nil
+		klog.ErrorS(updateErr, "Failed to update the resourceBinding status", "resourceBinding", resourceBinding.Status)
+		// Retry only for specific errors or conditions
+		errAfterRetries := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
ryanzhang-oss, line 242: I also realized that we can't blindly retry this as the binding spec may have changed. I think the safer way is to delete the available condition if the applied Condition is false. Worst case is that we might have to add a "lastUpdatedTime" field to our conditions.

pkg/controllers/workgenerator/controller.go
```diff
@@ -127,6 +127,40 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+		// the resource snapshot share the same informer cache with this controller. We don't need to retry in this case
+		// since the resource snapshot will not come back. We will get another event if the binding is pointing to a new resource.
+		// However, this error can happen when the resource snapshot exists during the IT test when the client that creates
+		// the resource snapshot is not the same as the controller client so that we need to retry in this case.
+		// This error can also happen if the user uses a customized rollout controller that does not share the same informer cache with this controller.
+		return controllerruntime.Result{Requeue: true}, nil
+	}
+	// requeue if we failed to sync the work
+	// If we update the works, their status will be changed and will be detected by the watch event.
+	return controllerruntime.Result{}, syncErr
```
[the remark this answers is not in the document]
britaniar: Correct, not new.

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: we should only update the sync/override/applied/available condition on the binding status which is safe.

pkg/controllers/workgenerator/controller.go
```diff
@@ -189,35 +223,47 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
-		return controllerruntime.Result{Requeue: true}, nil
+		klog.ErrorS(updateErr, "Failed to update the resourceBinding status", "resourceBinding", resourceBinding.Status)
+		// Retry only for specific errors or conditions
+		errAfterRetries := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			var latestBinding fleetv1beta1.ClusterResourceBinding
+			if err := r.Client.Get(ctx, client.ObjectKeyFromObject(&resourceBinding), &latestBinding); err != nil {
+				return err
+			}
+			resourceBinding = latestBinding
+			buildResourceBindingStatus(&resourceBinding, overrideSucceeded, workUpdated, syncErr)
```
ryanzhang-oss, line 248: better not to re-compute this again as it might not be idempotent.

test/e2e/enveloped_object_placement_test.go
```diff
@@ -143,7 +143,7 @@ var _ = Describe("placing wrapped resources using a CRP", func() {
 			for idx := range allMemberClusters {
 				memberCluster := allMemberClusters[idx]
 				workResourcesPlacedActual := checkEnvelopQuotaAndMutationWebhookPlacement(memberCluster)
-				Eventually(workResourcesPlacedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place work resources on member cluster %s", memberCluster.ClusterName)
+				Eventually(workResourcesPlacedActual, longEventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place work resources on member cluster %s", memberCluster.ClusterName)
```
ryanzhang-oss, line 146: why do we need this?

pkg/controllers/workgenerator/controller.go
```diff
@@ -618,6 +672,8 @@ func setBindingStatus(works map[string]*fleetv1beta1.Work, resourceBinding *flee
 	if appliedCond.Status == metav1.ConditionTrue {
 		availableCond = buildAllWorkAvailableCondition(works, resourceBinding)
 		resourceBinding.SetConditions(availableCond)
+		//} else {
+		//	resourceBinding.RemoveCondition(string(fleetv1beta1.ResourceBindingAvailable))
```
ryanzhang-oss: why is this commented out?

britaniar: That's an old version.

pkg/controllers/workgenerator/controller.go
```diff
@@ -180,7 +179,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 			Message:            "All of the works are synchronized to the latest",
```
zhiying-lin, line 179: to make the fix simple, we can remove all the conditions (not set equals unknown) to reset and then add conditions.

pkg/controllers/workgenerator/controller.go
```diff
@@ -180,23 +180,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+			//	ObservedGeneration: resourceBinding.Generation,
+			//})
+			//resourceBinding.SetConditions(metav1.Condition{
+			//	Status:             metav1.ConditionFalse,
+			//	Type:               string(fleetv1beta1.ResourceBindingAvailable),
+			//	Reason:             condition.WorkNeedSyncedReason,
+			//	Message:            "In the processing of synchronizing the work to the member cluster",
+			//	ObservedGeneration: resourceBinding.Generation,
+			//})
+			resourceBinding.RemoveCondition(string(fleetv1beta1.ResourceBindingApplied))
```
michaelawyu, line 199: Just a nit: we could set the status to unknown if necessary. It would satisfy both the semantics and our use case.

pkg/controllers/workgenerator/controller.go
```diff
@@ -220,6 +230,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 	return controllerruntime.Result{}, syncErr
 }
 
+func (r *Reconciler) updateBinding(ctx context.Context, resourceBinding *fleetv1beta1.ClusterResourceBinding) error {
+	if updateErr := r.Client.Status().Update(ctx, resourceBinding); updateErr != nil {
```
michaelawyu, line 234: Just a nit: we could just wrap the whole thing in the retry block I assume?

pkg/controllers/workgenerator/controller.go
```diff
@@ -220,6 +230,51 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+		klog.ErrorS(updateErr, "Failed to update the resourceBinding status", "resourceBinding", resourceBinding.Status)
+		// Retry only for specific errors or conditions
+		errAfterRetries := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
michaelawyu, line 237: Earlier we had a case where constant conflicts and consequent backoffs lead to significant latencies; if that's a concern in this PR as well, we might consider using patches instead (provided that this controller solely owns the responsibility for updating the conditions).

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: since the resource binding has changed and we haven't applied it yet, I think it's safe to say that the new resource applied/available is "false".  "Unknown" works too. Removing probably also works semantically.

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: the retry should speed up the reconcile (instead of giving up). I think patch works, however, this controller only owns the specific condition of the status so needs to be careful not to touch the other fields.

test/e2e/rollout_test.go
```diff
@@ -542,6 +547,133 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update custom resource")
+		})
+
+		It("should update CRP status as expected", func() {
+			crpStatusUpdatedActual := customizedCRPStatusUpdatedActual(crpName, wantSelectedResources, []string{memberCluster1EastProdName}, nil, "2", false)
+			Eventually(crpStatusUpdatedActual, longEventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
+		})
+
+		It("should update binding transition time", func() {
+			// List all bindings associated with the given CRP.
```
ryanzhang-oss, line 657: e2e should test UX instead of internal implementation details. It should see the resources are not placed before the "notReady" time instead

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -1308,6 +1309,95 @@ var _ = Describe("Test Work Generator Controller", func() {
 				}, timeout, interval).Should(BeEmpty(), fmt.Sprintf("binding(%s) mismatch (-want +got)", binding.Name))
 			})
 		})
+
+		Context("Should not touch/reset RolloutStarted condition when the binding is updated", func() {
```
ryanzhang-oss, line 1313: I am not sure if I follow the test logic here, when was the resource changed?

test/e2e/rollout_test.go
```diff
@@ -542,6 +548,231 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+			crpStatusUpdatedActual := customizedCRPStatusUpdatedActual(crpName, wantSelectedResources, []string{memberCluster1EastProdName}, nil, "0", false)
+			Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
+		})
+
+		It("should place the resources on member cluster", func() {
+			workResourcesPlacedActual := waitForTestResourceToBePlaced(memberCluster1EastProd, &testCustomResource)
+			Eventually(workResourcesPlacedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place work resources on member cluster %s", memberCluster1EastProd.ClusterName)
+		})
+
+		It("should update custom resource", func() {
```
ryanzhang-oss, line 619:
```suggestion
		It("update the custom resource", func() {
```

test/e2e/rollout_test.go
```diff
@@ -542,6 +548,231 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+					return err
+				}
+				cr.Spec.Foo = "bar1" // Previously was "foo1"
+				return hubClient.Update(ctx, &cr)
+			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update custom resource")
+		})
+
+		It("should not update the resource on member cluster", func() {
+			// subtracting a second because transition between IT takes ~1 second
+			unavailablePeriod := time.Duration(*crp.Spec.Strategy.RollingUpdate.UnavailablePeriodSeconds)*time.Second - time.Second
```
ryanzhang-oss, line 633: better to leave a bit more buffer so the test case is not flaky, how about 5 seconds.

test/e2e/rollout_test.go
```diff
@@ -542,6 +548,231 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+					if diff := cmp.Diff(clusterRole, wantClusterRole, ignoreObjectMetaAutoGeneratedFields, ignoreObjectMetaAnnotationField); diff != "" {
+						return fmt.Errorf("clusterRole diff (-got, +want): %s", diff)
+					}
+					return nil
+				}
+				Eventually(resourcesPlacedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to place resources on member cluster %s", memberCluster.ClusterName)
+			}
+		})
+
+		It("should update custom resource", func() {
```
ryanzhang-oss, line 739:
```suggestion
		It("update the custom resource", func() {
```

test/e2e/rollout_test.go
```diff
@@ -542,6 +548,231 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+				}
+				cr.Rules[0].Verbs[0] = "list" // Previously was "get"
+				return hubClient.Update(ctx, cr)
+			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update ClusterRole")
+		})
+
+		It("should update the resource on member cluster", func() {
+			Consistently(func() error {
+				cr := &rbacv1.ClusterRole{}
+				err := hubClient.Get(ctx, types.NamespacedName{Name: clusterRoleName}, cr)
```
ryanzhang-oss, line 754: I don't get it, why check the resource on hub? Why is the resource supposed to change instantly on the member?

test/e2e/rollout_test.go
```diff
@@ -542,6 +548,231 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+			Eventually(crpStatusUpdatedActual, longEventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
+		})
+
+		AfterAll(func() {
+			// Remove the custom deletion blocker finalizer from the CRP.
+			ensureCRPAndRelatedResourcesDeletion(crpName, []*framework.Cluster{memberCluster1EastProd})
+		})
+	})
+
+	Context("Test a CRP place workload objects successfully, should not wait to place/update resource", Ordered, func() {
```
ryanzhang-oss, line 658: I am not sure what this case is testing

britaniar: Testing that there should not be a wait time since these will be ready and available as soon as they are created.

test/e2e/rollout_test.go
britaniar: I meant to check the resource on member cluster. I was thinking it would be changed pretty quickly since it is one of the resources that be ready and available as soon as they are created.

test/e2e/rollout_test.go
britaniar: I guess this test mostly tests the calculating the wait time more than this PR.

pkg/controllers/workgenerator/controller.go
```diff
@@ -220,6 +222,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+		if err := r.Client.Status().Update(ctx, resourceBinding); err != nil {
+			klog.ErrorS(err, "Failed to update the resourceBinding status on retry", "resourceBinding", klog.KObj(resourceBinding), "resourceBindingStatus", resourceBinding.Status)
+			return err
+		}
+		klog.V(2).InfoS("Successfully updated the resourceBinding status", "resourceBinding", klog.KObj(resourceBinding), "resourceBindingStatus", resourceBinding.Status)
+		return nil
+	})
+	if errAfterRetries != nil {
+		klog.ErrorS(errAfterRetries, "Failed to update binding status after retries", "resourceBinding", klog.KObj(resourceBinding))
+		return controller.NewUpdateIgnoreConflictError(errAfterRetries)
```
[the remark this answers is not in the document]
zhiying-lin: it could be, if it retires too many times, it will return the last error which could be conflict error.

pkg/controllers/workgenerator/controller.go
```diff
@@ -220,6 +221,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+func (r *Reconciler) updateBindingStatusWithRetry(ctx context.Context, resourceBinding *fleetv1beta1.ClusterResourceBinding) error {
+	// Retry only for specific errors or conditions
+	errAfterRetries := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var latestBinding fleetv1beta1.ClusterResourceBinding
+		if err := r.Client.Get(ctx, client.ObjectKeyFromObject(resourceBinding), &latestBinding); err != nil {
+			return err
+		}
+		// Work generator is the only controller that updates conditions excluding rollout started which is updated by rollout controller.
+		if rolloutCond := latestBinding.GetCondition(string(fleetv1beta1.ResourceBindingRolloutStarted)); rolloutCond != nil {
+			resourceBinding.SetConditions(*rolloutCond)
```
ryanzhang-oss, line 234: The setCondition call will change the lastTransitionTime if the status has changed which is not what we want. ie. if it was bound but now its unschedued, this will mark the unscheudeuled lastTransitionTime = Now() instead of when the rollout unselect it.  This has to be an exact copy.

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -1308,6 +1309,124 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+				verifyBindStatusAppliedNotAvailable(binding, false)
+				checkRolloutStartedNotUpdated(rolloutCond, binding)
+				// mark the work available
+				markWorkAvailable(&work)
+				// check the binding status that it should be marked as available true eventually
+				verifyBindStatusAvail(binding, false)
+				checkRolloutStartedNotUpdated(rolloutCond, binding)
+			})
+
+			It("Should treat the unscheduled binding as bound and not remove work", func() {
```
ryanzhang-oss, line 1401: this is exactly the case I mentioned above but the test can't hit the issue as the conflict condition is a race. The LTT of the rolling condition will change if we did a retry.

pkg/controllers/workgenerator/controller.go
```diff
@@ -220,6 +221,33 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 	return controllerruntime.Result{}, syncErr
 }
 
+// updateBindingStatusWIthRetry sends the update request to API server with retry.
+func (r *Reconciler) updateBindingStatusWithRetry(ctx context.Context, resourceBinding *fleetv1beta1.ClusterResourceBinding) error {
+	// Retry only for specific errors or conditions
+	errAfterRetries := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
ryanzhang-oss, line 227: This is an extra read without even trying to update. Ideally, we should update first and then retry on failure.

ryanzhang-oss, line 227:
```suggestion
	errAfterRetries := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```

test/e2e/rollout_test.go
```diff
@@ -542,6 +546,112 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+				err := hubClient.Get(ctx, types.NamespacedName{Name: testCustomResource.Name, Namespace: workNamespace.Name}, &cr)
+				if err != nil {
+					return err
+				}
+				cr.Spec.Foo = "bar1" // Previously was "foo1"
+				return hubClient.Update(ctx, &cr)
+			}, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update custom resource")
+		})
+
+		It("should not update the resource on member cluster", func() {
```
ryanzhang-oss, line 629:
```suggestion
		It("should not update the resource on member cluster before the unavailable second", func() {
```

test/e2e/rollout_test.go
```diff
@@ -542,6 +546,112 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+			Consistently(func() error {
+				var cr testv1alpha1.TestResource
+				err := memberCluster1EastProd.KubeClient.Get(ctx, types.NamespacedName{Name: testCustomResource.Name, Namespace: workNamespace.Name}, &cr)
+				if err != nil {
+					return err
+				}
+				if cr.Spec.Foo != "bar1" { // Previously was "foo1"
+					return nil
+				}
+				return err // Other errors should be returned
```
ryanzhang-oss, line 641: wouldn't this alwasy return nil, thus always pass?

pkg/controllers/workgenerator/controller.go
britaniar: Having everything wrapped inside was suggested [earlier](https://github.com/Azure/fleet/pull/906#discussion_r1733295068). Also wrapped in [here](https://github.com/Azure/fleet/blob/de6689a41f382fdbeae044cce9559c175c799d1b/pkg/controllers/workgenerator/controller.go#L275).

ryanzhang-oss: @michaelawyu any thoughts?

test/e2e/rollout_test.go
britaniar: you're right.

pkg/controllers/workgenerator/controller.go
britaniar: Changed it to your suggestion.

pkg/controllers/workgenerator/controller.go
```diff
@@ -220,6 +221,38 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
+	err := r.Client.Status().Update(ctx, resourceBinding)
+	if err != nil {
+		klog.ErrorS(err, "Failed to update the resourceBinding status, will retry", "resourceBinding", klog.KObj(resourceBinding), "resourceBindingStatus", resourceBinding.Status)
+		errAfterRetries := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			var latestBinding fleetv1beta1.ClusterResourceBinding
+			if err := r.Client.Get(ctx, client.ObjectKeyFromObject(resourceBinding), &latestBinding); err != nil {
+				return err
+			}
+			// Work generator is the only controller that updates conditions excluding rollout started which is updated by rollout controller.
+			if rolloutCond := latestBinding.GetCondition(string(fleetv1beta1.ResourceBindingRolloutStarted)); rolloutCond != nil && resourceBinding.GetCondition(string(fleetv1beta1.ResourceBindingRolloutStarted)) == nil {
```
ryanzhang-oss, line 236: what happens when the "resourceBinding.GetCondition(string(fleetv1beta1.ResourceBindingRolloutStarted)) != nil"

britaniar: It’s added twice

britaniar: During the first retry it doesn’t exist so it gets added. If it retries again it append another rollout started condition, but realized that it could be an updated rollout started condition when it retries.

ryanzhang-oss (approved):

PR #909, opened 2024-08-28
test: Fix bug

zhiying-lin:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/workgenerator/controller.go
+++ b/pkg/controllers/workgenerator/controller.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"go.uber.org/atomic"
 	"golang.org/x/sync/errgroup"
 	corev1 "k8s.io/api/core/v1"
@@ -435,7 +436,7 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding *fleetv1be
 		work := generateSnapshotWorkObj(workNamePrefix, resourceBinding, snapshot, simpleManifests)
 		activeWork[work.Name] = work
 		newWork = append(newWork, work)
-		klog.V(2).InfoS("Successfully generated the work objects for the resource snapshot", "resourceSnapshot", klog.KObj(snapshot), "work status", work.Status)
+		klog.V(2).InfoS("Successfully generated the work objects for the resource snapshot", "resourceSnapshot", klog.KObj(snapshot))
 
 		// issue all the create/update requests for the corresponding works for each snapshot in parallel
 		for ni := range newWork {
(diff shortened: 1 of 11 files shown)


pkg/controllers/workgenerator/controller.go
```diff
@@ -180,24 +183,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 			Message:            "All of the works are synchronized to the latest",
 		})
 		if workUpdated {
-			// revert the applied condition and failedPlacement if we made any changes to the work
-			resourceBinding.Status.FailedPlacements = nil
+			// revert the applied condition if we made any changes to the work
```
ryanzhang-oss, line 186: Why only revert the applied? Where are applied and available set in this case? It seems they are both not there.

pkg/controllers/workgenerator/controller.go
```diff
@@ -120,6 +120,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
 
 	workUpdated := false
 	overrideSucceeded := false
+	// Reset the conditions and failed placements.
+	for i := condition.OverriddenCondition; i < condition.TotalCondition; i++ {
+		resourceBinding.RemoveCondition(string(i.ResourceBindingConditionType()))
+	}
+	resourceBinding.Status.FailedPlacements = nil
```
ryanzhang-oss, line 127: I think the original design says the binding will always have all the conditions like CRP (unlike work). I am not sure if this will affect other component (i.e. CRP)

pkg/controllers/workgenerator/controller.go
zhiying-lin: we set the applied condition as false below. In that case, available condition nil means unknown.

In the CRP status, if the previous condition is not true, we will reset the remaining and leave them as "unknown".

pkg/controllers/workgenerator/controller.go
zhiying-lin: if i remember it correctly, the original design is to keep the existing conditions there and only update the new ones. It won't affect other component and they all use "condition.IsConditionStatusTrue(availableCondition, binding.GetGeneration())". Nil will return false to the caller. 


In the CRP code, we also reset all the existing conditions. https://github.com/Azure/fleet/blob/main/pkg/controllers/clusterresourceplacement/placement_status.go#L170

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: why set it to false then?  it seems that leaving it not set has the same effect?

ryanzhang-oss:
Thanks for the help, @britaniar is resuming her PR.

PR #911, opened 2024-08-29
interface: add APIs for eviction/disruption budget

michaelawyu:

This PR adds the APIs for supporting eviction and disruption budget.

I have:

Applied new resources on a kind cluster,

image
image

N/A


michaelawyu:
Internal only: please also see the mini UX doc for further details about this API (ping me if a link hasn't been sent).

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,149 @@
...
+// +kubebuilder:storageversion
+
+// ClusterResourcePlacementDisruptionBudget is the policy applied to a ClusterResourcePlacement
+// object that specifies its disruption budget, i.e., how many placements (clusters) can be
+// down at the same time due to **voluntary** disruptions (e.g., evictions).
+//
+// To apply a ClusterResourcePlacementDisruptionBudget to a ClusterResourcePlacement, use the
+// same name for the ClusterResourcePlacementDisruptionBudget object as the ClusterResourcePlacement
+// object. This guarantees a 1:1 link between the two objects.
+type ClusterResourcePlacementDisruptionBudget struct {
```
ryanzhang-oss, line 25: type PlacementDisruptionBudget struct {

How about

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,149 @@
...
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster,categories={fleet,fleet-placement},shortName=crpdb
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+
+// ClusterResourcePlacementDisruptionBudget is the policy applied to a ClusterResourcePlacement
+// object that specifies its disruption budget, i.e., how many placements (clusters) can be
+// down at the same time due to **voluntary** disruptions (e.g., evictions).
+//
+// To apply a ClusterResourcePlacementDisruptionBudget to a ClusterResourcePlacement, use the
```
ryanzhang-oss, line 22: is it possible to do a selector?

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,123 @@
...
+// create) the same Eviction object. Note also that an Eviction object will be
+// ignored once it is deemed invalid (e.g., such an object might be targeting a CRP object or
+// a placement that does not exist yet), even if it does become valid later
+// (e.g., the CRP object or the placement appears later). To fix the situation, re-create the
+// Eviction object.
+//
+// Executed evictions might be kept around for a while for auditing purposes; the Fleet controllers might
+// have a TTL set up for such objects and will garbage collect them automatically. For further
+// information, see the Fleet documentation.
+type ClusterResourcePlacementEviction struct {
```
ryanzhang-oss, line 42: type PlacementEviction struct {

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,123 @@
...
+	// +optional
+	Status ClusterResourcePlacementEvictionStatus `json:"status,omitempty"`
+}
+
+// ClusterResourcePlacementEvictionSpec is the desired state of the ClusterResourcePlacementEviction.
+type ClusterResourcePlacementEvictionSpec struct {
+	// ClusterResourcePlacementName is the name of the ClusterResourcePlacement object which
+	// the Eviction object targets at.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="The ClusterResourcePlacementName field is immutable"
```
ryanzhang-oss, line 62: can u point me to the code to handle the ":XValidation:rule" tag?

apis/placement/v1alpha1/disruptionbudget_types.go
michaelawyu: Yeah, will update the name.

apis/placement/v1alpha1/disruptionbudget_types.go
michaelawyu: Sure, I will update the definition 👌

apis/placement/v1alpha1/eviction_types.go
michaelawyu: Hi Ryan! This might be helpful: https://github.com/kubernetes-sigs/controller-tools/blob/46379c61e7abf72e4deda5e7c9d581e82a5c6140/pkg/crd/markers/validation.go#L329

apis/placement/v1alpha1/disruptionbudget_types.go
circy9: I actually like having DisruptionBudget as part of a CRP. It only has two fields. Reusing the two fields across CRPs -> fragmented experience for a user who only manages one CRP. Instead of having one CRP that contains disruptionbudget, now the user need to define two objects.

Making the two fields part of the CRP provides simpler experience instead of adding one more level of indirection.

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,123 @@
...
+	// +kubebuilder:validation:MaxLength=255
+	ClusterResourcePlacementName string `json:"clusterResourcePlacementName"`
```
circy9, line 64: PlacementReference?

apis/placement/v1alpha1/disruptionbudget_types.go
circy9: When you have DisruptionBudget with CRP selectors -> N CRPs, there will be M DisruptionBudget upon one CRP. Both implementation and readability (users need to think about which DisruptionBudget applies to a given CRP) become much more complicated.

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,123 @@
...
+// +kubebuilder:storageversion
+
+// ClusterResourcePlacementEviction is an eviction attempt on a specific placement from
+// a ClusterResourcePlacement object; one may use this API to force the removal of specific
+// resources from a cluster.
+//
+// An eviction is a **voluntary** disruption; its execution is subject to the disruption budget
+// linked with the target ClusterResourcePlacement object (if present).
+//
+// Beware that an eviction alone does not guarantee that a placement will not re-appear; i.e.,
```
circy9, line 24: Can we guarantee the evicted cluster won't be selected again? Our API is different from pod eviction now (we don't evict a binding), it is node eviction (we evict a cluster).

In which case it is useful to evict a workload from cluster A and then the scheduler selects A again?

apis/placement/v1alpha1/disruptionbudget_types.go
circy9: I think using a selector probably is needed for blue-green upgrades with two deployments for the same app: blue deployment, and green deployment. So you can say I want to have at least N instances running for blue+green.

However, in this API, we claimed that we have 1-1 link between CRP and PDB.

apis/placement/v1alpha1/eviction_types.go
michaelawyu: Fixed 👌

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,149 @@
...
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster,categories={fleet,fleet-placement},shortName=crpdb
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+
+// ClusterResourcePlacementDisruptionBudget is the policy applied to a ClusterResourcePlacement
+// object that specifies its disruption budget, i.e., how many placements (clusters) can be
+// down at the same time due to **voluntary** disruptions (e.g., evictions).
+//
```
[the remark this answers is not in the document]
michaelawyu: Yeah, we follow this as well. The budget does not protect against involuntary disruptions, but will count them. I will make this clear in the comment.

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,123 @@
...
+// create) the same Eviction object. Note also that an Eviction object will be
+// ignored once it is deemed invalid (e.g., such an object might be targeting a CRP object or
+// a placement that does not exist yet), even if it does become valid later
+// (e.g., the CRP object or the placement appears later). To fix the situation, re-create the
+// Eviction object.
+//
+// Executed evictions might be kept around for a while for auditing purposes; the Fleet controllers might
+// have a TTL set up for such objects and will garbage collect them automatically. For further
+// information, see the Fleet documentation.
+type PlacementEviction struct {
```
zhiying-lin, line 42: If the customer does not care about the pdb, do we allow or suggest them to delete the binding directly from the ux perspective?

apis/placement/v1alpha1/disruptionbudget_types.go
michaelawyu: Just to document a few pros and cons I've collected in the past days:

Inline:

* Pros: 1 DB : 1 CRP mapping that is easy to reason about + easy to track
* Cons: less flexible + too much stuff in the CRP API

Individual object:

* Pros: flexible (1 DB : N CRP mapping) + can have status for users to inspect current status (X disruptions allowed)
    * for K8s PDB, M : N mapping is not allowed; a 500 error will be returned if multiple PDBs references the same pod [link](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/)
* Cons: more difficult to reason about + more difficult to track

apis/placement/v1alpha1/eviction_types.go
Arvindthiru: Yes, the user should be able to use eviction alone

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,150 @@
...
+// PlacementDisruptionBudget.
+type PlacementDisruptionBudgetStatus struct {
+	// Conditions is the list of currently observed conditions for the
+	// PlacementDisruptionBudget object.
+	//
+	// Available condition types include:
+	// * Applied: whether the disruption budget has been successfully applied to the linked
+	//   ClusterResourcePlacement object.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
circy9, line 117: Do we need to provide the current status like?

total: the total number of target clusters of the placement.
currentAvailable: the number of currently available clusters.
desiredAvailable: the number of desired available clusters.
disruptionsAllowed: the number of clusters that are currently allowed to be disrupted.

Arvindthiru: After discussing with @michaelawyu offline, the PDB is just used for information purposes. 

When the eviction controller reconciles an eviction, it needs to list all PDBs associated with the CRP re-calculates voluntary disruptions allowed by PDBs and take a decision on the eviction.

Even if we update the PDB's status with the information suggested above there is a high chance the info. is stale.

The actual decision is always taken inside the eviction controller

michaelawyu: Yeah, just to re-cap a few things that Arvind and I discussed:

* The eviction controller will evaluate DBs at the time of execution;
* It's definitely helpful to have status that reports info. such as max allowed disruptions, just like K8s does with PDBs; those are for informational purposes only -> considering the timeline, this does not necessarily have to be done right away

circy9: If we have to calculate those numbers anyway, why not exposing them in the API?

It will be a puzzle for the user to understand why an eviction is rejected by PDB. Or will the eviction status contain such information?

circy9: If the eviction status will contain these numbers and which PDB blocked the eviction, it is ok to not report these numbers here.

michaelawyu: Understood. 👌 The eviction object status will include the reason why an eviction couldn't go through (e.g., `failed to execute as DB abc does not allow the op: max allowed disruptions: 2, current disruptions: 3`).

The concern we had earlier was mostly timeline-based: due to freshness constraints, the eviction controller must calculate the numbers at the time of execution, yet any status we report on the DB objects will be refreshed only periodically (the eviction controller couldn't rely on that information). The latter is not on the critical path, so just to be on the safer side for delivery it might be easier for us to focus on the main flow first. These pieces of information are obviously useful to users, of course ✨

circy9: SGTM.

circy9 (approved):

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,123 @@
...
+	Spec PlacementEvictionSpec `json:"spec"`
+
+	// Status is the observed state of the PlacementEviction.
+	// +optional
+	Status PlacementEvictionStatus `json:"status,omitempty"`
+}
+
+// PlacementEvictionSpec is the desired state of the PlacementEviction.
+type PlacementEvictionSpec struct {
+	// PlacementReference is the name of the ClusterResourcePlacement object which
```
circy9, line 59: Do we allow to N evictions pointing to the same placement? If not, please add a comment here and state which failures it may cause if there are N evictions for the same placement.

ryanzhang-oss (dismissed):

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,123 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster,categories={fleet,fleet-placement}
```
ryanzhang-oss, line 13: does it need to be cluster scoped?

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,150 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster,categories={fleet,fleet-placement},shortName=crpdb
```
ryanzhang-oss, line 14: does it need to be cluster scoped?

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,123 @@
...
+	// Available condition types include:
+	// * Valid: whether the Eviction object is valid, i.e., it targets at a valid placement.
+	// * Executed: whether the Eviction object has been executed.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// PlacementEvictionConditionType identifies a specific condition of the
+// PlacementEviction.
+type ClusterResourcePlacementConditionType string
```
ryanzhang-oss, line 87: The type should be PlacementEvictionConditionType

apis/placement/v1alpha1/disruptionbudget_types.go
Arvindthiru: Yes, since CRP is cluster-scoped

apis/placement/v1alpha1/eviction_types.go
Arvindthiru: Yes, since it's targeting a CRP and Placement i.e. CRB which are both cluster-scoped

apis/placement/v1alpha1/eviction_types.go
Arvindthiru: Yes, we can have N evictions pointing to the same placement. We evaluate each eviction as we reconcile them.

circy9: Why do we want to support a behavior different from the k8s experience?

https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors

"The eviction API will disallow eviction of any pod covered by multiple PDBs, so most users will want to avoid overlapping selectors. One reasonable use of overlapping PDBs is when pods are being transitioned from one PDB to another."

circy9: Oh, I think I asked the wrong question. I was trying to ask whether we allow N PDB against one CRP.

Arvindthiru: We don't allow N PDB against one CRP, the current API only applies a PDB if it has the same name as the CRP

Arvindthiru:
One concern: for CRPPDB spec validation I'm planning,

On adding a validating webhook configuration if user specified minAvailable and maxUnavailable
Unexpected error in the eviction controller (just in case user deletes the validating webhook configuration)

Should we also add a condition for CRPPDB to mention whether it's valid or invalid?

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Minimum desired number of available placements.
+	// +optional
+	DesiredAvailable int32 `json:"desiredAvailable"`
+
+	// Total number of placements counted by this disruption budget.
+	// +optional
+	TotalPlacements int32 `json:"totalPlacements"`
+
+	// TODO: Add a map to track ongoing evictions, to protect against concurrent evictions.
```
Arvindthiru, line 126: The PDB status will act as the concurrency control tool in the future, hence decided to add more fields now like

- DisruptionAllowed
- CurrentAvailable
- DesiredAvailable
- TotalPlacements

which will be filled before an eviction is executed

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,166 @@
...
+// To apply a ClusterResourcePlacementDisruptionBudget to a ClusterResourcePlacement, use the
+// same name for the ClusterResourcePlacementDisruptionBudget object as the ClusterResourcePlacement
+// object. This guarantees a 1:1 link between the two objects.
+type ClusterResourcePlacementDisruptionBudget struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the desired state of the ClusterResourcePlacementDisruptionBudget.
+	// +required
+	Spec ClusterResourcePlacementDisruptionBudgetSpec `json:"spec"`
```
ryanzhang-oss, line 32: please keep all secondary level object genric for reuse.

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,166 @@
...
+type PlacementDisruptionBudgetConditionType string
+
+const (
+	// PlacementDisruptionBudgetConditionTypeDisruptionAllowed indicates whether the disruption budget
+	// allows placements to be disrupted by voluntary disruptions.
+	//
+	// The following values are possible:
+	// * True: the disruption budget allows disruption for ClusterResourcePlacement.
+	// * False: the disruption budget does not allow any voluntary disruption for ClusterResourcePlacement.
+	PlacementDisruptionBudgetConditionTypeDisruptionAllowed PlacementDisruptionBudgetConditionType = "DisruptionAllowed"
```
ryanzhang-oss, line 148:
```suggestion
	PDBConditionDisruptionAllowed PlacementDisruptionBudgetConditionType = "DisruptionAllowed"
```

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -65,8 +66,6 @@ type PlacementDisruptionBudgetSpec struct {
 	// This field is mutually exclusive with the MinAvailable field in the spec; exactly one
 	// of them can be set at a time.
 	//
-	// Defaults to 25%.
```
Arvindthiru: Error returned when both MaxUnavailable and MinAvailable aren't specified,

The ClusterResourcePlacementDisruptionBudget "test-crpdb" is invalid: 
* spec: Required value
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,165 @@
...
+//
+// To apply a ClusterResourcePlacementDisruptionBudget to a ClusterResourcePlacement, use the
+// same name for the ClusterResourcePlacementDisruptionBudget object as the ClusterResourcePlacement
+// object. This guarantees a 1:1 link between the two objects.
+type ClusterResourcePlacementDisruptionBudget struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec is the desired state of the ClusterResourcePlacementDisruptionBudget.
+	// +kubebuilder:validation:XValidation:rule="!(has(self.maxUnavailable) && has(self.minAvailable))",message="Both MaxUnavailable and MinAvailable cannot be specified"
```
ryanzhang-oss, line 31: what if both of them are not set?

apis/placement/v1alpha1/disruptionbudget_types.go
ryanzhang-oss: if this is only used in the future, why not add them later?  Add it now means we need a PDB controller.

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,125 @@
...
+	Spec PlacementEvictionSpec `json:"spec"`
+
+	// Status is the observed state of the ClusterResourcePlacementEviction.
+	// +optional
+	Status PlacementEvictionStatus `json:"status,omitempty"`
+}
+
+// PlacementEvictionSpec is the desired state of the ClusterResourcePlacementEviction.
+type PlacementEvictionSpec struct {
+	// PlacementReference is the name of the ClusterResourcePlacement object which
```
ryanzhang-oss, line 59:
```suggestion
	// PlacementReference is the name of the Placement object which
```

apis/placement/v1alpha1/eviction_types.go
```diff
@@ -0,0 +1,125 @@
...
+	// Note that all fields in the spec are immutable.
+	// +required
+	Spec PlacementEvictionSpec `json:"spec"`
+
+	// Status is the observed state of the ClusterResourcePlacementEviction.
+	// +optional
+	Status PlacementEvictionStatus `json:"status,omitempty"`
+}
+
+// PlacementEvictionSpec is the desired state of the ClusterResourcePlacementEviction.
```
ryanzhang-oss, line 57:
```suggestion
// PlacementEvictionSpec is the desired state of the parent PlacementEviction.
```

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -0,0 +1,165 @@
...
+	// +required
+	Spec PlacementDisruptionBudgetSpec `json:"spec"`
+
+	// Status is the observed state of the ClusterResourcePlacementDisruptionBudget.
+	// +optional
+	Status PlacementDisruptionBudgetStatus `json:"status,omitempty"`
+}
+
+// PlacementDisruptionBudgetSpec is the desired state of the
+// ClusterResourcePlacementDisruptionBudget.
```
ryanzhang-oss, line 41:
```suggestion
// PlacementDisruptionBudget.
```

apis/placement/v1alpha1/disruptionbudget_types.go
Arvindthiru: Removed PDB status

ryanzhang-oss (approved):

Arvindthiru:
Upgrading kindest node image to use 1.30 throws this error when I tried to apply the new CRD,

Error from server (Forbidden): error when creating "./config/crd/bases/placement.kubernetes-fleet.io_clusterresourceplacementdisruptionbudgets.yaml": admission webhook "fleet.customresourcedefinition.guardrail.validating" denied the request: user: 'kubernetes-admin' in 'groups: [kubeadm:cluster-admins system:authenticated]' is not allowed to CREATE resource apiextensions.k8s.io/v1, Kind=CustomResourceDefinition/: /clusterresourceplacementdisruptionbudgets.placement.kubernetes-fleet.io

Will handle this update in a separate PR

ryanzhang-oss (approved):

PR #913, opened 2024-09-06
feat: Add drift information to the CRP status

ryanzhang-oss:

Add drift information to the CRP status

Fixes #

I have:


apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -657,53 +657,78 @@ const (
...
 
+// DriftedResourcePlacement contains the drifted resource details.
+type DriftedResourcePlacement struct {
+	// The resource that has drifted.
+	// +kubebuilder:validation:Required
+	ResourceIdentifier `json:",inline"`
+
+	// The ObservedGeneration of the resource on the target cluster that caused this drift.
+	// +kubebuilder:validation:Required
+	ObservedGeneration int64 `json:"observedGeneration"`
```
michaelawyu, line 717: Hi Ryan! Just a quick question: this assumes that we will be switching to our own condition implementations to relay this piece of information back, right?

michaelawyu: Also, considering that `observedGeneration` is kind of a well-known name in the K8s community, do we want to be more specific that it's about the generation on the target cluster? This is just a nit though, of course.

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -657,53 +657,78 @@ const (
...
+
+	// The details of difference between the resource on the target cluster and the resource to be applied from the hub.
+	// We will also emit an event with the difference details.
+	// +kubebuilder:validation:maxLength=8192
+	// +kubebuilder:validation:Optional
+	DriftDetail string `json:"driftDetail,omitempty"`
```
michaelawyu, line 723: TBH I am a bit worried about the size limit as well. Should we mention directly that we might be truncating this information as necessary and the user is expected to check on their own for large diffs?

apis/placement/v1beta1/clusterresourceplacement_types.go
michaelawyu: Just a random thought, but a `driftedSince` field could also be useful, just to pinpoint the cause

michaelawyu (commented):
Just a few nits; otherwise LGTM

apis/placement/v1beta1/clusterresourceplacement_types.go
```diff
@@ -657,53 +657,78 @@ const (
...
 	Condition metav1.Condition `json:"condition"`
 }
 
+// DriftedResourcePlacement contains the drifted resource details.
+type DriftedResourcePlacement struct {
+	// The resource that has drifted.
+	// +kubebuilder:validation:Required
+	ResourceIdentifier `json:",inline"`
+
+	// The ObservedGeneration of the resource on the target cluster that caused this drift.
```
circy9, line 715:
```suggestion
	// The ObservedGeneration of the resource on the target cluster that contains this drift.
```

apis/placement/v1beta1/clusterresourceplacement_types.go
circy9: Add observedAt to denote when the observation is made?

apis/placement/v1beta1/clusterresourceplacement_types.go
circy9, line 723: Use ObservedDifferences?

Drift is a noun to describe the action of A becoming more and more different from B.

apis/placement/v1beta1/clusterresourceplacement_types.go
circy9, line 717: ObservedGenerationInTargetCluster?

circy9 (approved):

apis/placement/v1beta1/clusterresourceplacement_types.go
circy9: done

PR #918, opened 2024-09-25
feat: add counter prometheus metric from fleet guard rail rejection

Arvindthiru:

Fixes #

I have:

image

### DIFF
--- a/pkg/webhook/validation/uservalidation.go
+++ b/pkg/webhook/validation/uservalidation.go
@@ -20,6 +20,7 @@ import (
 	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
 	fleetv1alpha1 "go.goms.io/fleet/apis/v1alpha1"
 	"go.goms.io/fleet/pkg/utils"
+	"go.goms.io/fleet/pkg/utils/controller/metrics"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/webhook/validation/uservalidation.go
```diff
@@ -57,13 +59,16 @@ func ValidateUserForFleetCRD(req admission.Request, whiteListedUsers []string, g
 }
 
 // ValidateUserForResource checks to see if user is allowed to modify argued resource modified by request.
-func ValidateUserForResource(req admission.Request, whiteListedUsers []string) admission.Response {
+func ValidateUserForResource(req admission.Request, whiteListedUsers []string, isTerminalCall bool) admission.Response {
```
Arvindthiru, line 62: We call this function here https://github.com/Azure/fleet/blob/main/pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go#L151 where it's we don't reject immediately we call another function to allow/reject hence added the isTerminalCall variable to enable counting the rejection correctly

Arvindthiru:
Looks like the API Server exposes the rejection count as a prometheus metric by default https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhook-metrics

image

PR #920, opened 2024-09-25
fix: remove the unexpected log

zhiying-lin:

This error occurs as the workgenerator controller caught a intermediate state where a resource binding does not have an expected condition -> should self-resolve after retries.

Remove the unexpected error.

Fixes #

I have:


pkg/controllers/workgenerator/controller.go
```diff
@@ -118,6 +118,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
 
+	// When the binding is in the unscheduled state, rollout controller won't update the condition anymore.
+	// We treat the unscheduled binding as bound until the rollout controller deletes the binding and here controller still
+	// updates the status for troubleshooting purpose.
+	// Requeue until the rollout controller finishes processing the binding.
+	if resourceBinding.Spec.State == fleetv1beta1.BindingStateBound {
+		rolloutStartedCondition := resourceBinding.GetCondition(string(fleetv1beta1.ResourceBindingRolloutStarted))
+		// Though the bounded binding is not taking the latest resourceSnapshot, we still needs to reconcile the works.
+		if !condition.IsConditionStatusFalse(rolloutStartedCondition, resourceBinding.Generation) &&
+			!condition.IsConditionStatusTrue(rolloutStartedCondition, resourceBinding.Generation) {
```
ryanzhang-oss, line 129: can we just check if the rollout condition exsit?

pkg/controllers/workgenerator/controller.go
```diff
@@ -234,18 +249,13 @@ func (r *Reconciler) updateBindingStatusWithRetry(ctx context.Context, resourceB
 			}
 			// Work generator is the only controller that updates conditions excluding rollout started which is updated by rollout controller.
 			if rolloutCond := latestBinding.GetCondition(string(fleetv1beta1.ResourceBindingRolloutStarted)); rolloutCond != nil {
-				found := false
```
ryanzhang-oss, line 237: now that we check the rolloutCondition before, this is now a real "unexpected" behavior then?

PR #921, opened 2024-09-26
fix: fix unexpected alert race condition 

ryanzhang-oss:

Fixes #

I have: add a check for rollout condition before the status is generated

### DIFF
--- a/pkg/controllers/workgenerator/controller_integration_test.go
+++ b/pkg/controllers/workgenerator/controller_integration_test.go
@@ -131,9 +131,7 @@ var _ = Describe("Test Work Generator Controller", func() {
 				ResourceSnapshotName: masterSnapshot.Name,
 				TargetCluster:        memberClusterName,
 			}
-			binding = generateClusterResourceBinding(spec)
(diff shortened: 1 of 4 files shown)


pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -146,6 +146,24 @@ var _ = Describe("Test Work Generator Controller", func() {
 			// flip the binding state to bound and check the work is created
 			binding.Spec.State = placementv1beta1.BindingStateBound
 			Expect(k8sClient.Update(ctx, binding)).Should(Succeed())
+			Eventually(func() error {
+				// Eventually update the binding with for RolloutStarted condition in case it runs into a conflict error
+				if err := k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding); err != nil {
+					return err
```
ryanzhang-oss, line 152: is there a way to not copy paste this code in every case? ie. put this in the create binging loigc.

michaelawyu (approved):
LGTM. I got a bit confused by the comments but it shouldn't be a blocker.

britaniar (approved):

PR #924, opened 2024-10-18
feat: add, remove finalizer for CRB in Scheduler and scheduler watcher for CRB

Arvindthiru:

Fixes #

I have:

- Adding scheduler clean up finalizers for CRBs when CRBs are created in a scheduling cycle
- Removing scheduler clean up finalizers for CRBs when CRBs are deleting with deleting timestamp
- Added CRB scheduler watcher to pick up deleting CRBs and enqueue CRP name to trigger scheduling cycle, also added watcher IT
- Added logic to clean up CRBs, by removing scheduler clean up finalizer when CRP is deleting

These changes are needed for the new PlacementEviction API

### DIFF
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -366,7 +366,9 @@ var markUnscheduledForAndUpdate = func(ctx context.Context, hubClient client.Cli
 	// unscheduledBinding from "scheduled" to "bound".
 	binding.Spec.State = placementv1beta1.BindingStateUnscheduled
 	err := hubClient.Update(ctx, binding, &client.UpdateOptions{})
-	klog.V(2).InfoS("Marking binding as unscheduled", "clusterResourceBinding", klog.KObj(binding), "error", err)
+	if err == nil {
+		klog.V(2).InfoS("Marked binding as unscheduled", "clusterResourceBinding", klog.KObj(binding))
+	}
 	return err
 }
 
@@ -374,10 +376,13 @@ var markUnscheduledForAndUpdate = func(ctx context.Context, hubClient client.Cli
 var removeFinalizerAndUpdate = func(ctx context.Context, hubClient client.Client, binding *placementv1beta1.ClusterResourceBinding) error {
 	controllerutil.RemoveFinalizer(binding, placementv1beta1.SchedulerCRBCleanupFinalizer)
 	err := hubClient.Update(ctx, binding, &client.UpdateOptions{})
-	klog.V(2).InfoS("Remove scheduler CRB cleanup finalizer", "clusterResourceBinding", klog.KObj(binding), "error", err)
+	if err == nil {
+		klog.V(2).InfoS("Removed scheduler CRB cleanup finalizer", "clusterResourceBinding", klog.KObj(binding))
+	}
 	return err
 }
 
+// updateBindings iterates over bindings and updates them using the update function provided.
 func (f *framework) updateBindings(ctx context.Context, bindings []*placementv1beta1.ClusterResourceBinding, updateFn func(ctx context.Context, client client.Client, binding *placementv1beta1.ClusterResourceBinding) error) error {
 	// issue all the update requests in parallel
 	errs, cctx := errgroup.WithContext(ctx)
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -296,13 +296,13 @@ func (f *framework) RunSchedulingCycleFor(ctx context.Context, crpName string, p
 	bound, scheduled, obsolete, unscheduled, dangling, deleting := classifyBindings(policy, bindings, clusters)
 
 	// Remove scheduler CRB cleanup finalizer on all deleting bindings.
-	if err := f.removeFinalizer(ctx, deleting); err != nil {
+	if err := f.updateBindings(ctx, deleting, removeFinalizerAndUpdate); err != nil {
 		klog.ErrorS(err, "Failed to remove finalizers from deleting bindings", "clusterSchedulingPolicySnapshot", policyRef)
 		return ctrl.Result{}, err
 	}
 
 	// Mark all dangling bindings as unscheduled.
-	if err := f.markAsUnscheduledFor(ctx, dangling); err != nil {
+	if err := f.updateBindings(ctx, dangling, markUnscheduledForAndUpdate); err != nil {
 		klog.ErrorS(err, "Failed to mark dangling bindings as unscheduled", "clusterSchedulingPolicySnapshot", policyRef)
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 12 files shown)


pkg/scheduler/framework/framework.go
```diff
@@ -373,7 +380,39 @@ func (f *framework) markAsUnscheduledFor(ctx context.Context, bindings []*placem
 					// We will just retry for conflict errors since the scheduler holds the truth here.
 					if apierrors.IsConflict(err) {
 						// get the binding again to make sure we have the latest version to update again.
-						return f.client.Get(cctx, client.ObjectKeyFromObject(unscheduledBinding), unscheduledBinding)
```
Arvindthiru, line 376: Just returning error from GET call eats up error from the update call above. I'm not sure how we have not encountered issues till now

pkg/scheduler/scheduler.go
```diff
@@ -282,25 +282,31 @@ func (s *Scheduler) cleanUpAllBindingsFor(ctx context.Context, crp *fleetv1beta1
...
+
 	// Note that once a CRP has been marked for deletion, it will no longer enter the scheduling cycle,
 	// so any cleanup finalizer has to be removed here.
 	//
 	// Also note that for deleted CRPs, derived bindings are deleted right away by the scheduler;
 	// the scheduler no longer marks them as deleting and waits for another controller to actually
 	// run the deletion.
 	for idx := range bindingList.Items {
-		binding := bindingList.Items[idx]
+		binding := &bindingList.Items[idx]
```
ryanzhang-oss, line 302: any reason to change here? Is it this is more performant?

pkg/scheduler/scheduler.go
```diff
@@ -282,25 +282,31 @@ func (s *Scheduler) cleanUpAllBindingsFor(ctx context.Context, crp *fleetv1beta1
...
-	//
+	// Remove scheduler CRB cleanup finalizer from deleting bindings.
+	for idx := range bindingList.Items {
+		binding := &bindingList.Items[idx]
+		controllerutil.RemoveFinalizer(binding, fleetv1beta1.SchedulerCRBCleanupFinalizer)
+		if err := s.client.Update(ctx, binding); err != nil {
+			klog.ErrorS(err, "Failed to remove scheduler reconcile finalizer from cluster resource binding", "clusterResourceBinding", klog.KObj(binding))
+			return controller.NewUpdateIgnoreConflictError(err)
+		}
+	}
```
ryanzhang-oss, line 293: why not combine this loop with the one below?

ryanzhang-oss: I am not sure if there are tests about this?

pkg/scheduler/scheduler.go
Arvindthiru: Not necessarily, wanted to ensure the update and delete we issued on the same binding using address

pkg/scheduler/scheduler.go
Arvindthiru: We have,
UT - https://github.com/Azure/fleet/pull/924/files#diff-130e36b28b6a84d355879caa6b7927c98ec18c63adad6ab4fc486458763080bfR56
IT - Where we delete CRPs and expect all CRBs to be deleted https://github.com/Azure/fleet/blob/main/test/scheduler/pickall_integration_test.go#L60

pkg/scheduler/framework/framework.go
```diff
@@ -349,31 +356,46 @@ func (f *framework) collectBindings(ctx context.Context, crpName string) ([]plac
...
+var markUnscheduledForAndUpdate = func(ctx context.Context, hubClient client.Client, binding *placementv1beta1.ClusterResourceBinding) error {
+	// Remember the previous unscheduledBinding state so that we might be able to revert this change if this
+	// cluster is being selected again before the resources are removed from it. Need to do a get and set if
+	// we add more annotations to the binding.
+	binding.SetAnnotations(map[string]string{placementv1beta1.PreviousBindingStateAnnotation: string(binding.Spec.State)})
+	// Mark the unscheduledBinding as unscheduled which can conflict with the rollout controller which also changes the state of a
+	// unscheduledBinding from "scheduled" to "bound".
+	binding.Spec.State = placementv1beta1.BindingStateUnscheduled
+	err := hubClient.Update(ctx, binding, &client.UpdateOptions{})
+	klog.V(2).InfoS("Marking binding as unscheduled", "clusterResourceBinding", klog.KObj(binding), "error", err)
```
ryanzhang-oss, line 369: why not separate the error from info? The log type is different

pkg/scheduler/framework/framework_test.go
```diff
@@ -373,10 +377,14 @@ func TestClassifyBindings(t *testing.T) {
...
+
+	if diff := cmp.Diff(deleting, wantDeleting); diff != "" {
+		t.Errorf("classifyBIndings() deleting diff (-got, +want) = %s", diff)
+	}
 }
 
-// TestMarkAsUnscheduledFor tests the markAsUnscheduledFor method.
-func TestMarkAsUnscheduledFor(t *testing.T) {
+// TestUpdateBindingsMarkAsUnscheduledForAndUpdate tests the updateBinding method by passing markUnscheduledForAndUpdate update function.
+func TestUpdateBindingsMarkAsUnscheduledForAndUpdate(t *testing.T) {
```
ryanzhang-oss, line 387: It seems that since we use a fake client, we can't test the bug you find?

pkg/scheduler/scheduler.go
ryanzhang-oss: is there any difference between this and getting the address inside the function call as we normally do, see below?
```
		if err := s.client.Update(ctx, &binding); err != nil {

	       if err := s.client.Delete(ctx, &binding); err != nil && !errors.IsNotFound(err) {
````

Comment (Arvindthiru) [reply pkg/scheduler/scheduler.go:294]: No difference, But as for performance, It looks like every time we use & in golang we create a new pointer.

File: pkg/scheduler/framework/framework.go [hunk 6]
```diff
@@ -366,18 +366,23 @@ var markUnscheduledForAndUpdate = func(ctx context.Context, hubClient client.Cli
 	// unscheduledBinding from "scheduled" to "bound".
 	binding.Spec.State = placementv1beta1.BindingStateUnscheduled
 	err := hubClient.Update(ctx, binding, &client.UpdateOptions{})
-	klog.V(2).InfoS("Marking binding as unscheduled", "clusterResourceBinding", klog.KObj(binding), "error", err)
+	if err == nil {
```
ryanzhang-oss, line 369: where is the err case going to be logged?

Arvindthiru: Where updateBinding is called,

- https://github.com/Azure/fleet/pull/924/files#diff-fc2ffba6d515e0ea81ed25218cc712a5949e0178a15fe3ebc6274d7ee27f401eR299
- https://github.com/Azure/fleet/pull/924/files#diff-fc2ffba6d515e0ea81ed25218cc712a5949e0178a15fe3ebc6274d7ee27f401eR305
- https://github.com/Azure/fleet/pull/924/files#diff-fc2ffba6d515e0ea81ed25218cc712a5949e0178a15fe3ebc6274d7ee27f401eR686
- https://github.com/Azure/fleet/pull/924/files#diff-fc2ffba6d515e0ea81ed25218cc712a5949e0178a15fe3ebc6274d7ee27f401eR839

ryanzhang-oss (approved):

PR #927, opened 2024-10-21
fix: add log for untolerated taint

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/scheduler/framework/plugins/tainttoleration/filtering.go
+++ b/pkg/scheduler/framework/plugins/tainttoleration/filtering.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
 
 	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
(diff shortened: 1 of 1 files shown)


jwtty (approved):

pkg/scheduler/framework/plugins/tainttoleration/filtering.go
```diff
@@ -26,6 +27,8 @@ func (p *Plugin) Filter(
 	if !isUntolerated {
 		return nil
 	}
+	policyRef := klog.KObj(policy)
```
ryanzhang-oss, line 30: this is not reused, why create this var?

pkg/scheduler/framework/plugins/tainttoleration/filtering.go
```diff
@@ -26,6 +27,8 @@ func (p *Plugin) Filter(
...
+	klog.V(2).InfoS("Cluster is unschedulable, because taint cannot be tolerated", "clusterSchedulingPolicySnapshot", policyRef, "taint", taint)
```
ryanzhang-oss, line 31: I looked at the other plugins, I don't see logs there, is there any reason? Should we log more in each plug-in?

PR #936, opened 2024-10-24
fix: handle resource snapshot missing but work already synced and add cro/ro annotation to all the works

ryanzhang-oss:
1. Handle the case that the resource snapshot is deleted when the work generator try to generate the work from the binding. We can safely continue if the work is up to date.
2. Add cluster override resource snapshots and resource override snapshots hash annotation to work so we can be sure in the future

Fixes #

I have:


pkg/controllers/workgenerator/controller.go
```diff
@@ -135,18 +136,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req controllerruntime.Reques
...
-	}
 	resourceBinding.Status.FailedPlacements = nil
 	// list all the corresponding works
 	works, syncErr := r.listAllWorksAssociated(ctx, &resourceBinding)
 	if syncErr == nil {
 		// generate and apply the workUpdated works if we have all the works
 		overrideSucceeded, workUpdated, syncErr = r.syncAllWork(ctx, &resourceBinding, works, cluster)
 	}
-
+	// Reset the conditions and failed placements.
```
michaelawyu, line 146: Hi Ryan! Just a nit: is line 142 (139) `resourceBinding.Status.FailedPlacements = nil` being left behind?

pkg/controllers/workgenerator/controller.go
```diff
@@ -485,6 +503,32 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding *fleetv1be
 	return true, updateAny.Load(), nil
 }
 
+// areAllWorkSynced checks if all the works are synced with the resource binding.
+func areAllWorkSynced(existingWorks map[string]*fleetv1beta1.Work, resourceBinding *fleetv1beta1.ClusterResourceBinding, _, _ string) bool {
+	syncedCondition := resourceBinding.GetCondition(string(fleetv1beta1.ResourceBindingWorkSynchronized))
```
michaelawyu, line 508: Hi Ryan! For clusters that have been in the faulted state this condition might have been removed; of course we have instructed the impacted user to fix things manually and they are the only affected party we knew, so it might be alright

michaelawyu (commented):
Added a few comments. LGTM ;)

michaelawyu (dismissed):
LGTM ;)

test/e2e/rollout_test.go
```diff
@@ -472,6 +473,162 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+				Version:   appv1.SchemeGroupVersion.Version,
+				Kind:      utils.DeploymentKind,
+				Name:      testDeployment.Name,
+				Namespace: testDeployment.Namespace,
+			}
+			crpStatusActual := safeRolloutWorkloadCRPStatusUpdatedActual(wantSelectedResources, failedDeploymentResourceIdentifier, allMemberClusterNames, "1", 2)
+			Eventually(crpStatusActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP status as expected")
+		})
+
+		It("update work to trigger a work generator reconcile", func() {
```
michaelawyu, line 551: Hi Ryan! Is this step necessary? I assume that after the image change the works should already have the false available conditions.

michaelawyu: At this moment, it seems that the removal of the resource snapshot name part (as in the test description) never happened

michaelawyu: And normally the work object should have already been marked as unavailable after image change.

test/e2e/rollout_test.go
```diff
@@ -472,6 +473,162 @@ var _ = Describe("placing wrapped resources using a CRP", Ordered, func() {
...
+			}
+		})
+
+		It("change the image name in deployment, to make it available again", func() {
+			Eventually(func() error {
+				err := hubClient.Get(ctx, types.NamespacedName{Name: testDeployment.Name, Namespace: testDeployment.Namespace}, &testDeployment)
+				if err != nil {
+					return err
+				}
+				testDeployment.Spec.Template.Spec.Containers[0].Image = "1.26.2"
```
michaelawyu, line 613: Ah, Ryan, I think this would be still be an invalid image name? But it does align with the situation Infosys encountered (which is, even if the old snapshot has been removed and all clusters are failing, updates should still be processed).

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -441,6 +454,59 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+					// mark the work applied which should trigger a reconcile loop and copy the status from the work to the binding
+					markWorkApplied(&work)
+					// check the binding status that it should be marked as applied true eventually
+					verifyBindStatusAppliedNotAvailable(binding, false)
+					Expect(k8sClient.Get(ctx, types.NamespacedName{Name: binding.Name}, binding)).Should(Succeed())
+					// update the resource snapshot to the next version that doesn't exist
+					binding.Spec.ResourceSnapshotName = "next"
+					Expect(k8sClient.Update(ctx, binding)).Should(Succeed())
+					updateRolloutStartedGeneration(&binding)
+					// check the binding status that it should be marked as override succeed but not synced
```
michaelawyu, line 487: Hi Ryan! Just a nit: here it means rollout started but not overridden, right?

test/e2e/rollout_test.go
ryanzhang-oss: if the work doesn't change, the already available binding won't get reconcile then its status will stay as "ready" then it won't hit this bug.

PR #942, opened 2024-10-31
feat: Eviction controller

Arvindthiru:

Fixes #

I have:

UTs and ITs

Changes to rollout controller needs to be present to add E2E tests


pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+	if executedCondition != nil {
+		klog.V(2).InfoS("Eviction has executed condition specified, no need to reconcile", "clusterResourcePlacementEviction", evictionName)
+		return runtime.Result{}, nil
+	}
+
+	isCRPPresent := true
+	var crp placementv1beta1.ClusterResourcePlacement
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: eviction.Spec.PlacementName}, &crp); err != nil {
+		if !errors.IsNotFound(err) {
+			return runtime.Result{}, err
```
ryanzhang-oss, line 81: controller.NewAPIServerError(

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+		return runtime.Result{}, err
+	}
+
+	var evictionTargetBinding *placementv1beta1.ClusterResourceBinding
+	for i := range crbList.Items {
+		if crbList.Items[i].Spec.TargetCluster == eviction.Spec.ClusterName {
+			if evictionTargetBinding == nil {
+				evictionTargetBinding = &crbList.Items[i]
+			} else {
+				klog.V(2).InfoS(evictionInvalidMultipleCRB, "clusterResourcePlacementEviction", evictionName, "clusterResourcePlacement", eviction.Spec.PlacementName)
```
ryanzhang-oss, line 102: how can this happen? We don't have two bindings point to the same cluster

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+		markEvictionInvalid(&eviction, evictionInvalidMissingCRB)
+		return runtime.Result{}, r.updateEvictionStatus(ctx, &eviction)
+	}
+
+	markEvictionValid(&eviction)
+	isDBPresent := true
+	var db placementv1alpha1.ClusterResourcePlacementDisruptionBudget
+	if err := r.Client.Get(ctx, types.NamespacedName{Name: crp.Name}, &db); err != nil {
+		if !errors.IsNotFound(err) {
+			return runtime.Result{}, err
```
ryanzhang-oss, line 119: controller.NewAPIServerError(

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+
+	totalBindings := len(crbList.Items)
+	allowed, disruptionsAllowed, availableBindings := isEvictionAllowed(desiredBindings, crbList.Items, db)
+	if allowed {
+		if err := r.deleteClusterResourceBinding(ctx, evictionTargetBinding); err != nil {
+			return runtime.Result{}, err
+		}
+		markEvictionExecuted(&eviction, fmt.Sprintf(evictionAllowedPDBSpecified, disruptionsAllowed, availableBindings, desiredBindings, totalBindings))
+	} else {
+		markEvictionNotExecuted(&eviction, fmt.Sprintf(evictionBlockedPDBSpecified, disruptionsAllowed, availableBindings, desiredBindings, totalBindings))
```
ryanzhang-oss, line 150: this will stop the eviction forever as per line 71?

ryanzhang-oss: disruptionsAllowed is always <= 0 when it is blocked, that number seems not useful

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+		desiredBindings = len(crp.Spec.Policy.ClusterNames)
+	}
+
+	totalBindings := len(crbList.Items)
+	allowed, disruptionsAllowed, availableBindings := isEvictionAllowed(desiredBindings, crbList.Items, db)
+	if allowed {
+		if err := r.deleteClusterResourceBinding(ctx, evictionTargetBinding); err != nil {
+			return runtime.Result{}, err
+		}
+		markEvictionExecuted(&eviction, fmt.Sprintf(evictionAllowedPDBSpecified, disruptionsAllowed, availableBindings, desiredBindings, totalBindings))
```
ryanzhang-oss, line 148: do user care "disruptionsAllowed" number?

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+			return runtime.Result{}, err
+		}
+		markEvictionExecuted(&eviction, evictionAllowedNoPDB)
+		return runtime.Result{}, r.updateEvictionStatus(ctx, &eviction)
+	}
+
+	var desiredBindings int
+	switch crp.Spec.Policy.PlacementType {
+	case placementv1beta1.PickAllPlacementType:
+		desiredBindings = len(crbList.Items)
```
ryanzhang-oss, line 135: The list includes bindings that are "noScheduled", it's not the accurate representation of the desired number of bindings

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+	"go.goms.io/fleet/pkg/utils/controller"
+)
+
+const (
+	reasonClusterResourcePlacementEvictionValid       = "ClusterResourcePlacementEvictionValid"
+	reasonClusterResourcePlacementEvictionInvalid     = "ClusterResourcePlacementEvictionInvalid"
+	reasonClusterResourcePlacementEvictionExecuted    = "ClusterResourcePlacementEvictionExecuted"
+	reasonClusterResourcePlacementEvictionNotExecuted = "ClusterResourcePlacementEvictionNotExecuted"
+
+	evictionInvalidMissingCRP  = "Failed to find cluster resource placement targeted by eviction"
```
michaelawyu, line 34: Hi Arvind! Just a nit: would you mind adding a `Message` suffix/prefix to the variables, just to make things a bit more clear?

michaelawyu: Also, binding are generally speaking considered to be our internal APIs, to keep things a bit more clear on the user end, it might make more sense to just use cluster references in the message part.

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+		Reason:             reasonClusterResourcePlacementEvictionNotExecuted,
+		Message:            message,
+	}
+	eviction.SetConditions(cond)
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
+	return runtime.NewControllerManagedBy(mgr).
+		WithOptions(ctrl.Options{MaxConcurrentReconciles: 1}). // set the max number of concurrent reconciles
```
michaelawyu, line 254: A nit: we might need to add a note here to explain why we have to use 1 max. concurrent reconciliations for now.

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+	if err := r.Client.Delete(ctx, binding); err != nil {
+		klog.ErrorS(err, "Failed to delete cluster resource binding", "clusterResourceBinding", bindingRef)
+		return controller.NewDeleteIgnoreNotFoundError(err)
+	}
+	klog.V(2).InfoS("Issued delete on cluster resource binding, eviction succeeded", "clusterResourceBinding", bindingRef)
+	return nil
+}
+
+// isEvictionAllowed calculates if eviction allowed based on available bindings and spec specified in placement disruption budget.
+func isEvictionAllowed(desiredBindings int, bindings []placementv1beta1.ClusterResourceBinding, db placementv1alpha1.ClusterResourcePlacementDisruptionBudget) (bool, int, int) {
```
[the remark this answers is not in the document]
Arvindthiru: I think we need unavailable binding policy similar to https://kubernetes.io/docs/tasks/run-application/configure-pdb/#unhealthy-pod-eviction-policy

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,257 @@
...
+	availableBindings := 0
+	for i := range bindings {
+		availableCondition := bindings[i].GetCondition(string(placementv1beta1.ResourceBindingAvailable))
+		if condition.IsConditionStatusTrue(availableCondition, bindings[i].GetGeneration()) {
+			availableBindings++
+		}
+	}
+	var disruptionsAllowed int
+	if db.Spec.MaxUnavailable != nil {
```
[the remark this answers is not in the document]
Arvindthiru: We have API validation to prevent this concern https://github.com/Azure/fleet/blob/main/apis/placement/v1alpha1/disruptionbudget_types.go#L30

But I do agree that there should only be one outcome though

pkg/controllers/clusterresourceplacementeviction/controller.go
Arvindthiru: This check was added to cover the case handled here https://github.com/Azure/fleet/blob/main/pkg/controllers/rollout/controller.go#L224

pkg/controllers/clusterresourceplacementeviction/controller.go
Arvindthiru: The following states are terminal states for an eviction, so we block reconcile.
- invalid eviction
- executed eviction
- not executed eviction

The user needs to retry by creating another eviction

pkg/controllers/clusterresourceplacementeviction/controller.go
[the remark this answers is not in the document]
Arvindthiru: Currently when an eviction is performed, CRB watcher watches the deletion and triggers a scheduling cycle. Then the scheduler adds deleting binding in any state to a deleting list and removes finalizers.

Due to this the scheduler will currently create another CRB for the same cluster, since the old binding is accounted for in the deleting binding list.

michaelawyu:
Arvind and I had a discussion about some of the finer points; just to scribble down some of the thoughts/points we went through here:

* On the issue about whether eviction on a binding that has been marked for deletion should be considered as invalid, since this is an issue with no substantial effect (i.e., the binding is already gone and there's nothing anyone can do with it), our discussion was mostly on the logical/UX front. Both ways should be fine; the only minor concern is that, if we mark such evictions as invalid, people would need to look at the reason/message part to realize that this is a situation without solutions; it appears like an error, but it has no fix. Besides, K8s does not block repeated delete calls either. For these reasons, I am wondering if it would be OK for us to mark it as valid but with a reason/message that suggests the already deleted observation; this would seem to be easier for users to comprehend.

* For the second concern about whether we need to account for target numbers (e.g., the N number of PickN CRPs) for DB's `maxUnavailable`/`minAvailable` calculations, I have to admit that this seems to be an issue where we could never have a perfect solution 😣; specifically:

    * For downscaling scenarios, accounting for target numbers would yield a lower bound for `maxUnavailable` and higher bound for `minAvailable`, which does offer better protection; however,
    * For upscaling and insufficient number of picked cluster (5 clusters wanted, 4 scheduled) scenarios, accounting for target numbers would yield higher bound for `maxUnavailable` and lower bound for `minAvailable`, which weakens the protection.
    * Of course we could choose to honor the most strict bound applicable, but this would be a system that's fairly difficult for users to comprehend.

    With these being said, I am wondering if it would make sense for us to start with the simple model, like K8s' PDB does, which concerns only the observed count of bindings; this has its own disadvantages (e.g., limited protection when downscaling), but it excels at its ease to understand (this might be esp. true considering that our APIs has been quite complicated). We could always build upon that (like K8s does with 1.31's `unhealthyEvictionPolicy`) if the users demand so. Another reason why I am a bit inclined towards this is that, if the user would like to have best protection, they could always switch to `minAvailable` w/ absolute count in the DB -> the dichotomy between `maxUnavailable` and `minAvailable` already offers users a choose between flexibility and protection, so it might be OK if we don't add another layer of consideration.

pkg/controllers/clusterresourceplacementeviction/controller_intergration_test.go
```diff
@@ -0,0 +1,1031 @@
...
+const (
+	eventuallyDuration   = time.Minute * 2
+	eventuallyInterval   = time.Millisecond * 250
+	consistentlyDuration = time.Second * 10
+	consistentlyInterval = time.Millisecond * 500
+)
+
+var _ = Describe("Test ClusterResourcePlacementEviction Controller", Ordered, func() {
+	Context("Invalid Eviction - ClusterResourcePlacement not found", func() {
+		crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
```
michaelawyu, line 53: This is just a general thought: because we run UTs and integration tests (of individual controllers) all together with the `go test` command, paralleled runs are never really enabled.

michaelawyu (commented):
Just a few minor comments; LGTM ;)

pkg/controllers/clusterresourceplacementeviction/controller_intergration_test.go
```diff
@@ -0,0 +1,1031 @@
...
+)
+
+const (
+	eventuallyDuration   = time.Minute * 2
+	eventuallyInterval   = time.Millisecond * 250
+	consistentlyDuration = time.Second * 10
+	consistentlyInterval = time.Millisecond * 500
+)
+
+var _ = Describe("Test ClusterResourcePlacementEviction Controller", Ordered, func() {
```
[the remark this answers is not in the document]
Arvindthiru: We don't need Ordered, ran the IT using,

ginkgo -procs=2 -v -race ./pkg/controllers/clusterresourceplacementeviction/...

To ensure it can run in parallel will move it out of the controller directory and move it test/integration in a follow up PR. Using BeforeEach & AfterEach instead

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,300 @@
...
+		if !errors.IsNotFound(err) {
+			return runtime.Result{}, controller.NewAPIServerError(true, err)
+		}
+		isCRPPresent = false
+	}
+	if !isCRPPresent {
+		klog.V(2).InfoS(evictionInvalidMissingCRPMessage, "clusterResourcePlacementEviction", evictionName, "clusterResourcePlacement", eviction.Spec.PlacementName)
+		markEvictionInvalid(&eviction, evictionInvalidMissingCRPMessage)
+		return runtime.Result{}, r.updateEvictionStatus(ctx, &eviction)
+	}
```
ryanzhang-oss, line 92:
```suggestion
	var crp placementv1beta1.ClusterResourcePlacement
	if err := r.Client.Get(ctx, types.NamespacedName{Name: eviction.Spec.PlacementName}, &crp); err != nil {
		if errors.IsNotFound(err) {
			klog.V(2).InfoS(evictionInvalidMissingCRPMessage, "clusterResourcePlacementEviction", evictionName, "clusterResourcePlacement", eviction.Spec.PlacementName)
		      markEvictionInvalid(&eviction, evictionInvalidMissingCRPMessage)
		     return runtime.Result{}, r.updateEvictionStatus(ctx, &eviction)
		}
         	return runtime.Result{}, controller.NewAPIServerError(true, err)	
	}
```

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,300 @@
...
+
+	var evictionTargetBinding *placementv1beta1.ClusterResourceBinding
+	for i := range crbList.Items {
+		if crbList.Items[i].Spec.TargetCluster == eviction.Spec.ClusterName {
+			if evictionTargetBinding == nil {
+				evictionTargetBinding = &crbList.Items[i]
+			} else {
+				klog.V(2).InfoS("Found more than one cluster resource binding for cluster targeted by eviction",
+					"clusterResourcePlacementEviction", evictionName, "clusterResourcePlacement", eviction.Spec.PlacementName)
+				markEvictionInvalid(&eviction, evictionInvalidMultipleCRBMessage)
```
ryanzhang-oss, line 107: what happens if one is deleting while the other is bound/scheduled?

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,300 @@
...
+	evictionBlockedPDBSpecifiedFmt = "Eviction is blocked by specified ClusterResourcePlacementDisruptionBudget, availablePlacements: %d, desiredPlacements: %d, totalPlacements: %d"
+)
+
+// Reconciler reconciles a ClusterResourcePlacementEviction object.
+type Reconciler struct {
+	client.Client
+}
+
+// Reconcile triggers a single eviction reconcile round.
+func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtime.Result, error) {
```
ryanzhang-oss, line 53: it's not a good pattern to put all the reconcile logic in one function.

pkg/controllers/clusterresourceplacementeviction/controller.go
ryanzhang-oss: What happens if the CRP is deleting?

ryanzhang-oss: this whole part can be inside a validating function

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,300 @@
...
+		isDBPresent = false
+	}
+
+	if !isDBPresent {
+		if err := r.deleteClusterResourceBinding(ctx, evictionTargetBinding); err != nil {
+			return runtime.Result{}, err
+		}
+		markEvictionExecuted(&eviction, evictionAllowedNoPDBMessage)
+		return runtime.Result{}, r.updateEvictionStatus(ctx, &eviction)
+	}
```
ryanzhang-oss, line 155: the isDBPresent variable is not needed

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,300 @@
...
+		disruptionsAllowed = availableBindings - minAvailable
+	}
+	if disruptionsAllowed < 0 {
+		disruptionsAllowed = 0
+	}
+	return disruptionsAllowed > 0, availableBindings
+}
+
+// calculateSchedulerTargetedBindings returns the count of bindings targeted by the scheduler.
+func calculateSchedulerTargetedBindings(bindings []placementv1beta1.ClusterResourceBinding) int {
```
ryanzhang-oss, line 230: This does not look right as we don't know what is the real target of a pickAll (https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors)

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+		markEvictionExecuted(&eviction, evictionAllowedNoPDBMessage)
+		return runtime.Result{}, r.updateEvictionStatus(ctx, &eviction)
+	}
+
+	bindingList := crbList.Items
+	var desiredBindings int
+	switch crp.Spec.Policy.PlacementType {
+	case placementv1beta1.PickAllPlacementType:
+		desiredBindings = calculateSchedulerTargetedBindings(bindingList)
```
ryanzhang-oss, line 161: Pick all is different (https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors)

pkg/controllers/clusterresourceplacementeviction/controller_test.go
```diff
@@ -0,0 +1,798 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "scheduled-binding",
+			Labels: map[string]string{placementv1beta1.CRPTrackingLabel: "test-crp"},
+		},
+		Spec: placementv1beta1.ResourceBindingSpec{
+			State:                        placementv1beta1.BindingStateScheduled,
+			ResourceSnapshotName:         "test-resource-snapshot",
+			SchedulingPolicySnapshotName: "test-scheduling-policy-snapshot",
+			TargetCluster:                "test-cluster-1",
+		},
```
ryanzhang-oss, line 35: do those fields matter in the test?

pkg/controllers/clusterresourceplacementeviction/controller_test.go
```diff
@@ -0,0 +1,798 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   "bound-available-binding",
+			Labels: map[string]string{placementv1beta1.CRPTrackingLabel: "test-crp"},
+		},
+		Spec: placementv1beta1.ResourceBindingSpec{
+			State:                        placementv1beta1.BindingStateBound,
+			ResourceSnapshotName:         "test-resource-snapshot",
+			SchedulingPolicySnapshotName: "test-scheduling-policy-snapshot",
+			TargetCluster:                "test-cluster-2",
+		},
```
ryanzhang-oss, line 47: same every where

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,303 @@
...
+		return controller.NewUpdateIgnoreConflictError(err)
+	}
+	klog.V(2).InfoS("Updated the status of a eviction", "clusterResourcePlacementEviction", evictionRef)
+	return nil
+}
+
+// deleteClusterResourceBinding deletes the specified cluster resource binding.
+func (r *Reconciler) deleteClusterResourceBinding(ctx context.Context, binding *placementv1beta1.ClusterResourceBinding) error {
+	bindingRef := klog.KObj(binding)
+	if err := r.Client.Delete(ctx, binding); err != nil {
```
michaelawyu, line 199: Hi Arvind! We need to add resource version check here to avoid dirty reads.

pkg/controllers/clusterresourceplacementeviction/controller.go
Arvindthiru: Marking eviction as invalid for now, the user can retry once the deleting eviction is removed

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,374 @@
...
+
+	executedCondition := eviction.GetCondition(string(placementv1alpha1.PlacementEvictionConditionTypeExecuted))
+	if executedCondition != nil {
+		klog.V(2).InfoS("Eviction has executed condition specified, no need to reconcile", "clusterResourcePlacementEviction", eviction.Name)
+		return true
+	}
+	return false
+}
+
+// isPlacementPresent checks to see if placement on target cluster is present.
```
ryanzhang-oss, line 248:
```suggestion
// isPlacementPresent checks to see if placement on target cluster could be present.
```

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,374 @@
...
+				return err
+			}
+			markEvictionExecuted(eviction, evictionAllowedNoPDBMessage)
+			return nil
+		}
+		return controller.NewAPIServerError(true, err)
+	}
+
+	// handle special case for PickAll CRP.
+	if crp.Spec.Policy.PlacementType == placementv1beta1.PickAllPlacementType {
```
ryanzhang-oss, line 212: need to update the API comment

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -0,0 +1,374 @@
...
+		Reason:             clusterResourcePlacementEvictionNotExecutedReason,
+		Message:            message,
+	}
+	eviction.SetConditions(cond)
+	klog.V(2).InfoS("Marked eviction as not executed", "clusterResourcePlacementEviction", klog.KObj(eviction))
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
+	// TODO(arvind): Increase max concurrent reconcile count, use map in PDB status to keep track of ongoing evictions.
```
ryanzhang-oss, line 362: we don't foresee this API being used heavily so I think 1 is fine for now

ryanzhang-oss (approved):

PR #951, opened 2024-11-11
feat: add ClusterStagedUpdateRun controller and handle deletion

jwtty:
Add the ClusterStagedUpdateRun controller and handle the CR deletion

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -13,7 +13,6 @@ import (
 
 	"go.goms.io/fleet/pkg/utils"
 	"go.goms.io/fleet/pkg/utils/controller"
-	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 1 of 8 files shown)


pkg/controllers/updaterun/controller.go
```diff
@@ -0,0 +1,158 @@
...
+	return runtime.Result{}, nil
+}
+
+// handleDelete handles the deletion of the clusterStagedUpdateRun object.
+// We need to wait for the update run to stop before deleting the clusterStagedUpdateRun object.
+// We will delete all the dependent resources, including approvalRequest objects, of the clusterStagedUpdateRun object.
+func (r *Reconciler) handleDelete(ctx context.Context, updateRun *placementv1alpha1.ClusterStagedUpdateRun) (bool, time.Duration, error) {
+	runObjRef := klog.KObj(updateRun)
+	// check if the update run is still running.
+	// the update run is considered as running if started condition is true while finished condition is not true or false.
```
ryanzhang-oss, line 88: why wait for the update run to complete? It can be stuck (with no approval or one application not working on one cluster)? How can we delete it then?

jwtty: removed this check.

pkg/controllers/updaterun/controller_integration_test.go
```diff
@@ -67,102 +65,21 @@ var _ = Describe("Test the clusterStagedUpdateRun controller", func() {
 	})
 
 	Context("Test deleting a clusterStagedUpdateRun", func() {
-		It("Should delete the clusterStagedUpdateRun if it's not started yet", func() {
+		It("Should delete the clusterStagedUpdateRun without any clusterApprovalRequests", func() {
```
ryanzhang-oss, line 68: can we add another case with cluserApprovalRequest (and not approved, thus stuck)?

jwtty: There is one with clusterApprovalRequests. But I'm not sure why deletion will get stuck if the requests are not approved?

jwtty: added more tests.

ryanzhang-oss (approved):

PR #956, opened 2024-11-13
fix: remove VnetName check from common cloudConfig package

britaniar:

Fixes #

I have: remove `VnetName` check from common package. This is so that the `VnetName` is not required on fleet-networking repo.

Will add check to another PR where I load the cloudConfig.

### DIFF
--- a/pkg/utils/cloudconfig/azure/config.go
+++ b/pkg/utils/cloudconfig/azure/config.go
@@ -88,10 +88,6 @@ func (cfg *CloudConfig) validate() error {
 		return fmt.Errorf("resource group is empty")
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/utils/cloudconfig/azure/config.go
```diff
@@ -88,7 +88,7 @@ func (cfg *CloudConfig) validate() error {
 		return fmt.Errorf("resource group is empty")
 	}
 
-	if cfg.VnetName == "" {
+	if cfg.UserAgent == "fleet-member-agent" && cfg.VnetName == "" {
```
zhiying-lin, line 91: Please move the vnetName check to where we new the config in the fleet repo.

As a common package, we should only validate the required the field for all the agents.
Any special check should not be hard-coded here.

PR #957, opened 2024-11-15
fix: allow aks-support user in fleet guard rail

Arvindthiru:

Fixes #

I have:


ryanzhang-oss (commented):
can we just say delete the guardrail during JIT session?

ryanzhang-oss (approved):

PR #960, opened 2024-11-16
feat: copy drift/diff details from work status to binding status

britaniar:

Fixes #

I have: 
   - copied work status drift/diff details to binding status
   - unit tests for new functions created


pkg/controllers/workgenerator/controller.go
```diff
@@ -1015,8 +1154,8 @@ func (r *Reconciler) SetupWithManager(mgr controllerruntime.Manager) error {
 				newAvailableCondition := meta.FindStatusCondition(newWork.Status.Conditions, fleetv1beta1.WorkConditionTypeAvailable)
 
 				// we try to filter out events, we only need to handle the updated event if the applied or available condition flip between true and false
-				// or the failed placements are changed.
-				if condition.EqualCondition(oldAppliedCondition, newAppliedCondition) && condition.EqualCondition(oldAvailableCondition, newAvailableCondition) {
+				// or the failed/diffed/drifted placements are changed.
+				if !comparePlacements(oldWork, newWork) && condition.EqualCondition(oldAppliedCondition, newAppliedCondition) && condition.EqualCondition(oldAvailableCondition, newAvailableCondition) {
```
michaelawyu, line 1158: Hi Britania! Glad that you caught the error in this flow; but the current setup kind of defeats the purpose of this check, which was meant to reduce the overhead for doing comparisons.

pkg/controllers/workgenerator/controller.go
```diff
@@ -1021,8 +1157,12 @@ func (r *Reconciler) SetupWithManager(mgr controllerruntime.Manager) error {
 						// we need to compare the failed placement if the work is not applied or available
```
michaelawyu, line 1157: Hi Britania! First of all, apologies again for the late response; I kind of lost track of the PR during last week's preparing for private preview. Totally my bad. Hopefully it's not blocking anything on your end. Below are some scenarios I've summarized:

michaelawyu: a) Failed placements (resources that cannot be applied or failed to get available) will only appear when either the `Applied` or `Available` conditions (on the work object) are set to `False`;

michaelawyu: b) Drifted placements can appear in any situation (`Applied` condition is `True` or `False`; though when it appears the condition is always set);

(a side note: the `Available` condition status does not concern this field)

michaelawyu: c) Diffed placements can only appear when the `Applied` condition is set to `False`.

(a side note: the `Available` condition status does not concern this field)

michaelawyu: So given this I think we might need to add another comparison in the `else` branch just in case, which would be the most straightforward approach to capture all changes (though maybe not in the most optimized manner). We could always add a TO-DO comment and re-evaluate it later.

michaelawyu (commented):
Added some comments, PTAL 🙏

pkg/controllers/workgenerator/controller.go
```diff
@@ -1015,14 +1154,23 @@ func (r *Reconciler) SetupWithManager(mgr controllerruntime.Manager) error {
...
 				if condition.EqualCondition(oldAppliedCondition, newAppliedCondition) && condition.EqualCondition(oldAvailableCondition, newAvailableCondition) {
+					oldDriftedPlacements := extractDriftedResourcePlacementsFromWork(oldWork)
+					newDriftedPlacements := extractDriftedResourcePlacementsFromWork(newWork)
+					driftsEqual := utils.IsDriftedResourcePlacementsEqual(oldDriftedPlacements, newDriftedPlacements)
 					if condition.IsConditionStatusFalse(newAppliedCondition, newWork.Generation) || condition.IsConditionStatusFalse(newAvailableCondition, newWork.Generation) {
+						diffsEqual := true
+						if condition.IsConditionStatusFalse(newAppliedCondition, newWork.Generation) {
+							oldDiffedPlacements := extractDiffedResourcePlacementsFromWork(oldWork)
+							newDiffedPlacements := extractDiffedResourcePlacementsFromWork(newWork)
+							diffsEqual = utils.IsDiffedResourcePlacementsEqual(oldDiffedPlacements, newDiffedPlacements)
```
ryanzhang-oss, line 1167: why not compare the two works directly?

pkg/utils/common.go
```diff
@@ -556,6 +556,92 @@ func IsFailedResourcePlacementsEqual(oldFailedResourcePlacements, newFailedResou
 	return true
 }
 
+// LessFuncDriftedResourcePlacements is a less function for sorting drifted resource placements
+var LessFuncDriftedResourcePlacements = func(a, b placementv1beta1.DriftedResourcePlacement) bool {
+	var aStr, bStr string
+	if a.Envelope != nil {
+		aStr = fmt.Sprintf(ResourceIdentifierWithEnvelopeIdentifierStringFormat, a.Group, a.Version, a.Kind, a.Namespace, a.Name, a.Envelope.Type, a.Envelope.Namespace, a.Envelope.Name)
```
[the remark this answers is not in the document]
ryanzhang-oss: is this function going to be shared by other components?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -1669,7 +1706,7 @@ func verifyBindStatusAppliedNotAvailable(binding *placementv1beta1.ClusterResour
 	}, timeout, interval).Should(BeEmpty(), fmt.Sprintf("binding(%s) mismatch (-want +got)", binding.Name))
 }
 
-func verifyBindStatusAvail(binding *placementv1beta1.ClusterResourceBinding, hasOverride bool) {
+func verifyBindStatusAvail(binding *placementv1beta1.ClusterResourceBinding, hasOverride, hasDriftPlacements bool) {
```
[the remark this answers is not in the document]
britaniar: Diff only happens if applied is false. Drift can happen if applied is false or true.

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -1177,7 +1190,147 @@ func TestSetBindingStatus(t *testing.T) {
 					},
 				},
 			},
-			want: []fleetv1beta1.FailedResourcePlacement{
+			wantFailedResourcePlacements: []fleetv1beta1.FailedResourcePlacement{
```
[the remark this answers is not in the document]
britaniar: No, it's just showing up weird on the PR. That's why line 1334 shows up as no change.

pkg/utils/common.go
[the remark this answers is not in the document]
britaniar: I put it here because the failed placements is also here (right above).  It can be used for e2e as well.

pkg/controllers/workgenerator/controller.go
britaniar: I discussed with Chen and we had decided to keep the same process it already had. Do you think it would be better to do that instead?

ryanzhang-oss: it is probably more efficient. I wonder why we wanted to keep it?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -426,33 +429,67 @@ var _ = Describe("Test Work Generator Controller", func() {
...
-					verifyBindStatusNotAvailableWithOneFailedPlacement(binding, false)
+					verifyBindStatusNotAvailableWithOnePlacement(binding, false, true)
 					// mark the work available directly
 					markWorkAvailable(&work)
 					// check the binding status that it should be marked as available true eventually
-					verifyBindStatusAvail(binding, false)
+					verifyBindStatusAvail(binding, false, false)
+				})
+
+				It("Should update the binding when the diffed placement list has changed but over all condition didn't change", func() {
```
ryanzhang-oss, line 461: I wonder which step tested the "diffed placement list"?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -426,33 +429,67 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+					markOneManifestAvailable(&work, false)
+					// check the binding status that it should be applied but not available and with one failed placement
+					verifyBindStatusNotAvailableWithOnePlacement(binding, false, true)
+					// mark the work available directly
+					markWorkAvailable(&work)
+					// check the binding status that it should be marked as available true eventually
+					verifyBindStatusAvail(binding, false, false)
+				})
+
+				It("Should update the binding when the drifted placement list has changed but over all condition didn't change", func() {
```
ryanzhang-oss, line 480: I wonder which step tested the "drifted placement list"?

pkg/controllers/workgenerator/controller.go
britaniar: For the sake of not changing the code too much. Will change to compare the two works directly.

pkg/controllers/workgenerator/controller_integration_test.go
britaniar: Anytime we verify the binding status. We make sure the binding status has the necessary and same diffed placement list as the work. In those functions, `hasDiffedPlacements` is a Boolean to add the diffed placements to the want status. Then we compare the binding status with the wanted binding status. The vice versa is the work no longer has diffed placements.

pkg/controllers/workgenerator/controller_integration_test.go
britaniar: Anytime we verify the binding status. The main purpose of this PR is to make sure that the placements list are copied over from the work to the binding. In this case, since drifted can happen when the `Applied` condition is true or false, I tested to make sure the placement list stays even when it fails and is available.

britaniar: Oh I see the it actually does not change the list. Updated test to have the list actually change.

pkg/controllers/workgenerator/controller_integration_test.go
britaniar: The actual step is when `markWorkAsAppliedButNotAvailableWithFailedManifest` once Applied condition is true the diffed placement list should be nil.

pkg/controllers/workgenerator/controller.go
```diff
@@ -1009,42 +1149,29 @@ func (r *Reconciler) SetupWithManager(mgr controllerruntime.Manager) error {
...
+					// There is an edge case that, the work spec is the same but from different resourceSnapshots.
+					// WorkGenerator will update the work because of the label changes, but the generation is the same.
+					// When the normal update happens, the controller will set the applied condition as false and wait
+					// until the work condition has been changed.
+					// In this edge case, we need to requeue the binding to update the binding status.
+					if oldResourceSnapshot == newResourceSnapshot {
+						klog.V(2).InfoS("The work applied or available condition stayed as true, no need to reconcile", "oldWork", klog.KObj(oldWork), "newWork", klog.KObj(newWork))
+						return
 					}
 				}
```
ryanzhang-oss, line 1173: I wonder if we still need this?

britaniar: Yes. I am only comparing the status, so it is still necessary to have these checks that we implemented recently.

ryanzhang-oss (approved):

PR #963, opened 2024-11-19
feat: disable placing atm related resources

zhiying-lin:

The fix is needed for supporting new networking feature

Fixes #

I have:

### DIFF
--- a/pkg/utils/apiresources.go
+++ b/pkg/utils/apiresources.go
@@ -40,6 +40,18 @@ var (
 		Kind:    "ServiceImport",
 	}
 
+	trafficManagerProfileGVK = schema.GroupVersionKind{
+		Group:   NetworkingGroupName,
+		Version: "v1alpha1",
+		Kind:    "TrafficManagerProfile",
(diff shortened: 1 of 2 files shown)


michaelawyu (approved):
LGTM ;)

pkg/utils/apiresources.go
```diff
@@ -81,7 +93,10 @@ func NewResourceConfig(isAllowList bool) *ResourceConfig {
 	r.AddGroup(metricsV1beta1.GroupName)
 	r.AddGroupVersionKind(corev1PodGVK)
 	r.AddGroupVersionKind(corev1NodeGVK)
+	// disable networking resources
```
ryanzhang-oss, line 96: should add the entire fleet-network group instead of one by one

zhiying-lin: it depends. for example, the serviceExport and mcs should be placed to the members.

Need to check if we support disabling all networking gvk but excluding the  serviceExport and mcs today.

PR #972, opened 2024-11-24
feat: Add delete override option in the API

ryanzhang-oss:

Fixes #

I have:


apis/placement/v1alpha1/override_types.go
```diff
@@ -56,6 +56,7 @@ type OverridePolicy struct {
 	// OverrideRules defines an array of override rules to be applied on the selected resources.
 	// The order of the rules determines the override order.
 	// When there are two rules selecting the same fields on the target cluster, the last one will win.
+	// There can only be one override rule if it is a `Delete` type override rule.
```
zhiying-lin, line 59: there could be multiple override rules.

For example,

i want to override serviceExport on member-1, and delete serviceExport on member-2.

apis/placement/v1alpha1/override_types.go
```diff
@@ -78,10 +85,21 @@ type OverrideRule struct {
...
+	// +optional
+	JSONPatchOverrides []JSONPatchOverride `json:"jsonPatchOverrides,omitempty"`
 }
 
+// OverrideType defines the type of Override
+type OverrideType string
+
+const (
+	// JSONPatchOverrideType applies a JSON patch on the selected resources following [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902).
+	JSONPatchOverrideType OverrideType = "JSONPatch"
```
zhiying-lin, line 97:
```suggestion
	OverrideTypeJSONPatch OverrideType = "JSONPatch"
```

nit, usually people use the type as the prefix instead of the suffix when name enum.

apis/placement/v1alpha1/override_types.go
```diff
@@ -78,10 +85,21 @@ type OverrideRule struct {
...
+
+	// DeleteOverrideType deletes the selected resources on the target clusters.
+	DeleteOverrideType OverrideType = "Delete"
```
zhiying-lin, line 100:
```suggestion
	OverrideTypeRemove OverrideType = "Remove"
```
niiiiiiiiit, how about using the same terminology as "JSONPatch"?

apis/placement/v1alpha1/override_types.go
```diff
@@ -66,6 +67,12 @@ type OverridePolicy struct {
 
 // OverrideRule defines how to override the selected resources on the target clusters.
 type OverrideRule struct {
+	// OverrideType defines the type of the override rules.
+	// +kubebuilder:validation:Enum=JSONPatch;Delete
+	// +kubebuilder:default:JSONPatch
+	// +optional
+	OverrideType OverrideType `json:"overrideType,omitempty"`
```
circy9, line 74: Better to move this to be after ClusterSelector.

Why?
- Specify the target and then specify the desired action upon the target.
- OverrideType and JSONPatchOverrrides are used to describe the action. They should be together.

apis/placement/v1alpha1/override_types.go
circy9: What is the reason to use remove instead of delete?

The list of verbs are:
GET, POST, PUT, PATCH, DELETE

zhiying-lin: i was looking at the JSONPatch operator (add, remove, replace)  and tried to keep them consistent.

looks like the HTTP verb is better here and we already have PATCH.

apis/placement/v1alpha1/override_types.go
```diff
@@ -74,14 +74,32 @@ type OverrideRule struct {
 	// +optional
 	ClusterSelector *placementv1beta1.ClusterSelector `json:"clusterSelector,omitempty"`
 
+	// OverrideType defines the type of the override rules.
+	// +kubebuilder:validation:Enum=JSONPatch;Delete
+	// +kubebuilder:default:JSONPatch
+	// +optional
+	OverrideType OverrideType `json:"overrideType,omitempty"`
```
zhiying-lin, line 81:
```suggestion
	Type OverrideType `json:"type,omitempty"`
```

apis/placement/v1alpha1/override_types.go
```diff
@@ -74,14 +74,32 @@ type OverrideRule struct {
...
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinItems=1
 	// +kubebuilder:validation:MaxItems=20
-	// +required
-	JSONPatchOverrides []JSONPatchOverride `json:"jsonPatchOverrides"`
+	// +optional
+	JSONPatchOverrides []JSONPatchOverride `json:"jsonPatchOverrides,omitempty"`
 }
 
+// OverrideType defines the type of Override
```
zhiying-lin, line 92: missing period

PR #975, opened 2024-11-26
feat: implement stagedUpdateRun initialization

jwtty:
Implement the initialization stage of the `clusterStagedUpdateRun` controller.

Fixes #

I have:


pkg/controllers/rollout/controller.go
```diff
@@ -125,7 +126,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	// fill out all the default values for CRP just in case the mutation webhook is not enabled.
 	defaulter.SetDefaultsClusterResourcePlacement(&crp)
 
-	matchedCRO, matchedRO, err := r.fetchAllMatchingOverridesForResourceSnapshot(ctx, crp.Name, latestResourceSnapshot)
+	matchedCRO, matchedRO, err := overrider.FetchAllMatchingOverridesForResourceSnapshot(ctx, r.Client, r.InformerManager, crp.Name, latestResourceSnapshot)
```
jwtty, line 129: In this function, always the latest CRO/ROs are used, even with an old `ResourceSnapshop`. Do we want this behavior?

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,433 @@
...
+		err := fmt.Errorf("no latest policy snapshot associated with the clusterResourcePlacement: %s", placementName)
+		klog.ErrorS(err, "Failed to find the latest policy snapshot", "clusterResourcePlacement", placementName, "clusterStagedUpdateRun", updateRunRef)
+		// again, no more retries here.
+		return nil, -1, fmt.Errorf("%w: %s", errInitializedFailed, err.Error())
+	}
+
+	latestPolicySnapshot := policySnapshotList.Items[0]
+	updateRun.Status.PolicySnapshotIndexUsed = latestPolicySnapshot.Name
+	// Get the cluster count from the policy snapshot.
+	clusterCount, err := annotations.ExtractNumOfClustersFromPolicySnapshot(&latestPolicySnapshot)
```
jwtty, line 121: From the code [here](url), the annotation is only added when the it's "pickN" strategy. This may not be correct. Need verification here.

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,433 @@
...
+		return nil, -1, fmt.Errorf("%w, %s", errInitializedFailed, annErr.Error())
+	}
+	updateRun.Status.PolicyObservedClusterCount = clusterCount
+	klog.V(2).InfoS("Found the latest policy snapshot", "latestPolicySnapshot", latestPolicySnapshot.Name, "observed cluster count", updateRun.Status.PolicyObservedClusterCount, "clusterStagedUpdateRun", updateRunRef)
+
+	if !condition.IsConditionStatusTrue(latestPolicySnapshot.GetCondition(string(placementv1beta1.PolicySnapshotScheduled)), latestPolicySnapshot.Generation) {
+		scheduleErr := fmt.Errorf("policy snapshot `%s` not fully scheduled yet", latestPolicySnapshot.Name)
+		klog.ErrorS(scheduleErr, "The policy snapshot is not scheduled successfully", "clusterResourcePlacement", placementName, "latestPolicySnapshot", latestPolicySnapshot.Name, "clusterStagedUpdateRun", updateRunRef)
+		// hmmmm, should we retry and see if the policy snapshot is scheduled later?
+		return nil, -1, fmt.Errorf("%w: %s", errInitializedFailed, scheduleErr.Error())
```
jwtty, line 135: Should we retry here instead of reporting initialization failure?

pkg/controllers/updaterun/initialization.go
ryanzhang-oss: I think you are correct, we should special handle the case that this annotation does not exsit.

pkg/controllers/updaterun/initialization.go
ryanzhang-oss: This function will be reused by the validation part where it should fail. I think the expectation is that the updateRun is only issued after we have a stable schedule.

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,433 @@
...
+	}
+	if err := r.Client.List(ctx, &bindingList, resourceBindingMatcher); err != nil {
+		klog.ErrorS(err, "Failed to list clusterResourceBindings", "clusterResourcePlacement", placementName, "latestPolicySnapshot", latestPolicySnapshot.Name, "clusterStagedUpdateRun", updateRunRef)
+		// list err can be retried.
+		return nil, nil, err
+	}
+	var toBeDeletedBindings, selectedBindings []*placementv1beta1.ClusterResourceBinding
+	for i, binding := range bindingList.Items {
+		if binding.Spec.SchedulingPolicySnapshotName == latestPolicySnapshot.Name {
+			if binding.Spec.State != placementv1beta1.BindingStateScheduled {
```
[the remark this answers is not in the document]
ryanzhang-oss: The bindings cannot be in the "Bond" state during the initialization as this is the controller that marks it as bound. However, this function was used in my implementation so we probably can't reuse this logic anymore after the initialization.

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,433 @@
...
+		placementv1beta1.CRPTrackingLabel: placementName,
+	}
+	if err := r.Client.List(ctx, &bindingList, resourceBindingMatcher); err != nil {
+		klog.ErrorS(err, "Failed to list clusterResourceBindings", "clusterResourcePlacement", placementName, "latestPolicySnapshot", latestPolicySnapshot.Name, "clusterStagedUpdateRun", updateRunRef)
+		// list err can be retried.
+		return nil, nil, err
+	}
+	var toBeDeletedBindings, selectedBindings []*placementv1beta1.ClusterResourceBinding
+	for i, binding := range bindingList.Items {
+		if binding.Spec.SchedulingPolicySnapshotName == latestPolicySnapshot.Name {
```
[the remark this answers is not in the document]
ryanzhang-oss: The rollingUpdate controller works with mutliple versions of the policySnapshot together, it ony cares about the min/max availabilty. Here, we work with only 1 version.

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,433 @@
...
+			klog.V(2).InfoS("Found a scheduled binding", "binding", binding.Name, "clusterResourcePlacement", placementName, "latestPolicySnapshot", latestPolicySnapshot.Name, "clusterStagedUpdateRun", updateRunRef)
+			selectedBindings = append(selectedBindings, &bindingList.Items[i])
+		} else {
+			klog.V(2).InfoS("Found a to-be-deleted binding", "binding", binding.Name, "clusterResourcePlacement", placementName, "latestPolicySnapshot", latestPolicySnapshot.Name, "clusterStagedUpdateRun", updateRunRef)
+			toBeDeletedBindings = append(toBeDeletedBindings, &bindingList.Items[i])
+		}
+	}
+
+	// should this be a valid case?
+	if len(selectedBindings) == 0 {
```
jwtty, line 174: Should this still be a valid case?

pkg/controllers/updaterun/initialization.go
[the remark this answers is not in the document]
jwtty: TODO: add a finalizer for the resourceSnapshot.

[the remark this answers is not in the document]
jwtty: As per discussion, bound and scheduled are both possible.

pkg/controllers/updaterun/initialization.go
jwtty: Should be invalid when both selectedBindings and deletingBindings are empty.

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,433 @@
...
+		return err
+	}
+	if len(masterResourceSnapshot.Annotations[placementv1beta1.ResourceGroupHashAnnotation]) == 0 {
+		nomasterErr := fmt.Errorf("resource snapshot `%s` is not a master snapshot", updateRun.Spec.ResourceSnapshotIndex)
+		klog.ErrorS(nomasterErr, "Failed to get the master resource snapshot", "resourceSnapshot", updateRun.Spec.ResourceSnapshotIndex, "clusterStagedUpdateRun", updateRunRef)
+		// no more retries here.
+		return fmt.Errorf("%w: %s", errInitializedFailed, nomasterErr.Error())
+	}
+	// Fetch all the matching overrides.
+	matchedCRO, matchedRO, err := overrider.FetchAllMatchingOverridesForResourceSnapshot(ctx, r.Client, r.InformerManager, updateRun.Spec.PlacementName, &masterResourceSnapshot)
```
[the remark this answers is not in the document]
jwtty: As per discussion, we currently only consider the latest-version overrides.

pkg/controllers/updaterun/suite_test.go
```diff
@@ -84,8 +88,31 @@ var _ = BeforeSuite(func() {
 	By("set k8s client same as the controller manager")
 	k8sClient = mgr.GetClient()
 
+	// setup our main reconciler
+	fakeInformer := &informer.FakeManager{
```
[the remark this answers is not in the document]
jwtty: The FakeManager is also used in workgenerator and rollout controller ITs. The only usage is to check whether a resource is cluster scoped or not, which is used by the override finder. But I've created a real InformerManager instead.

pkg/controllers/updaterun/initialization.go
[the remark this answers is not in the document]
jwtty: The finalizer manipulation will be in a separate PR.

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+		// no more retries here.
+		return fmt.Errorf("%w: %s", errInitializedFailed, missingErr.Error())
+	}
+	return nil
+}
+
+// validateAfterStageTask valides the afterStageTasks in the stage defined in the clusterStagedUpdateStrategy.
+// The error returned from this function is not retryable.
+func validateAfterStageTask(tasks []placementv1alpha1.AfterStageTask) error {
```
ryanzhang-oss, line 354: this probably can be converted to CEL validation

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -0,0 +1,758 @@
...
+			By("Creating a new clusterStagedUpdateRun")
+			Expect(k8sClient.Create(ctx, updateRun)).To(Succeed())
+
+			By("Validating the initialization failed")
+			validateFailedInitCondition(ctx, updateRun,
+				"parent clusterResourcePlacement does not have an external rollout strategy")
+		})
+
+		It("Should copy the ApplyStrategy in the CRP to the UpdateRun", func() {
+			By("Creating a new clusteResourcePlacement with rolling update strategy")
```
ryanzhang-oss, line 164:
```suggestion
			By("Creating a new clusterResourcePlacement with external rolling update strategy")
```

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -0,0 +1,758 @@
...
+		testUpdateStrategyName = "updatestrategy-" + utils.RandStr()
+		testCROName = "cro-" + utils.RandStr()
+		updateRunNamespacedName = types.NamespacedName{Name: testUpdateRunName}
+
+		updateRun = getTestClusterStagedUpdateRun()
+		crp = getTestClusterResourcePlacement()
+		policySnapshot = getTestClusterSchedulingPolicySnapshot(1)
+		updateStrategy = getTestClusterStagedUpdateStrategy()
+		clusterResourceOverride = getTestClusterResourceOverride()
+
```
ryanzhang-oss, line 62: Are those "create" functions or "get" functions?

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -0,0 +1,758 @@
...
+			By("Creating a new clusterStagedUpdateRun")
+			Expect(k8sClient.Create(ctx, updateRun)).To(Succeed())
+
+			By("Validating the initialization not failed due to no selected cluster")
+			// it should fail due to strategy not found
+			validateFailedInitCondition(ctx, updateRun, "referenced clusterStagedUpdateStrategy not found")
+		})
+
+		It("Should retry if the bindings are not in Scheduled or Bound state", func() {
```
ryanzhang-oss, line 384: we should not retry, the code is not right

pkg/controllers/updaterun/controller.go
```diff
@@ -69,7 +84,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
+	initCond := meta.FindStatusCondition(updateRun.Status.Conditions, string(placementv1alpha1.StagedUpdateRunConditionInitialized))
+	if !condition.IsConditionStatusTrue(initCond, updateRun.Generation) {
+		if condition.IsConditionStatusFalse(initCond, updateRun.Generation) {
+			klog.V(2).InfoS("The clusterStagedUpdateRun has failed to initialize", "errorMsg", initCond.Message, "clusterStagedUpdateRun", runObjRef)
+			return runtime.Result{}, nil
+		}
+		if toBeUpdatedBindings, toBeDeletedBindings, err = r.initialize(ctx, &updateRun); err != nil {
+			klog.ErrorS(err, "Failed to initialize the clusterStagedUpdateRun", "clusterStagedUpdateRun", runObjRef)
+			// errInitializedFailed cannot be retried.
+			if errors.Is(err, errInitializedFailed) {
```
ryanzhang-oss, line 99: we also should not retry if the error is "unexpected"

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,452 @@
...
+	// Fetch the ClusterResourcePlacement object.
+	placementName := updateRun.Spec.PlacementName
+	var crp placementv1beta1.ClusterResourcePlacement
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: placementName}, &crp); err != nil {
+		klog.ErrorS(err, "Failed to get ClusterResourcePlacement", "clusterResourcePlacement", placementName, "clusterStagedUpdateRun", updateRunRef)
+		if apierrors.IsNotFound(err) {
+			// we won't continue or retry the initialization if the ClusterResourcePlacement is not found.
+			return "", fmt.Errorf("%w: %s", errInitializedFailed, "parent clusterResourcePlacement not found")
+		}
+		return "", err
```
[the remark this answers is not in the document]
jwtty: fixed all instances. Thanks!

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,452 @@
...
+	// Get the latest policy snapshot.
+	var policySnapshotList placementv1beta1.ClusterSchedulingPolicySnapshotList
+	latestPolicyMatcher := client.MatchingLabels{
+		placementv1beta1.CRPTrackingLabel:      placementName,
+		placementv1beta1.IsLatestSnapshotLabel: "true",
+	}
+	if err := r.Client.List(ctx, &policySnapshotList, latestPolicyMatcher); err != nil {
+		klog.ErrorS(err, "Failed to list the latest policy snapshots", "clusterResourcePlacement", placementName, "clusterStagedUpdateRun", updateRunRef)
+		// err can be retried.
+		return nil, -1, err
```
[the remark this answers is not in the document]
jwtty: fixed all instances, thanks!

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,452 @@
...
+func (r *Reconciler) validateCRP(ctx context.Context, updateRun *placementv1alpha1.ClusterStagedUpdateRun) (string, error) {
+	updateRunRef := klog.KObj(updateRun)
+	// Fetch the ClusterResourcePlacement object.
+	placementName := updateRun.Spec.PlacementName
+	var crp placementv1beta1.ClusterResourcePlacement
+	if err := r.Client.Get(ctx, client.ObjectKey{Name: placementName}, &crp); err != nil {
+		klog.ErrorS(err, "Failed to get ClusterResourcePlacement", "clusterResourcePlacement", placementName, "clusterStagedUpdateRun", updateRunRef)
+		if apierrors.IsNotFound(err) {
+			// we won't continue or retry the initialization if the ClusterResourcePlacement is not found.
+			return "", fmt.Errorf("%w: %s", errInitializedFailed, "parent clusterResourcePlacement not found")
```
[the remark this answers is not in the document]
jwtty: fixed

pkg/controllers/updaterun/initialization.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+		return "", err
+	}
+	// Check if the ClusterResourcePlacement has an external rollout strategy.
+	if crp.Spec.Strategy.Type != placementv1beta1.ExternalRolloutStrategyType {
+		klog.V(2).InfoS("The clusterResourcePlacement does not have an external rollout strategy", "clusterResourcePlacement", placementName, "clusterStagedUpdateRun", updateRunRef)
+		return "", fmt.Errorf("%w: %s", errInitializedFailed, "parent clusterResourcePlacement does not have an external rollout strategy, current strategy: "+string(crp.Spec.Strategy.Type))
```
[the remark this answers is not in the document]
jwtty: fixed

pkg/controllers/updaterun/initialization_integration_test.go
jwtty: Thanks, fixed all instances.

pkg/controllers/updaterun/initialization_integration_test.go
jwtty: yea you are right. Forgot to change the error msg. Fixed.

pkg/controllers/updaterun/initialization_integration_test.go
jwtty: I changed to `generate-` prefix instead.

pkg/controllers/updaterun/controller.go
jwtty: I believe I wrapped all terminating errors including unexpected ones as `errInitializedFailed`. I think adding another case would make things more complicated.

pkg/controllers/updaterun/initialization.go
jwtty: Got an error when trying to add CEL validation: 
`unable to install CRDs onto control plane: unable to create CRD instances: unable to create CRD "clusterstagedupdateruns.placement.kubernetes-fleet.io": CustomResourceDefinition.apiextensions.k8s.io "clusterstagedupdateruns.placement.kubernetes-fleet.io" is invalid: spec.validation.openAPIV3Schema.properties[status].properties[stagedUpdateStrategySnapshot].properties[stages].items.properties[afterStageTasks].uniqueItems: Forbidden: uniqueItems cannot be set to true since the runtime complexity becomes quadratic`

pkg/controllers/updaterun/controller.go
ryanzhang-oss: oh, I miss read the code. However, since the outer error is always failedToInitialize, can we unwrap when we return the error. It might be easier to classify the error and create alert/monitor that way, especially now we emit error types like user error, unexpected or API error.

pkg/controllers/updaterun/controller_integration_test.go
```diff
@@ -214,7 +434,7 @@ func getTestApprovalRequest(name string) *placementv1alpha1.ClusterApprovalReque
 	}
 }
 
-func validateUpdateRunHasFinalizer(ctx context.Context, k8sClient client.Client, updateRun *placementv1alpha1.ClusterStagedUpdateRun) {
+func validateUpdateRunHasFinalizer(ctx context.Context, updateRun *placementv1alpha1.ClusterStagedUpdateRun) {
 	namepsacedName := types.NamespacedName{Name: updateRun.Name}
```
ryanzhang-oss, line 438:
```suggestion
	namepspacedName := types.NamespacedName{Name: updateRun.Name}
```

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -0,0 +1,740 @@
...
+					return fmt.Errorf("status mismatch: (-want +got):\n%s", diff)
+				}
+
+				return nil
+			}, timeout, interval).Should(Succeed(), "failed to validate the clusterStagedUpdateRun in the status")
+		})
+	})
+})
+
+func validateFailedInitCondition(ctx context.Context, updateRun *placementv1alpha1.ClusterStagedUpdateRun, message string) {
```
ryanzhang-oss, line 725: no need to add this in this PR but I think it will be good to also verify the type of error (user error, unexpected, or api error) in this function.

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -0,0 +1,740 @@
...
+				Status:             metav1.ConditionTrue,
+				ObservedGeneration: policySnapshot.Generation,
+				Reason:             "scheduled",
+			})
+			Expect(k8sClient.Status().Update(ctx, policySnapshot)).Should(Succeed(), "failed to update the policy snapshot condition")
+
+			By("Creating a new clusterStagedUpdateRun")
+			Expect(k8sClient.Create(ctx, updateRun)).To(Succeed())
+
+			By("Validating the initialization failed")
```
ryanzhang-oss, line 264: I don't seem to see where this fail? There is no validateFailedInitCondition function all.  Maybe this test case is just to verify the policySnapshot properties are copied to updateRun?

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -0,0 +1,740 @@
...
+					return fmt.Errorf("updateRun status `PolicySnapshotIndexUsed` mismatch: got %s, want %s", updateRun.Status.PolicySnapshotIndexUsed, policySnapshot.Name)
+				}
+				if updateRun.Status.PolicyObservedClusterCount != numberOfClustersAnnotation {
+					return fmt.Errorf("updateRun status `PolicyObservedClusterCount` mismatch: got %d, want %d", updateRun.Status.PolicyObservedClusterCount, numberOfClustersAnnotation)
+				}
+				return nil
+			}, timeout, interval).Should(Succeed(), "failed to update the updateRun status with policy snapshot details")
+		})
+
+		It("Should copy the latest policy snapshot details to the updateRun status -- pickFixed policy", func() {
```
ryanzhang-oss, line 279: same, no validateFailedInitCondition

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -0,0 +1,740 @@
...
+					return fmt.Errorf("updateRun status `PolicySnapshotIndexUsed` mismatch: got %s, want %s", updateRun.Status.PolicySnapshotIndexUsed, policySnapshot.Name)
+				}
+				if updateRun.Status.PolicyObservedClusterCount != 2 {
+					return fmt.Errorf("updateRun status `PolicyObservedClusterCount` mismatch: got %d, want %d", updateRun.Status.PolicyObservedClusterCount, 2)
+				}
+				return nil
+			}, timeout, interval).Should(Succeed(), "failed to update the updateRun status with policy snapshot details")
+		})
+
+		It("Should copy the latest policy snapshot details to the updateRun status -- pickAll policy", func() {
```
ryanzhang-oss, line 312: ditto

pkg/controllers/updaterun/initialization_integration_test.go
```diff
@@ -0,0 +1,740 @@
...
+
+			By("Validating the initialization not failed due to no selected cluster")
+			// it should fail due to strategy not found
+			validateFailedInitCondition(ctx, updateRun, "referenced clusterStagedUpdateStrategy not found")
+		})
+
+		It("Should fail to initialize if the bindings are not in Scheduled or Bound state", func() {
+			By("Creating a not scheduled clusterResourceBinding")
+			binding := generateTestClusterResourceBinding(policySnapshot.Name, "cluster-1")
+			binding.Spec.State = ""
```
ryanzhang-oss, line 387: what about 'unscheduled' case?

pkg/controllers/updaterun/initialization_integration_test.go
jwtty: good catch. Updated the message.

pkg/controllers/updaterun/initialization_integration_test.go
jwtty: fixed.

pkg/controllers/updaterun/initialization_integration_test.go
jwtty: fixed.

pkg/controllers/updaterun/initialization_integration_test.go
jwtty: Amended with the `Unscheduled` state.

pkg/controllers/updaterun/initialization_integration_test.go
jwtty: This one is related to above discussion. The initialization errors are not returned by the reconciler but just written to the initialization condition. The controller returns nil for these errors to avoid retrying. We can only check the error by checking the condition.

ryanzhang-oss (approved):

PR #977, opened 2024-11-27
feat: Add delete override implementation

ryanzhang-oss:

Fixes #

I have:


pkg/utils/validator/resourceoverride.go
```diff
@@ -102,9 +102,16 @@ func validateOverridePolicy(policy *fleetv1alpha1.OverridePolicy) error {
 				}
 			}
 		}
+		switch rule.OverrideType {
```
zhiying-lin, line 105: can be replaced by using cel

for example,

// +kubebuilder:validation:XValidation:rule="self.overrideType == 'JSONPatch' || self.jsonPatchOverrides.size() == 0 ",message="jsonPatchOverrides must be empty when type is not JSONPatch"

pkg/controllers/workgenerator/override.go
```diff
@@ -131,13 +137,12 @@ func (r *Reconciler) applyOverrides(resource *placementv1beta1.ResourceContent,
...
 			klog.ErrorS(err, "Failed to apply the override rules", "clusterResourceOverrideSnapshot", klog.KObj(snapshot))
-			return err
+			return false, err
 		}
 	}
 	klog.V(2).InfoS("Applied clusterResourceOverrideSnapshots", "resource", klog.KObj(&uResource), "numberOfOverrides", len(croMap[key]))
 
-	// If the resource is selected by both ClusterResourceOverride and ResourceOverride, ResourceOverride will win when
-	// resolving conflicts.
+	// If the resource is selected by both ClusterResourceOverride and ResourceOverride, ResourceOverride will win when resolving conflicts.
```
zhiying-lin, line 145: what happens for a namespace scope resource which is selected by both cro and ro?

for example, cro deletes the resource, and ro overrides the resource.

pkg/controllers/workgenerator/override_test.go
```diff
@@ -730,11 +808,11 @@ func TestApplyOverrides_namespacedScopeResource(t *testing.T) {
 				Kind:    "Deployment",
 			}: true,
 		},
-		IsClusterScopedResource: false,
+		IsClusterScopedResource: true,
```
zhiying-lin, line 811: should be "false"? deployment is not a cluster scoped resource.

pkg/controllers/workgenerator/override_test.go
```diff
@@ -1253,21 +1410,73 @@ func TestApplyOverrides_namespacedScopeResource(t *testing.T) {
 			},
 			wantErr: controller.ErrUserError,
 		},
+		{
```
zhiying-lin, line 1413: add more tests about delete type:
- cro delete, ro patch
- cro patch, ro delete
- cro delete, ro delete

pkg/controllers/workgenerator/controller.go
```diff
@@ -447,7 +452,7 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding *fleetv1be
 				activeWork[work.Name] = work
 				newWork = append(newWork, work)
 			} else {
-				simpleManifests = append(simpleManifests, fleetv1beta1.Manifest(selectedResource))
+				simpleManifests = append(simpleManifests, fleetv1beta1.Manifest(*selectedResource))
 			}
 		}
 		if len(simpleManifests) == 0 {
```
zhiying-lin, line 458: this log needs to be updated.

pkg/utils/validator/resourceoverride.go
ryanzhang-oss: good to know but unless we remove this validator, we need to fix it even with the CEL, right?

zhiying-lin: yeah, but the fix is much simpler, just need to add the type check.

pkg/controllers/workgenerator/override_test.go
```diff
@@ -658,13 +937,65 @@ func TestApplyOverrides_clusterScopedResource(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			wantErr: controller.ErrUserError,
+		},
+		{
+			name: "delete during the clusterResourceOverride",
```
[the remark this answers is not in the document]
ryanzhang-oss: isn't test 1719  the delete after patch. There is another delete after patch case for a field earlier.

pkg/utils/validator/resourceoverride.go
ryanzhang-oss: I think there is some advantage to keep the logic (and test) together intead of spliting them into two places.

zhiying-lin (approved):

PR #979, opened 2024-12-02
interface: support crp attached override

zhiying-lin:
Support triggering rollout when placement is set in the overrides

Fixes #

I have:


apis/placement/v1alpha1/override_types.go
```diff
@@ -48,6 +48,21 @@ type ClusterResourceOverrideSpec struct {
...
+	// If so, the override will trigger the placement rollout immediately when the rollout strategy type is RollingUpdate.
+	// Otherwise, it will be applied to the next rollout.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec.placement is immutable"
+	// +optional
+	Placement *PlacementRef `json:"placement,omitempty"`
+}
+
+// PlacementRef is the reference to a placement.
+// For now, we only support ClusterResourcePlacement.
+type PlacementRef struct {
```
ryanzhang-oss, line 62: we will have both CRP and RP, should this contain namespace too?

apis/placement/v1alpha1/override_types.go
```diff
@@ -48,6 +48,21 @@ type ClusterResourceOverrideSpec struct {
 	// Policy defines how to override the selected resources on the target clusters.
 	// +required
 	Policy *OverridePolicy `json:"policy"`
+
+	// Placement defines whether the override is applied to a specific placement or not.
+	// If so, the override will trigger the placement rollout immediately when the rollout strategy type is RollingUpdate.
+	// Otherwise, it will be applied to the next rollout.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec.placement is immutable"
```
ryanzhang-oss, line 55: is this here for the backward compatibility sake? How do we handle the case that the ref is nil? I thought it was too computational heavy?

apis/placement/v1alpha1/override_types.go
```diff
@@ -48,6 +48,21 @@ type ClusterResourceOverrideSpec struct {
...
+	// +optional
+	Placement *PlacementRef `json:"placement,omitempty"`
```
circy9, line 57: As the target of the placement, this field should be the first one?

circy9, line 57: Can we make it required initially? Supporting both seems overly complicated. We can support both if needed as "required -> optional" is backward compatible.

apis/placement/v1alpha1/override_types.go
zhiying-lin: It's mentioned in the comment, if the placement is nil, the override will be applied to the next rollout, and it won't trigger the selected placement immediately.

apis/placement/v1alpha1/override_types.go
zhiying-lin: If we want to make it required, we need to upgrade our api to v1beta1 cause it's a breaking change.

apis/placement/v1alpha1/override_types.go
zhiying-lin: There are four combinations,

RO attached to CRP
RO attached to RP, we won't support cross namespace at the beginning (security concern).
CRO attached to CRP
CRO attached to RP (not supported)

So that namespace won't be used. If needed, we can add later.

circy9 (approved):

ryanzhang-oss (approved):

PR #980, opened 2024-12-02
feat: Add CustomResourceDefinition trackability

britaniar:

Fixes #

I have: add trackability for CustomResourceDefinitions

### DIFF
--- a/pkg/controllers/work/apply_controller.go
+++ b/pkg/controllers/work/apply_controller.go
@@ -29,6 +29,7 @@ import (
 	"go.uber.org/atomic"
 	appv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -448,6 +449,9 @@ func trackResourceAvailability(gvr schema.GroupVersionResource, curObj *unstruct
 	case utils.ServiceGVR:
 		return trackServiceAvailability(curObj)
 
+	case utils.CustomResourceDefinitionGVR:
+		return trackCRDAvailability(curObj)
+
 	default:
 		if isDataResource(gvr) {
 			klog.V(2).InfoS("Data resources are available immediately", "gvr", gvr, "resource", klog.KObj(curObj))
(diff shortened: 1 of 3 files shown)


pkg/controllers/work/apply_controller.go
```diff
@@ -458,6 +462,30 @@ func trackResourceAvailability(gvr schema.GroupVersionResource, curObj *unstruct
 	}
 }
 
+func trackCRDAvailability(curObj *unstructured.Unstructured) (ApplyAction, error) {
+	var crd apiextensionsv1.CustomResourceDefinition
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(curObj.Object, &crd); err != nil {
+		return errorApplyAction, controller.NewUnexpectedBehaviorError(err)
+	}
+	// Check if there is a condition of type "Established"
+	var establishedCondition *apiextensionsv1.CustomResourceDefinitionCondition
```
ryanzhang-oss, line 471: what about "NamesAccepted" condition?

ryanzhang-oss (commented):
Why is there already test at test/e2e/rollout_test.go:1074 for CRD readiness?

pkg/controllers/work/apply_controller.go
britaniar: My thought process is that if the established condition exists then the name was accepted and the CRD is either unavailable or available. That is the main condition. Otherwise, if "NamesAccepted" is true we still need to check for established and if it is false then it is unavailable (i.e. established doesn't exist).

britaniar:
The [waitForCRDToBeReady](https://github.com/Azure/fleet/blob/30226419b9cc36c476bf73cee42d645984f662d2/test/e2e/rollout_test.go#L1074) in the rollout e2e was because we needed it to be ready in the hub before we can create the custom resource in the hub that it defines. It made the test flaky because it would sometimes take longer to become established.

pkg/controllers/work/apply_controller.go
ryanzhang-oss: In this case, why the waitForCRDToBeReady waited for both conditions? I wonder if you have verified that the "NamesAccepted" condition is set to true before the "Established?

britaniar:
```
// Established means that the resource has become active. A resource is established when all names are
// accepted without a conflict for the first time. A resource stays established until deleted, even during
// a later NamesAccepted due to changed names. Note that not all names can be changed.
```
https://github.com/kubernetes/kubernetes/blob/a27498ef42f27e7cddadb2c5b31bc80fed81d964/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/types.go#L301C2-L303C88

Looks like there could be a case in which "NamesAccepted" fails after it is established and stays established. If we check for "NamesAccepted" to be true, will we consider still consider it as `manifestNotAvailableYetAction`? Technically it is still available but the name failed to be changed.

pkg/controllers/work/apply_controller.go
```diff
@@ -458,6 +462,35 @@ func trackResourceAvailability(gvr schema.GroupVersionResource, curObj *unstruct
 	}
 }
 
+func trackCRDAvailability(curObj *unstructured.Unstructured) (ApplyAction, error) {
+	var crd apiextensionsv1.CustomResourceDefinition
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(curObj.Object, &crd); err != nil {
+		return errorApplyAction, controller.NewUnexpectedBehaviorError(err)
+	}
+
+	// Check if both "Established" and "NamesAccepted" conditions exist and are True
```
ryanzhang-oss, line 471: apiextensionshelpers "k8s.io/apiextensions-apiserver/pkg/apihelpers"


apiextensionshelpers.IsCRDConditionTrue(crd, apiextensionsv1.NamesAccepted)

ryanzhang-oss (approved):

PR #985, opened 2024-12-09
feat: drift detection and takeover implementation (1/)

michaelawyu:

This PR adds the implementation of the new work applier with drift detection and takeover experience.

I have:

- [*] Unit tests
- [*] Integration tests (see the RC branch)
- [*] Manual tests

As the new drift detection and takeover experience requires significant changes to the work applier (e.g., originally the work applier simply applies with no need to account for drifts/diffs/ownership mgmt, etc.), to avoid difficulties caused by maintenance of working intermediate states, the new experience will be submitted as a new controller that is not currently enabled.

A preview of the full experience is available in the preview PR (and in the preview env.).


pkg/controllers/workapplier/apply.go
```diff
@@ -0,0 +1,539 @@
...
+		return fmt.Errorf("manifest is set to have multiple owner references but co-ownership is disallowed")
+	}
+
+	// Do a sanity check to verify that no AppliedWork object is directly added as an owner
+	// in the manifest object. Normally the branch will never get triggered as Fleet would
+	// perform sanitization on the manifest object before applying it, which removes all owner
+	// references.
+	for _, ownerRef := range manifestObjOwnerRefs {
+		if ownerRef.APIVersion == fleetv1beta1.GroupVersion.String() && ownerRef.Kind == fleetv1beta1.AppliedWorkKind {
+			return fmt.Errorf("an AppliedWork object is unexpectedly added as an owner in the manifest object")
```
ryanzhang-oss, line 443: not unexpectedError?

pkg/controllers/workapplier/apply.go
```diff
@@ -0,0 +1,539 @@
...
+	// round of sanitization.
+	//
+	// TO-DO (chenyu1): this processing step can be removed if the enveloped objects are
+	// pre-processed by the Fleet hub agent as well, provided that there are no further
+	// backwards compatibility concerns.
+	manifestObjCopy := sanitizeManifestObject(manifestObj)
+
+	// Compute the hash of the manifest object.
+	//
+	// Originally the manifest hash is kept only if three-way merge patch (client side apply esque
```
ryanzhang-oss, line 77:
```suggestion
	// Originally the manifest hash is kept only if three-way merge patch (client side apply
```

pkg/controllers/workapplier/apply.go
```diff
@@ -0,0 +1,539 @@
...
+	//
+	// This is OK as partial comparison concerns only fields that are currently present
+	// in the manifest object, and Fleet will clear out system managed and read-only fields
+	// before the comparison.
+	//
+	// Note that full comparison can be carried out directly without involving the apply op.
+	return r.serverSideApply(ctx, gvr, manifestObj, inMemberClusterObj, true, false, true)
+}
+
+func (r *Reconciler) apply(
```
ryanzhang-oss, line 55: we should try to differentiate the type of errors
1. userError
2. none-retriable error
3. retriable error

ryanzhang-oss: This is very important when we get to upgrade so that we know if it's something we can continue or give up

pkg/controllers/workapplier/apply.go
```diff
@@ -0,0 +1,539 @@
...
+	return nil
+}
+
+// setOwnerRef sets the expected owner reference (reference to an AppliedWork object)
+// on a manifest to be applied.
+func setOwnerRef(obj *unstructured.Unstructured, expectedAppliedWorkOwnerRef *metav1.OwnerReference) {
+	ownerRefs := obj.GetOwnerReferences()
+	if ownerRefs == nil {
+		ownerRefs = []metav1.OwnerReference{}
+	}
```
ryanzhang-oss, line 409: I don't think you need this, the append func handles ownerRef == nil case

pkg/controllers/workapplier/apply.go
```diff
@@ -0,0 +1,539 @@
...
+	// Typically owner references is a system-managed field, and at this moment Fleet will
+	// clear owner references (if any) set in the manifest object. However, for consistency
+	// reasons, here Fleet will still assume that there might be some owner references set
+	// in the manifest object.
```
ryanzhang-oss, line 413: TODO: try to preserve the original owner reference

pkg/controllers/workapplier/drift_detection_takeover.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	return updatedOwnerRefs, nil
+}
+
+// discardFieldsIrrelevantInComparisonFrom discards fields that are irrelevant when comparing
+// the manifest and live objects (or two manifest objects).
+//
+// Note that this method will return an object copy; the original object will be left untouched.
+func discardFieldsIrrelevantInComparisonFrom(obj *unstructured.Unstructured) *unstructured.Unstructured {
```
ryanzhang-oss, line 319: This func probably belongs to util as it's called by multiple modules

pkg/controllers/workapplier/preprocess.go
```diff
@@ -0,0 +1,547 @@
...
+	if gvr != nil {
+		identifier.Resource = gvr.Resource
+	}
+
+	return identifier
+}
+
+// prepareExistingManifestCondQIdx returns a map that allows quicker look up of a manifest
+// condition given a work resource identifier.
+func prepareExistingManifestCondQIdx(existingManifestConditions []fleetv1beta1.ManifestCondition) map[string]int {
```
ryanzhang-oss, line 273: this function is very similar to the prepareRebuiltManifestCondQIdx function in status.go. Is there a way to unify them and move it to util.go?

pkg/controllers/workapplier/preprocess.go
```diff
@@ -0,0 +1,547 @@
...
+			continue
+		}
+
+		existingManifestConditionQIdx[wriStr] = idx
+	}
+	return existingManifestConditionQIdx
+}
+
+// prepareManifestCondForWA prepares a manifest condition for the write-ahead process.
+func prepareManifestCondForWA(
```
ryanzhang-oss, line 292:
```suggestion
func prepareManifestCondForWriteAhead(
```

ryanzhang-oss (approved):

michaelawyu:
Per offline discussions, will merge this PR now and open a new one to address comments. If there’s any further concern, please let me know.

pkg/controllers/workapplier/apply.go
michaelawyu: Done.

pkg/controllers/workapplier/apply.go
michaelawyu: Done.

pkg/controllers/workapplier/apply.go
michaelawyu: Done; the categorization is done slightly different from other controllers: it's not passed back as there are internal only information that should not be exposed to the end user in the K8s condition; some errors (e.g., drift/diff found) are not registered as user errors as they are part of the flow.

pkg/controllers/workapplier/apply.go
michaelawyu: TO-DO added.

pkg/controllers/workapplier/preprocess.go
michaelawyu: Hi Ryan! Yeah, the two functions are very similar; but currently I fear that I do not seem to have an elegant enough solution to unify the two.

I could use generics and write something like this:

```
// prepareManifestCondQIdx returns a map that allows quicker look up of a manifest
// condition given a work resource identifier.
func prepareManifestCondQIdx[E fleetv1beta1.ManifestCondition | *manifestProcessingBundle](S []E) map[string]int {
	manifestQIdx := make(map[string]int)

	for idx := range S {
		E := S[idx]

		var id *fleetv1beta1.WorkResourceIdentifier
		mc, mcOK := any(E).(fleetv1beta1.ManifestCondition)
		b, bOK := any(E).(*manifestProcessingBundle)
		switch {
		case mcOK:
			id = &mc.Identifier
		case bOK:
			id = b.id
		default:
                        // This will never run, or the code will not compile.
			panic("unexpected type is passed in")
		}

		wriStr, err := formatWRIString(id)
		if err != nil {
			continue
		}

		manifestQIdx[wriStr] = idx
	}
	return manifestQIdx
}
```

But it seems to be overcomplicating things 😣 Alternatives include a) introducing an interface that allows retrieval of identifiers from `ManifestCondition` and `manifestProcessingBundle` types, or b) extracting the identifiers into a separate slice first, both might not be fully satisfactory (overkilling, or too much overhead) 😣 If you have a preference, please let me know.

PR #987, opened 2024-12-11
fix: use CRP defaulter to account for nil CRP policy in eviction controller

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterresourceplacementeviction/controller.go
+++ b/pkg/controllers/clusterresourceplacementeviction/controller.go
@@ -274,15 +274,13 @@ func isEvictionAllowed(bindings []placementv1beta1.ClusterResourceBinding, crp p
 	var placementType placementv1beta1.PlacementType
 	if crp.Spec.Policy != nil {
 		placementType = crp.Spec.Policy.PlacementType
-	} else {
-		placementType = placementv1beta1.PickAllPlacementType
 	}
 	switch placementType {
 	case placementv1beta1.PickNPlacementType:
 		desiredBindings = int(*crp.Spec.Policy.NumberOfClusters)
 	case placementv1beta1.PickFixedPlacementType:
 		desiredBindings = len(crp.Spec.Policy.ClusterNames)
-	case placementv1beta1.PickAllPlacementType:
+	default:
 		// we don't know the desired bindings for PickAll.
 	}
 
--- a/pkg/controllers/clusterresourceplacementeviction/controller.go
+++ b/pkg/controllers/clusterresourceplacementeviction/controller.go
@@ -209,7 +209,7 @@ func (r *Reconciler) executeEviction(ctx context.Context, validationResult *evic
 	}
 
 	// handle special case for PickAll CRP.
-	if crp.Spec.Policy.PlacementType == placementv1beta1.PickAllPlacementType {
+	if crp.Spec.Policy == nil || crp.Spec.Policy.PlacementType == placementv1beta1.PickAllPlacementType {
 		if db.Spec.MaxUnavailable != nil || (db.Spec.MinAvailable != nil && db.Spec.MinAvailable.Type == intstr.String) {
 			markEvictionNotExecuted(eviction, evictionBlockedMisconfiguredPDBSpecifiedMessage)
 			return nil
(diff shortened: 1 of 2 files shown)


pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -209,7 +209,7 @@ func (r *Reconciler) executeEviction(ctx context.Context, validationResult *evic
 	}
 
 	// handle special case for PickAll CRP.
-	if crp.Spec.Policy.PlacementType == placementv1beta1.PickAllPlacementType {
+	if crp.Spec.Policy == nil || crp.Spec.Policy.PlacementType == placementv1beta1.PickAllPlacementType {
```
jwtty, line 212: Why the default value here does not take effect: https://github.com/Azure/fleet/blob/main/apis/placement/v1beta1/clusterresourceplacement_types.go#L136?

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -271,7 +271,12 @@ func isEvictionAllowed(bindings []placementv1beta1.ClusterResourceBinding, crp p
 	}
 
 	var desiredBindings int
-	placementType := crp.Spec.Policy.PlacementType
+	var placementType placementv1beta1.PlacementType
+	if crp.Spec.Policy != nil {
```
michaelawyu, line 275: Hi Arvind! Just a nit: we could use PickAll as a default and set the value if policy is not nil (to avoid empty strings being set a default).

michaelawyu (commented):
Just a nit. LGTM.

michaelawyu (approved):

pkg/controllers/clusterresourceplacementeviction/controller.go
ryanzhang-oss: can we follow the pattern to call defaulter.SetDefaultsClusterResourcePlacement() so we don't need to handle those cases specially

pkg/controllers/clusterresourceplacementeviction/controller_test.go
```diff
@@ -503,16 +530,7 @@ func TestExecuteEviction(t *testing.T) {
...
-					ObjectMeta: metav1.ObjectMeta{
-						Name: testCRPName,
-					},
-					Spec: placementv1beta1.ClusterResourcePlacementSpec{
-						Policy: &placementv1beta1.PlacementPolicy{
-							PlacementType: placementv1beta1.PickAllPlacementType,
-						},
-					},
-				},
+				crp: ptr.To(buildTestPickAllCRP(testCRPName)),
```
ryanzhang-oss, line 533: I wonder why sometimes we use this while sometimes we just use testCRP

pkg/controllers/clusterresourceplacementeviction/controller_test.go
```diff
@@ -31,6 +33,14 @@ const (
 	testEvictionName         = "test-eviction"
 )
 
+var (
+	validationResultCmpOptions = []cmp.Option{
+		cmp.AllowUnexported(evictionValidationResult{}),
+		cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion"),
+		cmpopts.IgnoreFields(placementv1beta1.ClusterResourceBinding{}, "ResourceVersion"),
```
zhiying-lin, line 40: This should be covered by the above ignore, cause "ResourceVersion" is part of metav1.ObjectMeta{}

pkg/controllers/clusterresourceplacementeviction/controller_test.go
Arvindthiru: testCRP is only used for validateEviction test where we use defaulter to set default values on the CRP. For all the other tests we can build CRP which doesn't need to have all default values

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -277,7 +282,7 @@ func isEvictionAllowed(bindings []placementv1beta1.ClusterResourceBinding, crp p
 		desiredBindings = int(*crp.Spec.Policy.NumberOfClusters)
 	case placementv1beta1.PickFixedPlacementType:
```
ryanzhang-oss, line 283: Why do we allow eviction on fixed type? It will be put back

Arvindthiru: Will handle this case in a follow up PR

ryanzhang-oss (approved):

PR #989, opened 2024-12-13
fix: allow 0 maxUnavailable 

ryanzhang-oss:
allow 0 maxUnavailable so that user can migrate one application from one cluster to another without down time.

Fixes #

I have:

### DIFF
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -388,8 +388,8 @@ func validateRolloutStrategy(rolloutStrategy placementv1beta1.RolloutStrategy) e
 			if err != nil {
 				allErr = append(allErr, fmt.Errorf("maxUnavailable `%+v` is invalid: %w", rolloutStrategy.RollingUpdate.MaxUnavailable, err))
 			}
-			if value < 1 {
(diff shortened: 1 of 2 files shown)


pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -388,8 +388,8 @@ func validateRolloutStrategy(rolloutStrategy placementv1beta1.RolloutStrategy) e
 			if err != nil {
 				allErr = append(allErr, fmt.Errorf("maxUnavailable `%+v` is invalid: %w", rolloutStrategy.RollingUpdate.MaxUnavailable, err))
 			}
-			if value < 1 {
-				allErr = append(allErr, fmt.Errorf("maxUnavailable must be greater than or equal to 1, got `%+v`", rolloutStrategy.RollingUpdate.MaxUnavailable))
+			if value < 0 {
```
zhiying-lin, line 391: update our api comment? should we get the approval from liqian for that api comment change?

circy9: +1 on updating the api comment as well if it is different from the PR.

zhiying-lin (approved):

PR #991, opened 2024-12-16
feat: implement clusterStagedUpdateRun validation

jwtty:
This PR implements the `clusterStagedUpdateRun` validation

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -40,6 +40,10 @@ var (
 	// errInitializedFailed is the error when the ClusterStagedUpdateRun fails to initialize.
 	// It is a wrapped error of errStagedUpdatedAborted, because some initialization functions are reused in the validation step.
 	errInitializedFailed = fmt.Errorf("%w: failed to initialize the clusterStagedUpdateRun", errStagedUpdatedAborted)
+
+	// stageUpdatingWaitTime is the time to wait before rechecking the stage update status.
+	// Put it as a variable for convenient testing.
+	stageUpdatingWaitTime = 60 * time.Second
 )
 
 // Reconciler reconciles a ClusterStagedUpdateRun object.
@@ -108,16 +112,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		// Check if the clusterStagedUpdateRun is finished.
 		finishedCond := meta.FindStatusCondition(updateRun.Status.Conditions, string(placementv1alpha1.StagedUpdateRunConditionSucceeded))
 		if condition.IsConditionStatusTrue(finishedCond, updateRun.Generation) || condition.IsConditionStatusFalse(finishedCond, updateRun.Generation) {
-			klog.V(2).InfoS("The clusterStagedUpdateRun is finished", "clusterStagedUpdateRun", runObjRef)
+			klog.V(2).InfoS("The clusterStagedUpdateRun is finished", "finishedSuccessfully", finishedCond.Status, "clusterStagedUpdateRun", runObjRef)
 			return runtime.Result{}, nil
 		}
-		// TODO(wantjian): validate the clusterStagedUpdateRun and generate the updatingStage etc.
+
+		// Validate the clusterStagedUpdateRun status to ensure the update can be continued and get the updating stage index and cluster indices.
+		if updatingStageIndex, toBeUpdatedBindings, toBeDeletedBindings, err = r.validate(ctx, &updateRun); err != nil {
+			// errStagedUpdatedAborted cannot be retried.
+			if errors.Is(err, errStagedUpdatedAborted) {
+				return runtime.Result{}, r.recordUpdateRunFailed(ctx, &updateRun, err.Error())
+			}
+			return runtime.Result{}, err
+		}
+		klog.V(2).InfoS("The clusterStagedUpdateRun is validated", "clusterStagedUpdateRun", runObjRef)
 	}
 
-	// TODO(wantjian): execute the clusterStagedUpdateRun.
+	// TODO(wantjian): execute the clusterStagedUpdateRun and fix the requeue time.
 	klog.V(2).InfoS("Executing the clusterStagedUpdateRun", "clusterStagedUpdateRun", runObjRef, "updatingStageIndex", updatingStageIndex,
 		"toBeUpdatedBindings count", len(toBeUpdatedBindings), "toBeDeletedBindings count", len(toBeDeletedBindings))
-	return runtime.Result{}, nil
+	return runtime.Result{RequeueAfter: stageUpdatingWaitTime}, nil
 }
 
 // handleDelete handles the deletion of the clusterStagedUpdateRun object.
--- a/pkg/controllers/updaterun/controller_integration_test.go
+++ b/pkg/controllers/updaterun/controller_integration_test.go
@@ -35,6 +35,8 @@ const (
 	timeout = time.Second * 10
 	// interval is the time to wait between retries for Eventually and Consistently
 	interval = time.Millisecond * 250
+	// duration is the time to duration to check for Consistently
+	duration = time.Second * 20
 
 	// numTargetClusters is the number of scheduled clusters
 	numTargetClusters = 10
@@ -107,8 +109,8 @@ var _ = Describe("Test the clusterStagedUpdateRun controller", func() {
 			validateUpdateRunHasFinalizer(ctx, updateRun)
 
 			By("Updating the clusterStagedUpdateRun to failed")
-			startedcond := getTrueCondition(updateRun, string(placementv1alpha1.StagedUpdateRunConditionProgressing))
-			finishedcond := getFalseCondition(updateRun, string(placementv1alpha1.StagedUpdateRunConditionSucceeded))
+			startedcond := generateTrueCondition(updateRun, placementv1alpha1.StagedUpdateRunConditionProgressing)
+			finishedcond := generateFalseCondition(updateRun, placementv1alpha1.StagedUpdateRunConditionSucceeded)
 			meta.SetStatusCondition(&updateRun.Status.Conditions, startedcond)
 			meta.SetStatusCondition(&updateRun.Status.Conditions, finishedcond)
 			Expect(k8sClient.Status().Update(ctx, updateRun)).Should(Succeed(), "failed to update the clusterStagedUpdateRun")
@@ -136,7 +138,7 @@ var _ = Describe("Test the clusterStagedUpdateRun controller", func() {
 			validateUpdateRunHasFinalizer(ctx, updateRun)
 
 			By("Updating the clusterStagedUpdateRun status to processing")
-			startedcond := getTrueCondition(updateRun, string(placementv1alpha1.StagedUpdateRunConditionProgressing))
+			startedcond := generateTrueCondition(updateRun, placementv1alpha1.StagedUpdateRunConditionProgressing)
 			meta.SetStatusCondition(&updateRun.Status.Conditions, startedcond)
 			Expect(k8sClient.Status().Update(ctx, updateRun)).Should(Succeed(), "failed to add condition to the clusterStagedUpdateRun")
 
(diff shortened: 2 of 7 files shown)


pkg/controllers/updaterun/controller_integration_test.go
```diff
@@ -35,6 +35,8 @@ const (
 	timeout = time.Second * 10
 	// interval is the time to wait between retries for Eventually and Consistently
 	interval = time.Millisecond * 250
+	// duration is the time to duration to check for Consistently
+	duration = time.Second * 20
```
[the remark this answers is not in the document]
jwtty: `duration` is used in initialization_integration_test.go

pkg/controllers/updaterun/validation.go
```diff
@@ -0,0 +1,319 @@
...
+			klog.ErrorS(unexpectedErr, "Detected more than one updating clusters in the stage", "clusterStagedUpdateRun", klog.KObj(updateRun))
+			return -1, -1, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
+		}
+	}
+	return updatingStageIndex, lastFinishedStageIndex, nil
+}
+
+// validateDeleteStageStatus validates the delete stage in the clusterStagedUpdateRun.
+// It returns the updating stage index, or any error encountered.
+func validateDeleteStageStatus(
```
jwtty, line 250: The clusters in the deletionStage are not validated. I think it's ok because cluster might unjoin the fleet?

pkg/controllers/updaterun/validation.go
```diff
@@ -0,0 +1,319 @@
...
+	klog.V(2).InfoS("Start to validate the clusterStagedUpdateRun", "clusterStagedUpdateRun", updateRunRef)
+
+	// Validate the ClusterResourcePlacement object referenced by the ClusterStagedUpdateRun.
+	placementName, err := r.validateCRP(ctx, updateRunCopy)
+	if err != nil {
+		return -1, nil, nil, err
+	}
+	// Validate the applyStrategy.
+	if !reflect.DeepEqual(updateRun.Status.ApplyStrategy, updateRunCopy.Status.ApplyStrategy) {
+		mismatchErr := fmt.Errorf("the applyStrategy in the clusterStagedUpdateRun is outdated, latest: %v, recorded: %v", updateRunCopy.Status.ApplyStrategy, updateRun.Status.ApplyStrategy)
```
ryanzhang-oss, line 44: is this an user error?

pkg/controllers/updaterun/validation.go
```diff
@@ -0,0 +1,319 @@
...
+		klog.ErrorS(mismatchErr, "the applyStrategy in the clusterResourcePlacement has changed", "clusterResourcePlacement", placementName, "clusterStagedUpdateRun", updateRunRef)
+		return -1, nil, nil, fmt.Errorf("%w: %s", errStagedUpdatedAborted, mismatchErr.Error())
+	}
+
+	// Retrieve the latest policy snapshot.
+	latestPolicySnapshot, clusterCount, err := r.determinePolicySnapshot(ctx, placementName, updateRunCopy)
```
ryanzhang-oss, line 50: please check if the policy in that func can be nil.

pkg/controllers/updaterun/validation.go
```diff
@@ -0,0 +1,319 @@
...
+	scheduledBindings, toBeDeletedBindings, err := r.collectScheduledClusters(ctx, placementName, latestPolicySnapshot, updateRunCopy)
+	if err != nil {
+		return -1, nil, nil, err
+	}
+
+	// if condition.IsConditionStatusFalse(meta.FindStatusCondition(updateRun.Status.Conditions, string(placementv1alpha1.StagedUpdateRunConditionProgressing)), updateRun.Generation) {
+	// 	// The clusterStagedUpdateRun has not started yet.
+	// 	klog.V(2).InfoS("Starting the staged update run from the beginning", "clusterStagedUpdateRun", updateRunRef)
+	// 	return 0, scheduledBindings, toBeDeletedBindings, nil
+	// }
```
ryanzhang-oss, line 77: why comment it out?

pkg/controllers/updaterun/validation.go
```diff
@@ -0,0 +1,319 @@
...
+		if err != nil {
+			return -1, -1, err
+		}
+	}
+	return updatingStageIndex, lastFinishedStageIndex, nil
+}
+
+// determineUpdatingStage validates a stage status and looks for the updating stage and last finished stage.
+// It returns latest updatingStageIndex, lastFinishedStageIndex and any error encountered.
+func determineUpdatingStage(
```
ryanzhang-oss, line 172: the function probably could use a bit more descriptive name

pkg/controllers/updaterun/validation_test.go
```diff
@@ -0,0 +1,358 @@
...
+*/
+
+package updaterun
+
+import (
+	"fmt"
+	"testing"
+
+	placementv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
+	"go.goms.io/fleet/pkg/utils/controller"
```
ryanzhang-oss, line 13:
```suggestion


	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

	placementv1alpha1 "go.goms.io/fleet/apis/placement/v1alpha1"
	"go.goms.io/fleet/pkg/utils/controller"```

File: pkg/controllers/updaterun/validation_test.go [hunk 8]
```diff
@@ -0,0 +1,358 @@
...
+			lastFinishedStageIndex: -1,
+			stageStatus: &placementv1alpha1.StageUpdatingStatus{
+				StageName:  "test-stage",
+				Conditions: []metav1.Condition{generateTrueCondition(updateRun, placementv1alpha1.StageUpdatingConditionProgressing)},
+			},
+			wantErr:                    wrapErr(true, fmt.Errorf("the updating stage `test-stage` is not right after the last finished stage with index `-1`")),
+			wantUpdatingStageIndex:     -1,
+			wantLastFinishedStageIndex: -1,
+		},
+		{
```
ryanzhang-oss, line 135: Do we have the case
 stage update status cannot be true if there are still cluster updating in it

pkg/controllers/updaterun/validation.go
```diff
@@ -0,0 +1,319 @@
...
+		var updatingClusters []string
+		for j := range stageStatus.Clusters {
+			clusterStartedCond := meta.FindStatusCondition(stageStatus.Clusters[j].Conditions, string(placementv1alpha1.ClusterUpdatingConditionStarted))
+			clusterFinishedCond := meta.FindStatusCondition(stageStatus.Clusters[j].Conditions, string(placementv1alpha1.ClusterUpdatingConditionSucceeded))
+			if condition.IsConditionStatusTrue(clusterStartedCond, updateRun.Generation) &&
+				!(condition.IsConditionStatusTrue(clusterFinishedCond, updateRun.Generation) || condition.IsConditionStatusFalse(clusterFinishedCond, updateRun.Generation)) {
+				updatingClusters = append(updatingClusters, stageStatus.Clusters[j].ClusterName)
+			}
+		}
+		// We don't allow more than one clusters to be updating at the same time.
```
ryanzhang-oss, line 238: can u add a TODO to support more than one cluster updating at the same time.

pkg/controllers/updaterun/validation.go
```diff
@@ -0,0 +1,319 @@
...
+		var err error
+		updatingStageIndex, lastFinishedStageIndex, err = determineUpdatingStage(curStage, updatingStageIndex, lastFinishedStageIndex, &existingStageStatus[curStage], updateRun)
+		if err != nil {
+			return -1, -1, err
+		}
+	}
+	return updatingStageIndex, lastFinishedStageIndex, nil
+}
+
+// determineUpdatingStage validates a stage status and looks for the updating stage and last finished stage.
```
ryanzhang-oss, line 170: This comment could use a bit more wording too.

pkg/controllers/updaterun/validation.go
ryanzhang-oss: Do you mean that we don't compare it with the exisitng list?  I think we still should but since the deleted binding will disapper, it means the new list should be a subset of the old?

pkg/controllers/updaterun/validation_integration_test.go
```diff
@@ -0,0 +1,485 @@
...
+	})
+
+	Context("Test validateCRP", func() {
+		It("Should fail to validate if the CRP is not found", func() {
+			By("Creating a new clusterStagedUpdateRun")
+			Expect(k8sClient.Create(ctx, updateRun)).To(Succeed())
+
+			By("Validating the initialization succeeded")
+			want := generateSucceededInitializationStatus(crp, updateRun, policySnapshot, updateStrategy, clusterResourceOverride)
+			validateSucceededInitializationStatus(ctx, updateRun, want)
```
ryanzhang-oss, line 180: seems that those can be extracted out to the BeforeEach inside this context?

pkg/controllers/updaterun/validation_integration_test.go
```diff
@@ -0,0 +1,485 @@
...
+
+	Context("Test determinePolicySnapshot", func() {
+		It("Should fail to validate if the latest policySnapshot is not found", func() {
+			By("Creating a new clusterStagedUpdateRun")
+			Expect(k8sClient.Create(ctx, updateRun)).To(Succeed())
+
+			By("Validating the initialization succeeded")
+			want := generateSucceededInitializationStatus(crp, updateRun, policySnapshot, updateStrategy, clusterResourceOverride)
+			validateSucceededInitializationStatus(ctx, updateRun, want)
+
```
ryanzhang-oss, line 234: same

pkg/controllers/updaterun/initialization.go
```diff
@@ -340,8 +341,9 @@ func (r *Reconciler) computeRunStageStatus(
 				curStageUpdatingStatus.AfterStageTaskStatus[i].ApprovalRequestName = fmt.Sprintf(placementv1alpha1.ApprovalTaskNameFmt, updateRun.Name, stage.Name)
 			}
 		}
-		updateRun.Status.StagesStatus = append(updateRun.Status.StagesStatus, curStageUpdatingStatus)
+		stagesStatus = append(stagesStatus, curStageUpdatingStatus)
```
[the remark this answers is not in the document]
jwtty: In validation, we make a copy of the updateRun as the parameter of function. By using a local variable, we make sure the stageStatus is regenerated. With the previous field assignment, the newly generated stages are appended, resulting in validation failure.

pkg/controllers/updaterun/controller.go
```diff
@@ -108,16 +112,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		// Check if the clusterStagedUpdateRun is finished.
 		finishedCond := meta.FindStatusCondition(updateRun.Status.Conditions, string(placementv1alpha1.StagedUpdateRunConditionSucceeded))
 		if condition.IsConditionStatusTrue(finishedCond, updateRun.Generation) || condition.IsConditionStatusFalse(finishedCond, updateRun.Generation) {
-			klog.V(2).InfoS("The clusterStagedUpdateRun is finished", "clusterStagedUpdateRun", runObjRef)
+			klog.V(2).InfoS("The clusterStagedUpdateRun is finished", "finishedSuccessfully", finishedCond.Status, "clusterStagedUpdateRun", runObjRef)
```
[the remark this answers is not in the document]
jwtty: finishedCond represents whether the updateRun has completed, succeeded or failed. On above line, it could be either succeeded or failed.

pkg/controllers/updaterun/controller.go
```diff
@@ -108,16 +112,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
+		}
+		klog.V(2).InfoS("The clusterStagedUpdateRun is validated", "clusterStagedUpdateRun", runObjRef)
 	}
 
-	// TODO(wantjian): execute the clusterStagedUpdateRun.
+	// TODO(wantjian): execute the clusterStagedUpdateRun and fix the requeue time.
 	klog.V(2).InfoS("Executing the clusterStagedUpdateRun", "clusterStagedUpdateRun", runObjRef, "updatingStageIndex", updatingStageIndex,
 		"toBeUpdatedBindings count", len(toBeUpdatedBindings), "toBeDeletedBindings count", len(toBeDeletedBindings))
-	return runtime.Result{}, nil
+	return runtime.Result{RequeueAfter: stageUpdatingWaitTime}, nil
```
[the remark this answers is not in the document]
jwtty: Yes, it will be fixed in the execution PR. I add the requeue here to trigger the validation in tests.

pkg/controllers/updaterun/validation.go
jwtty: We could say updating the ApplyStrategy during a running stagedUpdateRun is a user error. Updated, thanks.

pkg/controllers/updaterun/validation.go
jwtty: No, determinePolicySnapshot returns error if the policy is not found.

pkg/controllers/updaterun/validation.go
jwtty: Added, thanks!

pkg/controllers/updaterun/validation.go
jwtty: Thanks, updated.

pkg/controllers/updaterun/validation_integration_test.go
jwtty: Yea, good catch! Fixed.

pkg/controllers/updaterun/validation_integration_test.go
jwtty: Thanks, fixed!

pkg/controllers/updaterun/validation.go
jwtty: I actually want to remove this. With this shortcut, we are not able to verify if there's any mismatch in the stageStatus. This case should be covered in code already. If no stage has started, `updatingStageIndex == lastFinishedStageIndex == -1`.
```
// If no stage is updating, continue from the last finished stage (which will result it starting from 0).
   if updatingStageIndex == -1 {
       updatingStageIndex = lastFinishedStageIndex + 1
   }
```

pkg/controllers/updaterun/validation.go
```diff
@@ -258,6 +256,22 @@ func validateDeleteStageStatus(
 		klog.ErrorS(unexpectedErr, "Failed to find the deletionStageStatus in the clusterStagedUpdateRun", "clusterStagedUpdateRun", updateRunRef)
 		return -1, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
 	}
+
+	// Validate whether toBeDeletedBindings are a subnet of the clusters in the delete stage status.
+	// We only validate if it's a subnet because clusters can unjoin from the fleet and we allow that.
```
ryanzhang-oss, line 261: the real reason is not about cluster leave the fleet ( we don't allow that for updating ones). It's about the action on the cluster in the deleting stages is to delete the binding so they will disappear in the next round of validation.

pkg/controllers/updaterun/validation.go
ryanzhang-oss: it's a bit different though. The condition is a short cut to save any validation. Do we want that?

pkg/controllers/updaterun/validation.go
jwtty: Yea, I realized that when I write the execution code. Actually the same check is done during the deletionStage execution, so I tend to remove this redundancy here.

pkg/controllers/updaterun/validation.go
jwtty: I think it still worth going through the validation to make sure the clusters in the stages are not changed. It should be a rare case to hit this code though: Only when initialization is completed and then updating the updateRun as started failed.

ryanzhang-oss (approved):

PR #993, opened 2024-12-16
feat: override triggers rollout

zhiying-lin:

- watch override snapshots in the rollout controller

Fixes #

I have:

added e2e tests

### DIFF
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -677,7 +677,7 @@ func handleClusterResourceOverrideSnapshot(o client.Object, q workqueue.RateLimi
 	// check if it is the latest resource resourceBinding
 	isLatest, err := strconv.ParseBool(snapshot.GetLabels()[fleetv1beta1.IsLatestSnapshotLabel])
 	if err != nil {
-		klog.ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("invalid annotation value %s : %w", fleetv1beta1.IsLatestSnapshotLabel, err)),
+		klog.ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("invalid label value %s : %w", fleetv1beta1.IsLatestSnapshotLabel, err)),
 			"Resource clusterResourceOverrideSnapshot has does not have a valid islatest label", "clusterResourceOverrideSnapshot", snapshotKRef)
 		return
 	}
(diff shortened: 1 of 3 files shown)


pkg/controllers/rollout/controller.go
```diff
@@ -643,6 +663,72 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
 		Complete(r)
 }
 
+// handleClusterResourceOverrideSnapshot parse the clusterResourceOverrideSnapshot label and enqueue the CRP name associated
+// with the clusterResourceOverrideSnapshot if set.
+func handleClusterResourceOverrideSnapshot(o client.Object, q workqueue.RateLimitingInterface) {
```
ryanzhang-oss, line 668: is there a way to combine these two functions? We will have RP and CRP co exist all over the place soon.

test/e2e/placement_ro_test.go
```diff
@@ -97,6 +97,55 @@ var _ = Describe("creating resourceOverride (selecting all clusters) to override
 		want := map[string]string{roTestAnnotationKey: roTestAnnotationValue}
 		checkIfOverrideAnnotationsOnAllMemberClusters(false, want)
 	})
+
+	It("update ro attached to this CRP only and annotation value", func() {
```
ryanzhang-oss, line 101:
```suggestion
	It("update ro attached to this CRP only and change annotation value", func() {
```

test/e2e/placement_cro_test.go
```diff
@@ -93,6 +93,54 @@ var _ = Describe("creating clusterResourceOverride (selecting all clusters) to o
 		want := map[string]string{croTestAnnotationKey: croTestAnnotationValue}
 		checkIfOverrideAnnotationsOnAllMemberClusters(true, want)
 	})
+
+	It("update cro attached to this CRP only and annotation value", func() {
```
ryanzhang-oss, line 97:
```suggestion
	It("update cro attached to this CRP only and change the annotation value", func() {
```

test/e2e/placement_cro_test.go
```diff
@@ -93,6 +93,54 @@ var _ = Describe("creating clusterResourceOverride (selecting all clusters) to o
...
+		Eventually(func() error {
+			cro := &placementv1alpha1.ClusterResourceOverride{}
+			if err := hubClient.Get(ctx, types.NamespacedName{Name: croName}, cro); err != nil {
+				return err
+			}
+			cro.Spec = placementv1alpha1.ClusterResourceOverrideSpec{
+				Placement: &placementv1alpha1.PlacementRef{
+					Name: crpName, // assigned CRP name
+				},
```
ryanzhang-oss, line 106: a bit related question,  do we need to generate a new snapshot just because the placement is attached? What about when a placement is removed? The new override snapshot won't be trigger any change.

pkg/controllers/rollout/controller.go
zhiying-lin: we could, but need to rely on the if else condition to handle some cases based on the type, including logging, which hurts our readability.

For the current case, prefer to do the copy & paste.

test/e2e/placement_cro_test.go
zhiying-lin: `do we need to generate a new snapshot just because the placement is attached? `
we need, since customer would like to apply the override to the placement immediately.

`What about when a placement is removed?`
we don't need to trigger the rollout, but need to create a snapshot for debugging purpose.

pkg/controllers/rollout/controller.go
ryanzhang-oss: is there a way to use generic templating?

zhiying-lin: it may be workable, either using https://google.github.io/styleguide/go/decisions#generics or interface.

ryanzhang-oss (approved):

PR #998, opened 2024-12-20
fix: handle PickFixed CRP in eviction controller

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/apis/placement/v1alpha1/disruptionbudget_types.go
+++ b/apis/placement/v1alpha1/disruptionbudget_types.go
@@ -44,8 +44,7 @@ type PlacementDisruptionBudgetSpec struct {
 	// If a percentage is specified, Fleet will calculate the corresponding absolute values
 	// as follows:
 	// * if the linked Placement object is of the PickFixed placement type,
-	//   the percentage is against the number of clusters specified in the placement (i.e., the
-	//   length of ClusterNames field in the placement policy);
+	//   we don't perform any calculation because eviction is not allowed for PickFixed CRP.
 	// * if the linked Placement object is of the PickAll placement type, MaxUnavailable cannot
 	//   be specified since we cannot derive the total number of clusters selected.
 	// * if the linked Placement object is of the PickN placement type,
@@ -74,8 +73,7 @@ type PlacementDisruptionBudgetSpec struct {
 	// If a percentage is specified, Fleet will calculate the corresponding absolute values
 	// as follows:
 	// * if the linked Placement object is of the PickFixed placement type,
-	//   the percentage is against the number of clusters specified in the placement (i.e., the
-	//   length of ClusterNames field in the placement policy);
+	//   we don't perform any calculation because eviction is not allowed for PickFixed CRP.
 	// * if the linked Placement object is of the PickAll placement type, MinAvailable can be
 	//   specified but only as an integer since we cannot derive the total number of clusters selected.
 	// * if the linked Placement object is of the PickN placement type,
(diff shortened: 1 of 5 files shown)


pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -277,13 +283,9 @@ func isEvictionAllowed(bindings []placementv1beta1.ClusterResourceBinding, crp p
...
+	// we don't know the desired bindings for PickAll and we won't evict a binding for PickFixed CRP.
+	if placementType == placementv1beta1.PickNPlacementType {
 		desiredBindings = int(*crp.Spec.Policy.NumberOfClusters)
-	case placementv1beta1.PickFixedPlacementType:
-		desiredBindings = len(crp.Spec.Policy.ClusterNames)
-	default:
-		// we don't know the desired bindings for PickAll.
 	}
 
 	var disruptionsAllowed int
```
ryanzhang-oss, line 291: I can't seem to find where  "For PickAll CRPs, MaxUnavailable won't be specified in DB" is enforced?

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -112,6 +113,11 @@ func (r *Reconciler) validateEviction(ctx context.Context, eviction *placementv1
 		markEvictionInvalid(eviction, evictionInvalidDeletingCRPMessage)
 		return validationResult, nil
 	}
+	if crp.Spec.Policy.PlacementType == placementv1beta1.PickFixedPlacementType {
```
ryanzhang-oss, line 116: does this also mean the PDB API comment should be modified?

pkg/controllers/clusterresourceplacementeviction/controller.go
Arvindthiru: https://github.com/Azure/fleet/blob/main/pkg/controllers/clusterresourceplacementeviction/controller.go#L195-L201

ryanzhang-oss: so we rejected any eviction associated with the PDB instead of rejecting the PBD itself

apis/placement/v1alpha1/disruptionbudget_types.go
```diff
@@ -44,8 +44,7 @@ type PlacementDisruptionBudgetSpec struct {
 	// If a percentage is specified, Fleet will calculate the corresponding absolute values
 	// as follows:
 	// * if the linked Placement object is of the PickFixed placement type,
-	//   the percentage is against the number of clusters specified in the placement (i.e., the
-	//   length of ClusterNames field in the placement policy);
+	//   we don't perform any calculation because eviction is not allowed for PickFixed CRP.
```
ryanzhang-oss, line 47: what about pickAll?

pkg/controllers/clusterresourceplacementeviction/controller.go
Arvindthiru: We use PDB as an information source as of now, even if the CRP targeted by PDB is missing we allow the PDB to exist. So in the case of,

1. PickAll CRP where the PDB specifies MaxUnavailable or MinAvailable as percentage we say the eviction cannot be executed due to a misconfigured PDB
2. In the case for PickFixed CRP, the eviction fails on validation, and we mark eviction as invalid

apis/placement/v1alpha1/disruptionbudget_types.go
Arvindthiru: PickAll API comments

For MaxUnavailable: https://github.com/Azure/fleet/blob/main/apis/placement/v1alpha1/disruptionbudget_types.go#L49-L50

For MinAvailable:
https://github.com/Azure/fleet/blob/main/apis/placement/v1alpha1/disruptionbudget_types.go#L79-L80

ryanzhang-oss (approved):

PR #999, opened 2024-12-20
feat: add PodDisruptionBudget trackability

britaniar:

Fixes #

I have: added PDB trackability and unit test


pkg/controllers/work/apply_controller.go
```diff
@@ -463,6 +468,24 @@ func trackResourceAvailability(gvr schema.GroupVersionResource, curObj *unstruct
...
 
+func trackPDBAvailability(curObj *unstructured.Unstructured) (ApplyAction, error) {
+	var pdb policyv1.PodDisruptionBudget
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(curObj.Object, &pdb); err != nil {
+		return errorApplyAction, controller.NewUnexpectedBehaviorError(err)
+	}
+	// Check if conditions are up-to-date
+	if poddisruptionbudget.ConditionsAreUpToDate(&pdb) {
+		if cond := meta.FindStatusCondition(pdb.Status.Conditions, policyv1.DisruptionAllowedCondition); cond != nil {
+			if cond.Status == metav1.ConditionTrue && cond.Reason == policyv1.SufficientPodsReason {
```
Arvindthiru, line 479: https://github.com/kubernetes/kubernetes/blob/35f584187a6d1250191aa24b0dcf735350f57508/staging/src/k8s.io/api/policy/v1/types.go#L140 I think InsufficientPods reason should also be accounted for, since a PDB can block an eviction and in this case it's still available

pkg/controllers/work/apply_controller.go
```diff
@@ -463,6 +468,24 @@ func trackResourceAvailability(gvr schema.GroupVersionResource, curObj *unstruct
 	}
 }
 
+func trackPDBAvailability(curObj *unstructured.Unstructured) (ApplyAction, error) {
+	var pdb policyv1.PodDisruptionBudget
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(curObj.Object, &pdb); err != nil {
+		return errorApplyAction, controller.NewUnexpectedBehaviorError(err)
+	}
+	// Check if conditions are up-to-date
+	if poddisruptionbudget.ConditionsAreUpToDate(&pdb) {
```
ryanzhang-oss, line 477: I wonder why this alone is not enough? We just want to make sure that the PDB is taken into effect, we don't care if it blocks eviction or not.

PR #1000, opened 2024-12-23
feat: implement clusterStagedUpdateRun execution

jwtty:
Implement clusterStagedUpdateRun execution. Also made some refactoring.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
@@ -40,10 +41,6 @@ var (
 	// errInitializedFailed is the error when the ClusterStagedUpdateRun fails to initialize.
 	// It is a wrapped error of errStagedUpdatedAborted, because some initialization functions are reused in the validation step.
 	errInitializedFailed = fmt.Errorf("%w: failed to initialize the clusterStagedUpdateRun", errStagedUpdatedAborted)
-
-	// stageUpdatingWaitTime is the time to wait before rechecking the stage update status.
-	// Put it as a variable for convenient testing.
-	stageUpdatingWaitTime = 60 * time.Second
 )
 
 // Reconciler reconciles a ClusterStagedUpdateRun object.
@@ -127,10 +124,35 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		klog.V(2).InfoS("The clusterStagedUpdateRun is validated", "clusterStagedUpdateRun", runObjRef)
 	}
 
-	// TODO(wantjian): execute the clusterStagedUpdateRun and fix the requeue time.
-	klog.V(2).InfoS("Executing the clusterStagedUpdateRun", "clusterStagedUpdateRun", runObjRef, "updatingStageIndex", updatingStageIndex,
-		"toBeUpdatedBindings count", len(toBeUpdatedBindings), "toBeDeletedBindings count", len(toBeDeletedBindings))
-	return runtime.Result{RequeueAfter: stageUpdatingWaitTime}, nil
+	// The previous run is completed but the update to the status failed.
+	if updatingStageIndex == -1 {
+		klog.V(2).InfoS("The clusterStagedUpdateRun is completed", "clusterStagedUpdateRun", runObjRef)
+		return runtime.Result{}, r.recordUpdateRunSucceeded(ctx, &updateRun)
+	}
+
+	// Execute the updateRun.
+	klog.V(2).InfoS("Continue to execute the clusterStagedUpdateRun", "updatingStageIndex", updatingStageIndex, "clusterStagedUpdateRun", runObjRef)
+	finished, waitTime, execErr := r.execute(ctx, &updateRun, updatingStageIndex, toBeUpdatedBindings, toBeDeletedBindings)
+	if execErr != nil && errors.Is(execErr, errStagedUpdatedAborted) {
+		// errStagedUpdatedAborted cannot be retried.
+		return runtime.Result{}, r.recordUpdateRunFailed(ctx, &updateRun, execErr.Error())
+	}
+
+	if finished {
+		klog.V(2).InfoS("The clusterStagedUpdateRun is completed", "clusterStagedUpdateRun", runObjRef)
+		return runtime.Result{}, r.recordUpdateRunSucceeded(ctx, &updateRun)
+	}
+
+	// The execution is not finished yet or it encounters a retriable error.
+	// We need to record the status and requeue.
+	if updateErr := r.recordUpdateRunStatus(ctx, &updateRun); updateErr != nil {
+		return runtime.Result{}, updateErr
+	}
+	klog.V(2).InfoS("The clusterStagedUpdateRun is not finished yet", "requeueWaitTime", waitTime, "execErr", execErr, "clusterStagedUpdateRun", runObjRef)
+	if execErr != nil {
+		return runtime.Result{}, execErr
+	}
+	return runtime.Result{Requeue: true, RequeueAfter: waitTime}, nil
 }
 
 // handleDelete handles the deletion of the clusterStagedUpdateRun object.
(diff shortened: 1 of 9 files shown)


pkg/controllers/updaterun/execution.go
```diff
@@ -0,0 +1,482 @@
...
+				// The cluster is marked as deleting but the binding is not deleting.
+				unexpectedErr := controller.NewUnexpectedBehaviorError(fmt.Errorf("the cluster `%s` in the deleting stage is marked as deleting but its corresponding binding is not deleting", curCluster.ClusterName))
+				klog.ErrorS(unexpectedErr, "The binding should be deleting before we mark a cluster deleting", "clusterStatus", curCluster, "clusterStagedUpdateRun", updateRunRef)
+				return false, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
+			}
+			deletingBinding++
+			continue
+		}
+		// The cluster status is not deleting yet
+		if err := r.Client.Delete(ctx, binding); err != nil {
```
jwtty, line 215: Compared with the original implementation, I did not check if binding is up-to-date before deleting. I don't think it's necessary. Do we want that?

pkg/controllers/updaterun/validation.go
```diff
@@ -305,30 +304,6 @@ func validateDeleteStageStatus(
...
-	// The delete stage is still updating.
-	if condition.IsConditionStatusTrue(deleteStageProgressingCond, updateRun.Generation) {
-		klog.InfoS("The delete stage is updating", "clusterStagedUpdateRun", updateRunRef)
-		return totalStages, nil
-	}
-	// All stages have finished, but the delete stage is not active or finished.
-	unexpectedErr := controller.NewUnexpectedBehaviorError(fmt.Errorf("the delete stage is not active, but all stages finished"))
-	klog.ErrorS(unexpectedErr, "There is no stage active", "clusterStagedUpdateRun", updateRunRef)
-	return -1, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
-}
```
ryanzhang-oss, line 317: why delete this check?

pkg/controllers/updaterun/execution_integration_test.go
```diff
@@ -0,0 +1,499 @@
...
+			By("Validating the 1st cluster has succeeded and 2nd cluster has started")
+			wantStatus.StagesStatus[0].Clusters[0].Conditions = append(wantStatus.StagesStatus[0].Clusters[0].Conditions, generateTrueCondition(updateRun, placementv1alpha1.ClusterUpdatingConditionSucceeded))
+			wantStatus.StagesStatus[0].Clusters[1].Conditions = append(wantStatus.StagesStatus[0].Clusters[1].Conditions, generateTrueCondition(updateRun, placementv1alpha1.ClusterUpdatingConditionStarted))
+			validateClusterStagedUpdateRunStatus(ctx, updateRun, wantStatus, "")
+
+			By("Validating the 1st stage has startTime set")
+			Expect(updateRun.Status.StagesStatus[0].StartTime).ShouldNot(BeNil())
+		})
+
+		It("Should mark the 2nd cluster in the 1st stage as succeeded", func() {
```
ryanzhang-oss, line 203:
```suggestion
		It("Should mark the 2nd cluster in the 1st stage as succeeded after marking the binding available", func() {
```

pkg/controllers/updaterun/validation.go
jwtty: After the last updating stage completes, execution does not mark the deletion stage as started. Instead, it returns a 0 wait-time and moves on to the delete stage in the next reconcile loop, just like finishing a regular updating stage and moving on to the next updating stage. With this check, the validation in the next reconcile would fail because the deletion stage is not marked as started yet.

pkg/controllers/updaterun/execution.go
```diff
@@ -0,0 +1,482 @@
...
+				klog.V(2).InfoS("Found the first binding that is updating but the cluster status has not been updated", "cluster", clusterStatus.ClusterName, "stage", updatingStageStatus.StageName, "clusterStagedUpdateRun", updateRunRef)
+				if binding.Spec.State != placementv1beta1.BindingStateBound {
+					binding.Spec.State = placementv1beta1.BindingStateBound
+					if err := r.Client.Update(ctx, binding); err != nil {
+						klog.ErrorS(err, "Failed to update a binding to be bound", "binding", klog.KObj(binding), "cluster", clusterStatus.ClusterName, "stage", updatingStageStatus.StageName, "clusterStagedUpdateRun", updateRunRef)
+						return 0, controller.NewUpdateIgnoreConflictError(err)
+					}
+					klog.V(2).InfoS("Updated the status of a binding to bound", "binding", klog.KObj(binding), "cluster", clusterStatus.ClusterName, "stage", updatingStageStatus.StageName, "clusterStagedUpdateRun", updateRunRef)
+				} else {
+					if _, updateErr := checkClusterUpdateResult(binding, clusterStatus, updatingStageStatus, updateRun); updateErr != nil {
```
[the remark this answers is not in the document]
jwtty: For now, yes. Because we currently do not identify any termination errors for cluster updates. I have a TODO in this function. But even in the future when we can identify some, I'll return an `errExecutionAborted` inside the `checkClusterUpdateResult` function and can still return the `updateErr` directly.

pkg/controllers/updaterun/execution.go
```diff
@@ -0,0 +1,482 @@
...
+	// clusterUpdatingWaitTime is the time to wait before rechecking the cluster update status.
+	// Put it as a variable for convenient testing.
+	clusterUpdatingWaitTime = 15 * time.Second
+
+	// stageUpdatingWaitTime is the time to wait before rechecking the stage update status.
+	// Put it as a variable for convenient testing.
+	stageUpdatingWaitTime = 60 * time.Second
+)
+
+// execute executes the update run by updating the clusters in the updating stage specified by updatingStageIndex.
```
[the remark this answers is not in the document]
jwtty: Thanks, updated.

pkg/controllers/updaterun/controller.go
```diff
@@ -127,10 +124,35 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		klog.V(2).InfoS("The clusterStagedUpdateRun is validated", "clusterStagedUpdateRun", runObjRef)
 	}
 
-	// TODO(wantjian): execute the clusterStagedUpdateRun and fix the requeue time.
-	klog.V(2).InfoS("Executing the clusterStagedUpdateRun", "clusterStagedUpdateRun", runObjRef, "updatingStageIndex", updatingStageIndex,
-		"toBeUpdatedBindings count", len(toBeUpdatedBindings), "toBeDeletedBindings count", len(toBeDeletedBindings))
-	return runtime.Result{RequeueAfter: stageUpdatingWaitTime}, nil
+	// The previous run is completed but the update to the status failed.
```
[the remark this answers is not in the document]
jwtty: The check I removed is about when all the updating stages finish and the deletion stage is about to start. In that case, `updatingStageIndex == len(updatingStages)`. When `updatingStageIndex == -1`, it's always when all the stages including the deletion have finished but the `recordUpdateRunSucceeded` on line 143 somehow fails.

pkg/controllers/updaterun/execution.go
jwtty: In case of both resource changes and scheduling policy changes, the binding to be deleted may not have up-to-date resources. If there are concurrent update runs, the same binding could be deleted by either updateRun. If there's a new scheduling policy change that marks this to-be-deleted binding as scheduled again, this updateRun can fail during validation. Or if there's chance that the change happens right after the validation passes and before deleting the bindings, I think I can add a check to make sure the binding's state is unScheduled or report unexpected error if not.

jwtty: I do have a concern: in rollout controller, we make sure to complete the deletion of the bindings on a cluster before scheduling a newly created binding on the same cluster: https://github.com/Azure/fleet/blob/main/pkg/controllers/rollout/controller.go#L226. I wonder if we should have similar logic in the updateRun controller too.

ryanzhang-oss (approved):

pkg/controllers/updaterun/execution.go
ryanzhang-oss: the concurrent updateRun is not valid if the schedule results are different. This is only for the case one needs to release different versions of the resources so the bindings should all be scheduled or bound.

ryanzhang-oss: As for the check no duplicate logic, should the check at https://github.com/Azure/fleet/blob/cb0c5f70b850322a221a70c18d51008ffe43f447/pkg/controllers/updaterun/initialization.go#L177 be suffice?

pkg/controllers/updaterun/controller.go
```diff
@@ -127,10 +124,35 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
+	if updatingStageIndex == -1 {
+		klog.V(2).InfoS("The clusterStagedUpdateRun is completed", "clusterStagedUpdateRun", runObjRef)
+		return runtime.Result{}, r.recordUpdateRunSucceeded(ctx, &updateRun)
+	}
+
+	// Execute the updateRun.
+	klog.V(2).InfoS("Continue to execute the clusterStagedUpdateRun", "updatingStageIndex", updatingStageIndex, "clusterStagedUpdateRun", runObjRef)
+	finished, waitTime, execErr := r.execute(ctx, &updateRun, updatingStageIndex, toBeUpdatedBindings, toBeDeletedBindings)
+	if execErr != nil && errors.Is(execErr, errStagedUpdatedAborted) {
```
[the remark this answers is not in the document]
jwtty: yea it's enough. Will update in the next PR. Thanks!

pkg/controllers/updaterun/execution.go
```diff
@@ -0,0 +1,484 @@
...
+		return false
+	}
+	if !reflect.DeepEqual(binding.Spec.ApplyStrategy, updateRun.Status.ApplyStrategy) {
+		klog.ErrorS(fmt.Errorf("binding has different applyStrategy, want: %v, got: %v", updateRun.Status.ApplyStrategy, binding.Spec.ApplyStrategy), "ClusterResourceBinding is not up-to-date", "clusterResourceBinding", klog.KObj(binding), "clusterStagedUpdateRun", klog.KObj(updateRun))
+		return false
+	}
+	return true
+}
+
+// checkClusterUpdateResult checks if the cluster has been updated successfully.
```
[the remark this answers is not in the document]
jwtty: Will update in next PR. Thanks!

pkg/controllers/updaterun/execution.go
```diff
@@ -0,0 +1,484 @@
...
+// It returns if the cluster has been updated successfully and the error if the cluster update failed.
+func checkClusterUpdateResult(
```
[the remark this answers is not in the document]
jwtty: sure, will update in the next PR.

pkg/controllers/updaterun/execution.go
jwtty: _"the concurrent updateRun is not valid if the schedule results are different. This is only for the case one needs to release different versions of the resources so the bindings should all be scheduled or bound."_

In this case, the bindings won't be marked as to-be-deleted because they both have the latest scheduling snapshot index.

PR #1004, opened 2025-01-01
feat: enable clusterStagedUpdateRun controller

jwtty:
Add the flag and conditionally enable the `clusterStagedUpdateRun` controller.

Fix a bug found during testing: the updateRun controller need to add `rolloutStarted` condition to binding too for the workgenerator to continue.
 
Also add an additional check during initialization to make sure a binding with old policyschedulingsnapshot is expected to be in `Unscheduled` state.

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/execution.go
+++ b/pkg/controllers/updaterun/execution.go
@@ -115,6 +115,9 @@ func (r *Reconciler) executeUpdatingStage(
 					return 0, controller.NewUpdateIgnoreConflictError(err)
 				}
 				klog.V(2).InfoS("Updated the status of a binding to bound", "binding", klog.KObj(binding), "cluster", clusterStatus.ClusterName, "stage", updatingStageStatus.StageName, "clusterStagedUpdateRun", updateRunRef)
+				if err := r.updateBindingRolloutStarted(ctx, binding); err != nil {
+					return 0, err
+				}
 			} else {
 				klog.V(2).InfoS("Found the first binding that is updating but the cluster status has not been updated", "cluster", clusterStatus.ClusterName, "stage", updatingStageStatus.StageName, "clusterStagedUpdateRun", updateRunRef)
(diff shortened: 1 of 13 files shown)


pkg/controllers/updaterun/execution.go
```diff
@@ -314,6 +327,24 @@ func (r *Reconciler) checkAfterStageTasksStatus(ctx context.Context, updatingSta
...
 }
 
+// updateBindingRolloutStarted updates the binding status to indicate the rollout has started.
+func (r *Reconciler) updateBindingRolloutStarted(ctx context.Context, binding *placementv1beta1.ClusterResourceBinding) error {
+	cond := metav1.Condition{
+		Type:               string(placementv1beta1.ResourceBindingRolloutStarted),
+		Status:             metav1.ConditionTrue,
+		ObservedGeneration: binding.Generation,
+		Reason:             condition.RolloutStartedReason,
+		Message:            "Detected the new changes on the resources and started the rollout process",
```
ryanzhang-oss, line 337: probably more informational with the resourceSnapShot and updatedRun name?

pkg/controllers/updaterun/execution.go
```diff
@@ -314,6 +327,24 @@ func (r *Reconciler) checkAfterStageTasksStatus(ctx context.Context, updatingSta
...
+	}
+	binding.SetConditions(cond)
```
ryanzhang-oss, line 339: do you care about "LastTransitionTime"?  as the binding status is most likely already True so it will not be updated.

pkg/controllers/updaterun/execution.go
jwtty: thanks, updated.

pkg/controllers/updaterun/execution.go
jwtty: Good catch! I didn't notice this during test. Fixed, thanks!

ryanzhang-oss (approved):

PR #1010, opened 2025-01-07
feat: drift detection and takeover implementation (1/ contd) 

michaelawyu:

This PR addresses the leftover comments on #985.

I have:

- [x] Unit tests
- [x] Integration tests

N/A


pkg/controllers/workapplier/apply.go
```diff
@@ -233,6 +240,7 @@ func (r *Reconciler) threeWayMergePatch(
 		Resource(*gvr).Namespace(manifestObj.GetNamespace()).
 		Patch(ctx, manifestObj.GetName(), patch.Type(), data, patchOpts)
 	if err != nil {
+		_ = controller.NewAPIServerError(false, err)
```
ryanzhang-oss, line 243: Is this just to print out an unexpected error as the API server error is not returned. I was thinking of seeing this on the work status so the rollout can decide if it needs to keep retry or not

michaelawyu: Done 👌

pkg/controllers/workapplier/apply.go
```diff
@@ -161,8 +162,11 @@ func (r *Reconciler) apply(
 			applyStrategy.ServerSideApplyConfig.ForceConflicts, isOptimisticLockEnabled, false,
 		)
 	default:
-		// An unexpected apply strategy has been set.
-		return nil, fmt.Errorf("unexpected apply strategy %s is found", applyStrategy.Type)
+		// An unexpected apply strategy has been set. Normally this will never run as the built-in
+		// validation would block invalid values.
+		wrappedErr := fmt.Errorf("unexpected apply strategy %s is found", applyStrategy.Type)
+		_ = controller.NewUnexpectedBehaviorError(wrappedErr)
+		return nil, wrappedErr
```
ryanzhang-oss, line 169: I wonder why don't we return the unexpectedError here but return APIServer errors in other places?

ryanzhang-oss: is that UnExpected erorr are "unExpected" so we don't want to record?

michaelawyu: Hi Ryan! It's still logged; the thing about this one is that the returned error will be formatted into a condition message which the user would be able to see, and the `unexpected behavior cannot be handled by the controller` part might read a bit weird; do we still want to expose them?

michaelawyu: I will make the changes if that is OK with you.

ryanzhang-oss: I can see pros and cons on either solution

ryanzhang-oss (approved):

PR #1008, opened 2025-01-06
interface: improve ClusterStagedUpdateRun APIs

jwtty:
### Description of your changes
1. Add additionalPrinterColumns for `clusterStagedUpdateRun` and `clusterApprovalRequest` objects.
2. Add validations in updateRun name length and stageName to ensure ApprovalRequest name valid .
3. Update the examples to use the v1beta1 APIs.

The `get` command outputs now look like:
```
kubectl get clusterstagedupdaterun
NAME          PLACEMENT           RESOURCE-SNAPSHOT               POLICY-SNAPSHOT        INITIALIZED   SUCCEEDED   AGE
example-run   example-placement   example-placement-0-snapshot   example-placement-7   True          True        13m

kubectl get clsuterstagedupdaterun -owide
NAME          PLACEMENT           RESOURCE-SNAPSHOT               STRATEGY           POLICY-SNAPSHOT        INITIALIZED   SUCCEEDED   AGE
example-run   example-placement   example-placement-0-snapshot   example-strategy   example-placement-7   True          True        3m47s

kubectl get clusterapprovalrequests
NAME                     UPDATE-RUN     STAGE        APPROVED   AGE
example-run-canary       example-run   canary       True       12m
example-run-production   example-run   production   True       11m
```

Fixes #

I have:

- [x] Run `make reviewable` to ensure this PR is ready for review.

PR #1014, opened 2025-01-10
fix: fix stagedUpdateRun PolicyObservedClusterCount

jwtty:
1. When scheduling policy is PickAll, we update PolicyObservedClusterCount to the number of scheduled/bound clusterResourceBindings, instead of -1. For PickFix and PickN, we validate if the number of bindings match the expected count.
2. Add more output in one of the integration test to help debugging UT flakes.

Fixes #

I have:


ryanzhang-oss (commented):
as we discussed off line, we just need to add something like this to the updateRun object

// +kubebuilder:validation:XValidation:rule="size(self.metadata.name) < 128",message="metadata.name max length is 127"

jwtty:
> as we discussed off line, we just need to add something like this to the updateRun object
> 
> // +kubebuilder:validation:XValidation:rule="size(self.metadata.name) < 128",message="metadata.name max length is 127"

Since we only need to add more API validations, I'll make the change in the API change PR: #1008

ryanzhang-oss (approved):

PR #1016, opened 2025-01-13
feat: enable work generator to process apply strategy updates + DiffReported condition

michaelawyu:

This PR:

a) fixes the issue where work generator will not refresh apply strategies; and
b) enables work generator to process `DiffReported` condition

I have:

- [x] Unit tests
- [x] Integration tests

E2E tests can only be added when the new work applier is enabled.


pkg/controllers/workgenerator/controller.go
```diff
@@ -859,6 +896,38 @@ func buildAllWorkAppliedCondition(works map[string]*fleetv1beta1.Work, binding *
 	}
 }
 
+// buildAllWorkDiffReportedCondition builds DiffReported condition for a ClusterResourceBinding
+// by checking the DiffReported condition of all the works associated with the binding.
+func buildAllWorkDiffReportedCondition(works map[string]*fleetv1beta1.Work, binding *fleetv1beta1.ClusterResourceBinding) metav1.Condition {
```
ryanzhang-oss, line 901: looks like only the binding.GetGeneration() is needed?

pkg/controllers/workgenerator/controller.go
```diff
@@ -757,9 +762,38 @@ func getWorkNamePrefixFromSnapshotName(resourceSnapshot *fleetv1beta1.ClusterRes
...
+
 	// try to gather the resource binding applied status if we didn't update any associated work spec this time
 	appliedCond := buildAllWorkAppliedCondition(works, resourceBinding)
 	resourceBinding.SetConditions(appliedCond)
+
+	// Set the DiffReported condition if (and only if) a ReportDiff apply strategy is currently
+	// being used.
+	if resourceBinding.Spec.ApplyStrategy != nil && resourceBinding.Spec.ApplyStrategy.Type == fleetv1beta1.ApplyStrategyTypeReportDiff {
+		diffReportedCond := buildAllWorkDiffReportedCondition(works, resourceBinding)
+		resourceBinding.SetConditions(diffReportedCond)
```
ryanzhang-oss, line 794: why not set the condition in the function itself instead of return the value and then set it?

pkg/controllers/workgenerator/controller.go
```diff
@@ -850,6 +884,9 @@ func buildAllWorkAppliedCondition(works map[string]*fleetv1beta1.Work, binding *
 			ObservedGeneration: binding.GetGeneration(),
 		}
 	}
+	// In the case where ReportDiff apply strategy is used, work applier will no longer update
+	// Applied condition on Work objects; since existing Applied conditions will all become stale,
+	// this function will return a False Applied condition.
```
ryanzhang-oss, line 889: need an e2e test to verify and safe guard this behavior.

pkg/controllers/workgenerator/controller.go
michaelawyu: Hi Ryan! The function was styled in consistence with the other two existing functions for similar purposes, `buildAllWorkAppliedCondition` and `buildAllWorkAvailableCondition`; I will update all three of them as you suggested.

pkg/controllers/workgenerator/controller.go
michaelawyu: Same as the comment above.

pkg/controllers/workgenerator/controller.go
michaelawyu: Yeah, will add one in the later PR.

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -2211,6 +2658,85 @@ func verifyBindStatusAvailableWithOnePlacement(binding *placementv1beta1.Cluster
...
+
+		gotApplyStrategy := work.Spec.ApplyStrategy
+		if diff := cmp.Diff(gotApplyStrategy, wantApplyStrategy); diff != "" {
+			return fmt.Errorf("work apply strategy mismatches (-got, +want):\n%s", diff)
+		}
+		return nil
+	}
+}
+
+func bindingStatusUpdatedActual(
```
ryanzhang-oss, line 2677: looks like this can be used to replace a bunch of "verifyBindStatusXXX" function? If so, please add a TODO

ryanzhang-oss: although the full comparison mode looks more verbose

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -1636,7 +1650,440 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+
+			resourceBindingSpec := placementv1beta1.ResourceBindingSpec{
+				State:                placementv1beta1.BindingStateBound,
+				ResourceSnapshotName: masterResourceSnapshot.Name,
+				TargetCluster:        memberClusterName,
+			}
+			createClusterResourceBinding(&clusterResourceBinding, resourceBindingSpec)
+		})
+
+		It("should create work object with expected apply strategy", func() {
```
ryanzhang-oss, line 1693:
```suggestion
		It("should create work object with no apply strategy", func() {
```

ryanzhang-oss: when does the binding have no applyStrategy?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -1636,7 +1650,440 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+			work := retrieveWork(memberClusterNamespaceName, fmt.Sprintf(placementv1beta1.FirstWorkNameFmt, testCRPName))
+
+			diffReportedCond := &metav1.Condition{
+				Type:               string(placementv1beta1.WorkConditionTypeDiffReported),
+				Status:             metav1.ConditionTrue,
+				Reason:             string(workapplier.ManifestProcessingReportDiffResultTypeNoDiffFound),
+				ObservedGeneration: work.GetGeneration(),
+			}
+			refreshWorkStatus(work, nil, nil, diffReportedCond, nil)
+		})
```
ryanzhang-oss, line 1829: we won't need this (which simulated a behavior of another controller) if we explicitly set the applied to be false ourself.

pkg/controllers/workgenerator/controller_integration_test.go
michaelawyu: Hi Ryan! `ApplyStrategy` is a optional pointer field, so a `nil` apply strategy can be quite common.

michaelawyu: Edited the It spec name as suggested.

pkg/controllers/workgenerator/controller_integration_test.go
michaelawyu: I will add a TO-DO comment and think about the best way to refactor the code 👌.

pkg/utils/common.go
```diff
@@ -607,6 +607,11 @@ var LessFuncDiffedResourcePlacements = func(a, b placementv1beta1.DiffedResource
 	return aStr < bStr
 }
 
+// LessFuncCondition is a less function for sorting conditions based on its types.
+var LessFuncConditions = func(a, b metav1.Condition) bool {
```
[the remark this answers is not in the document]
michaelawyu: Sure, will do 👌

pkg/controllers/workgenerator/controller.go
```diff
@@ -700,15 +703,16 @@ func (r *Reconciler) upsertWork(ctx context.Context, newWork, existingWork *flee
 	} else {
 		// we already checked the label in fetchAllResourceSnapShots function so no need to check again
 		resourceIndex, _ := labels.ExtractResourceIndexFromClusterResourceSnapshot(resourceSnapshot)
-		if workResourceIndex == resourceIndex {
-			// no need to do anything if the work is generated from the same resource/override snapshots
+		if workResourceIndex == resourceIndex && equality.Semantic.DeepEqual(existingWork.Spec.ApplyStrategy, newWork.Spec.ApplyStrategy) {
```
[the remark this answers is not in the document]
michaelawyu: Ah, this is actually a tricky one; to avoid apply strategy inconsistencies the refresh has to be handled before the check. I will some edits

pkg/controllers/workgenerator/controller_integration_test.go
michaelawyu: Hi Ryan! Part of the code tests the status populator, which is why the condition is being set (summarizing all `DiffReported` conditions from work objects).

pkg/controllers/workgenerator/controller.go
```diff
@@ -382,6 +385,32 @@ func (r *Reconciler) listAllWorksAssociated(ctx context.Context, resourceBinding
...
+	// strategy update.
+	errs, cctx := errgroup.WithContext(ctx)
+	for workName := range existingWorks {
+		w := existingWorks[workName]
+		errs.Go(func() error {
+			updated, err := r.syncApplyStrategy(ctx, resourceBinding, w)
+			if err != nil {
+				return err
+			}
+			if updated {
```
ryanzhang-oss, line 404: this will revert the applied condition on the binding. It seems that it's fine but please think through.

pkg/controllers/workgenerator/controller.go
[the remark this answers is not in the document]
ryanzhang-oss: do we still need this after we did the update of the applyStrategy first regardless?

pkg/controllers/workgenerator/controller.go
```diff
@@ -700,15 +751,16 @@ func (r *Reconciler) upsertWork(ctx context.Context, newWork, existingWork *flee
...
-		if workResourceIndex == resourceIndex {
-			// no need to do anything if the work is generated from the same resource/override snapshots
+		if workResourceIndex == resourceIndex && equality.Semantic.DeepEqual(existingWork.Spec.ApplyStrategy, newWork.Spec.ApplyStrategy) {
+			// no need to do anything if the work is generated from the same resource/override snapshots,
+			// and the apply strategy remains the same.
 			if existingWork.Annotations[fleetv1beta1.ParentResourceOverrideSnapshotHashAnnotation] == newWork.Annotations[fleetv1beta1.ParentResourceOverrideSnapshotHashAnnotation] &&
 				existingWork.Annotations[fleetv1beta1.ParentClusterResourceOverrideSnapshotHashAnnotation] == newWork.Annotations[fleetv1beta1.ParentClusterResourceOverrideSnapshotHashAnnotation] {
-				klog.V(2).InfoS("Work is associated with the desired resource/override snapshots", "existingROHash", existingWork.Annotations[fleetv1beta1.ParentResourceOverrideSnapshotHashAnnotation],
+				klog.V(2).InfoS("Work is associated with the desired resource/override snapshots and apply strategy", "existingROHash", existingWork.Annotations[fleetv1beta1.ParentResourceOverrideSnapshotHashAnnotation],
```
ryanzhang-oss, line 759: ditto

pkg/controllers/workgenerator/controller.go
```diff
@@ -850,6 +931,9 @@ func buildAllWorkAppliedCondition(works map[string]*fleetv1beta1.Work, binding *
 			ObservedGeneration: binding.GetGeneration(),
 		}
 	}
+	// In the case where ReportDiff apply strategy is used, work applier will no longer update
+	// Applied condition on Work objects; since existing Applied conditions will all become stale,
+	// this function will return a False Applied condition.
```
ryanzhang-oss, line 936: do we still need this comment?

pkg/controllers/workgenerator/controller.go
```diff
@@ -757,9 +810,37 @@ func getWorkNamePrefixFromSnapshotName(resourceSnapshot *fleetv1beta1.ClusterRes
...
+	// Note (chenyu1): the work generator will refresh the status of a ClusterResourceBinding with
+	// the following logic:
+	//
+	// a) If the currently active apply strategy (as dictated by the ClusterResourceBinding spec)
+	//    is ClientSideApply or ServerSideApply, the work generator will update the Applied and
+	//    Available conditions (plus the details about failed, diffed, and/or drifted placements)
+	//    in the status, as appropriate; the DiffReported condition will not be updated.
+	// b) If the currently active apply strategy is ReportDiff, the work generator will update
+	//    the DiffReported condition in the status, plus the details about diffed placements;
+	//    the Applied condition will always be set to False, despite the fact that the work applier
```
ryanzhang-oss, line 823: where is " the Applied condition will always be set to False" done explicitly?

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -2908,6 +3079,94 @@ func TestUpdateBindingStatusWithRetry(t *testing.T) {
...
+	workName := "test-work-1"
+
+	testCases := []struct {
+		name              string
+		resourceBinding   *fleetv1beta1.ClusterResourceBinding
+		work              *fleetv1beta1.Work
+		wantUpdated       bool
+		wantApplyStrategy *fleetv1beta1.ApplyStrategy
+	}{
+		{
```
ryanzhang-oss, line 3093: add a diff test case? the work has something but the binding has something else?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -1636,7 +1650,601 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+				}
+				return nil
+			}, timeout, interval).Should(Succeed(), "Failed to update apply strategy on the resource binding")
+		})
+
+		It("can add the rollout started condition to the resource binding", func() {
+			updateRolloutStartedGeneration(&clusterResourceBinding)
+		})
+
+		It("should update apply strategy on the work object", func() {
```
ryanzhang-oss, line 1726: why not continue to check the binding status too?

pkg/controllers/workgenerator/controller_integration_test.go
```diff
@@ -1636,7 +1650,601 @@ var _ = Describe("Test Work Generator Controller", func() {
...
+				ComparisonOption: placementv1beta1.ComparisonOptionTypeFullComparison,
+				WhenToApply:      placementv1beta1.WhenToApplyTypeIfNotDrifted,
+				Type:             placementv1beta1.ApplyStrategyTypeServerSideApply,
+				WhenToTakeOver:   placementv1beta1.WhenToTakeOverTypeNever,
+			}
+			applyStrategyUpdatedActual := workApplyStrategyUpdatedActual(memberClusterNamespaceName, workName, wantApplyStrategy)
+			Eventually(applyStrategyUpdatedActual, timeout, interval).Should(Succeed(), "Failed to create expected work object")
+		})
+
+		AfterAll(func() {
```
ryanzhang-oss, line 1738: can we not reuse this AfterAll between those ITs?

pkg/controllers/workgenerator/controller.go
```diff
@@ -382,6 +385,32 @@ func (r *Reconciler) listAllWorksAssociated(ctx context.Context, resourceBinding
...
+				updateAny.Store(true)
+			}
+			return nil
+		})
+	}
+	if updateErr := errs.Wait(); updateErr != nil {
+		return false, false, updateErr
+	}
+
 	// the hash256 function can can handle empty list https://go.dev/play/p/_4HW17fooXM
```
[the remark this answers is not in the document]
michaelawyu: Fixed 👌

pkg/controllers/workgenerator/controller.go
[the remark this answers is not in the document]
michaelawyu: I'll remove that check 👌

pkg/controllers/workgenerator/controller.go
michaelawyu: Done 👌

pkg/controllers/workgenerator/controller.go
michaelawyu: Hi Ryan! If the rollout controller has been revised to always push changes to bindings under the ReportDiff mode (either by treating all such bindings as ready ones or failed ones), I think it might make sense to not set Applied condition at all in the work generator with the ReportDiff mode (which, with the current logic, would have the Applied condition removed upon reconciliation).

pkg/controllers/workgenerator/controller.go
michaelawyu: Hi Ryan! When the apply strategy changes, the work object's generation is bumped (all existing conditions become stale). The work applier will not set the Applied condition in the ReportDiff mode, and the `buildAllWorkAppliedCondition` function in the work generator will hence set a false Applied condition

pkg/controllers/workgenerator/controller_test.go
michaelawyu: Added 👌

pkg/controllers/workgenerator/controller_integration_test.go
michaelawyu: Added the check 👌

pkg/controllers/workgenerator/controller_integration_test.go
michaelawyu: Move the logic into its own function 👌

pkg/controllers/workgenerator/controller.go
michaelawyu: Note to self: need to revise the work just updated status handling code a little and add additional test suites if a decision has been made,

pkg/controllers/workgenerator/controller.go
```diff
@@ -515,13 +544,30 @@ func (r *Reconciler) syncAllWork(ctx context.Context, resourceBinding *fleetv1be
...
+		klog.ErrorS(err, "Failed to update the apply strategy on the work", "work", klog.KObj(existingWork), "binding", klog.KObj(resourceBinding))
+		return true, controller.NewUpdateIgnoreConflictError(err)
+	}
+	klog.V(2).InfoS("Successfully updated the apply strategy on the work", "work", klog.KObj(existingWork), "binding", klog.KObj(resourceBinding))
+	return true, nil
+}
+
 // areAllWorkSynced checks if all the works are synced with the resource binding.
 func areAllWorkSynced(existingWorks map[string]*fleetv1beta1.Work, resourceBinding *fleetv1beta1.ClusterResourceBinding, _, _ string) bool {
-	syncedCondition := resourceBinding.GetCondition(string(fleetv1beta1.ResourceBindingWorkSynchronized))
```
michaelawyu, line 520: Hi Ryan! For this removal I am wondering if I could check with you to see if there's anything I missed/overlooked.

michaelawyu: There seems to be a bug with the current code: say, if there is a binding with a missing master resource snapshot, and the user issues an apply strategy change, the ideal outcome would be that the binding (and its works) receives the new apply strategy and continues to report the latest Applied/Available condition (under the stale resource snapshot). However, with this sync cond check, Fleet will always deem the work(s) to be out of sync (as the sync cond becomes stale with the apply strategy change), and the user will see an OverridenFailed error (cond) even if there are no overrides specified at all.

michaelawyu: And, another thing is, if the sync cond is True (and up to date), the works are guaranteed to be synchronized, so the check below seems redundant I assume? Instead the only case where we could skip the check below seems to be verifying if the sync cond is False and up to date, which would imply that things have already gone wrong.

michaelawyu: Apologies if I miss or misunderstand anything here 😣

ryanzhang-oss: > as the sync cond becomes stale with the apply strategy change

I think the problem is exactly this. The out of the loop update of the work before all the other logic probably breaks more than this part of the code in some corner cases. I think the original goal was to "isolate" this logic but it might have backfired.

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: > The work applier will not set the Applied condition in the ReportDiff mode,

IIRC, That's not really "explicit" but more like "by default".

> This of course would need the rollout controller to processing ReportDiff mode a bit differently, but it feels more like a clean cut

My PR is changing the Rollout controller to handle ReportDiff.

pkg/controllers/workgenerator/controller.go
michaelawyu: Hi Ryan! It might be (if a custom owner has been added before).

pkg/controllers/workgenerator/controller.go
michaelawyu: Edited after the latest discussed change.

pkg/controllers/workgenerator/controller.go
michaelawyu: Changes made after the latest round of discussions.

pkg/controllers/workgenerator/controller.go
michaelawyu: Changes made after the latest round of discussed changes.

michaelawyu:
Latest incorporated changes:
* Apply strategy is always updated
* The work generator will now populate Applied + Available condition if the apply strategy is CSA/SSA; or DiffReported condition if ReportDiff.
* Failures, drifts, diffs are being populated based on the True/False/stale status of the Applied/Available/DiffReported conditions plus the apply strategy in use to make sure that stale failures/drifts/diffs will not leak into the new round of status reporting.
* Added various new tests to verify the new behavior change.

pkg/controllers/workgenerator/controller.go
```diff
@@ -779,20 +877,81 @@ func setBindingStatus(works map[string]*fleetv1beta1.Work, resourceBinding *flee
...
+			// some of them have failed the availability check.
+			//
+			// In this case, set failed and drifted placements; no diffed placements will be set (diffs
+			// will only occur when there is an apply failure or the report diff mode is on).
+			failedManifests := extractFailedResourcePlacementsFromWork(w)
+			failedResourcePlacements = append(failedResourcePlacements, failedManifests...)
+
+			driftedManifests := extractDriftedResourcePlacementsFromWork(w)
+			driftedResourcePlacements = append(driftedResourcePlacements, driftedManifests...)
+		default:
```
ryanzhang-oss, line 945: isn't it better to explicitly have a case for that and then add an unexpected error in the default case so that we know we didn't miss anything.

pkg/controllers/workgenerator/controller.go
```diff
@@ -779,20 +877,81 @@ func setBindingStatus(works map[string]*fleetv1beta1.Work, resourceBinding *flee
...
+			//
+			// In this case, set failed, diffed, and drifted placements.
 			failedManifests := extractFailedResourcePlacementsFromWork(w)
 			failedResourcePlacements = append(failedResourcePlacements, failedManifests...)
-		}
-		// Diffed placements can only appear when the Applied condition is set to False.
-		if appliedCond.Status == metav1.ConditionFalse {
+
 			diffedManifests := extractDiffedResourcePlacementsFromWork(w)
 			diffedResourcePlacements = append(diffedResourcePlacements, diffedManifests...)
```
ryanzhang-oss, line 918: why report diff when "appliedSummarizedStatus == workConditionSummarizedStatusFalse"?

ryanzhang-oss:
```
 Fleet will report such differences when:
	//
	// * The CRP uses the ReportDiff apply strategy, which instructs Fleet to compare the hub
	//   cluster manifests against the live resources without actually performing any apply op; or
	// * Fleet finds a pre-existing resource on the member cluster side that does not match its
	//   hub cluster counterpart, and the CRP has been configured to only take over a resource if
	//   no configuration differences are found.
```

ryanzhang-oss (approved):
Please address the comments on a follow up PR

pkg/controllers/workgenerator/controller.go
michaelawyu: Hi Ryan! This is not for the report diff mode; apply op failures can also have diffs (when the takeover setting is set to `TakeOverIfNoDiff`.

pkg/controllers/workgenerator/controller.go
michaelawyu: Got it; will send a subsequent PR.

PR #1019, opened 2025-01-17
fix: use index number in clusterStagedUpdateRun and add ApprovalAccepted status to ApprovalRequests

jwtty:

1. Show index in PolicySnapshotIndexObserved in updateRun status
2. Accept index in ResourceSnapshotIndex in updateRun spec
3. Add `ApprovalAccepted` status in approvalRequests
4. Fix the waitTime flaky UT (use no earlier than instead of after)
5. Fix observedClusterCount difference for pickAll policy in validation.

Fixes #

I have:


pkg/controllers/updaterun/controller.go
```diff
@@ -235,28 +235,39 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
 			// We only care about when an approval request is approved.
 			UpdateFunc: func(ctx context.Context, e event.UpdateEvent, q workqueue.RateLimitingInterface) {
 				klog.V(2).InfoS("Handling a clusterApprovalRequest update event", "clusterApprovalRequest", klog.KObj(e.ObjectNew))
-				handleClusterApprovalRequest(e.ObjectNew, q)
-			},
-			GenericFunc: func(ctx context.Context, e event.GenericEvent, q workqueue.RateLimitingInterface) {
-				klog.V(2).InfoS("Handling a clusterApprovalRequest generic event", "clusterApprovalRequest", klog.KObj(e.Object))
-				handleClusterApprovalRequest(e.Object, q)
+				handleClusterApprovalRequest(e.ObjectOld, e.ObjectNew, q)
```
ryanzhang-oss, line 238: just curious, I wonder why removing the genricFunc? I put it there just for safety as I am not 100% sure what is considered "generic".

pkg/controllers/updaterun/controller.go
```diff
@@ -235,28 +235,39 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
...
 			},
 		}).Complete(r)
 }
 
 // handleClusterApprovalRequest finds the ClusterStagedUpdateRun creating the ClusterApprovalRequest,
-// and enqueues it to the ClusterStagedUpdateRun controller queue.
-func handleClusterApprovalRequest(obj client.Object, q workqueue.RateLimitingInterface) {
-	approvalRequest, ok := obj.(*placementv1beta1.ClusterApprovalRequest)
+// and enqueues it to the ClusterStagedUpdateRun controller queue only when the approved condition gets changed.
```
ryanzhang-oss, line 244:
```suggestion
// and enqueues it to the ClusterStagedUpdateRun controller queue only when the approved condition is changed.
```

pkg/controllers/updaterun/validation.go
```diff
@@ -50,13 +50,14 @@ func (r *Reconciler) validate(
...
-		mismatchErr := fmt.Errorf("the policy snapshot index used in the clusterStagedUpdateRun is outdated, latest: %s, recorded: %s", latestPolicySnapshot.Name, updateRun.Status.PolicySnapshotIndexUsed)
+	if updateRun.Status.PolicySnapshotIndexUsed != updateRunCopy.Status.PolicySnapshotIndexUsed {
+		mismatchErr := fmt.Errorf("the policy snapshot index used in the clusterStagedUpdateRun is outdated, latest: %s, recorded: %s", updateRunCopy.Status.PolicySnapshotIndexUsed, updateRun.Status.PolicySnapshotIndexUsed)
 		klog.ErrorS(mismatchErr, "there's a new latest policy snapshot", "clusterResourcePlacement", placementName, "clusterStagedUpdateRun", updateRunRef)
 		return -1, nil, nil, fmt.Errorf("%w: %s", errStagedUpdatedAborted, mismatchErr.Error())
 	}
 	// Make sure the cluster count in the policy snapshot has not changed.
-	if updateRun.Status.PolicyObservedClusterCount != clusterCount {
+	// PickAll policy case will be verified in validateStagesStatus.
+	if clusterCount != -1 && updateRun.Status.PolicyObservedClusterCount != clusterCount {
```
ryanzhang-oss, line 60: would it still the same if both are -1 as we don't allow policy change either.

pkg/controllers/updaterun/controller.go
jwtty: genericFunc handles genericEvents, whose operation is unknown, e.g. triggered by a timer or events outside of the cluster: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/event#GenericEvent. Since I only want to trigger reconcile upon approval status change, I removed the genericFunc.

ryanzhang-oss: I guess my question is "is it guaranteed that the status (and in other cases, the spec) of an object is not changed when the operation is unknown"?  I haven't spent enough time in the code base to figure that out.

pkg/controllers/updaterun/validation.go
ryanzhang-oss: It's not ideal to assign values to PolicyObservedClusterCount in both determinePolicySnapshot and collectScheduledClusters functions.   It will be good to add some comments (and maybe a TODO to refactor)

ryanzhang-oss (approved):

PR #1029, opened 2025-01-31
feat: move the rollout controller to event driven model

ryanzhang-oss:

1. move the rollout controller to event driven model
2. minor fixes on comments and logs
3. fix the bug that rollout got stuck when the master snapshot is deleted but the binding is not synced yet. 
     

Fixes #

I have:


apis/placement/v1beta1/binding_types.go
```diff
@@ -20,8 +20,9 @@ const (
 // +kubebuilder:resource:scope=Cluster,categories={fleet,fleet-placement},shortName=rb
 // +kubebuilder:subresource:status
 // +kubebuilder:storageversion
-// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="WorkSynchronized")].status`,name="WorkCreated",type=string
+// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="WorkSynchronized")].status`,name="WorkSynchronized",type=string
 // +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Applied")].status`,name="ResourcesApplied",type=string
+// +kubebuilder:printcolumn:JSONPath=`.status.conditions[?(@.type=="Available")].status`,name="ResourceAvailable",priority=1,type=string
```
michaelawyu, line 25: Hi Ryan! This needs a re-gen of the CRD YAML files I think.

pkg/controllers/rollout/controller.go
```diff
@@ -563,6 +562,16 @@ func (r *Reconciler) calculateRealTarget(crp *fleetv1beta1.ClusterResourcePlacem
 // A binding with not trackable resources is considered ready if the binding's current spec has been available before
 // the ready cutoff time.
 func isBindingReady(binding *fleetv1beta1.ClusterResourceBinding, readyTimeCutOff time.Time) (time.Duration, bool) {
+	// the binding is ready if the diff report has been reported
```
michaelawyu, line 565: Hi Ryan! Aside from the `isBindingReady` function we also have a `HasBindingFailed` function. Before the update the two functions are mutually exclusive, i.e., if `isBindingReady` returns true, `HasBindingFailed` would always return false. With the new edit this no longer seems to be stand.

michaelawyu: Which brings up a situation that, for a bound binding with the ReportDiff apply strategy, it would be considered as a ready binding (because the `DiffReported` condition is true) and a failed binding at the same (because it has a false `Applied` conditon and no `Available` condition).

pkg/controllers/rollout/controller.go
```diff
@@ -563,6 +562,16 @@ func (r *Reconciler) calculateRealTarget(crp *fleetv1beta1.ClusterResourcePlacem
...
+	diffReportCondition := binding.GetCondition(string(fleetv1beta1.ResourceBindingDiffReported))
```
michaelawyu, line 566: Another thing that I might not have fully thought through is that if we treat `DiffReported` bindings to be not-failed, ready bindings  it would be subject to rolling update strategy as well, which means it could also get stuck (e.g., member clusters get disconnected). Would this be a very confusing situation for end users (i.e., all the diffs reported would be stale)?

pkg/controllers/rollout/controller.go
```diff
@@ -812,6 +821,38 @@ func handleResourceBinding(binding client.Object, q workqueue.RateLimitingInterf
...
+		klog.ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("update event is nil")), "Failed to process update event")
+		return
+	}
+	oldBinding, oldOk := objectOld.(*fleetv1beta1.ClusterResourceBinding)
+	newBinding, newOk := objectNew.(*fleetv1beta1.ClusterResourceBinding)
+	if !oldOk || !newOk {
+		klog.ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("failed to cast runtime objects in update event to cluster resource binding objects")), "Failed to process update event")
+	}
+	if oldBinding.GetResourceVersion() != newBinding.GetResourceVersion() {
+		klog.V(2).InfoS("The binding spec have changed, need to notify rollout controller", "binding", klog.KObj(newBinding))
```
michaelawyu, line 838: Hi Ryan! Just a nit: as resource version would change even upon status refreshes, the log part might need to be revised.

pkg/controllers/rollout/controller.go
```diff
@@ -812,6 +821,38 @@ func handleResourceBinding(binding client.Object, q workqueue.RateLimitingInterf
...
+		enqueueResourceBinding(newBinding, q)
+		return
+	}
+	// these are the conditions we care about
+	conditionsToMonitor := []string{string(fleetv1beta1.ResourceBindingDiffReported), string(fleetv1beta1.ResourceBindingAvailable)}
```
michaelawyu, line 843: Hi Ryan! All the logic down below would not run if I understand it correctly; if the conditions change the resource versions must have already mismatched; on the contrary, if the resource versions stay the same there would be no status change at all I believe

michaelawyu (commented):
Hi Ryan! I've added some comments, PTAL 🙏

ryanzhang-oss:
The bug was the workGenerator controller will mark the binding as overridden failed but rollout controller doesn't consider that binding has "FAILED". This leads to the roll out stuck because the binding can be ready thus it prevent the maxNumber to go over.

There needs to be a follow up on the workGenerator/applier to differentiate (best effort) user error/not retriable error from retriable.

michaelawyu (dismissed):
LGTM ;) Has some minor comments; will add them later -> merging the PR now upon test completion to unblock progress

michaelawyu (approved):
LGTM.

PR #1037, opened 2025-02-10
docs: update ro docs

zhiying-lin:

- update the ro how-to docs to include new features
- reorg the structure of cro & ro docs

image

Fixes #

I have:

### DIFF
--- a/docs/howtos/cluster-resource-override.md
+++ b/docs/howtos/cluster-resource-override.md
@@ -122,7 +122,7 @@ The jsonPatchOverrides field supports the following fields:
 
 ##### Example: Override Labels
     
-For example, to add a label to the `ClusterRole` named `secret-reader` on clusters with the label `env: prod`,
+To overwrite the existing labels on the `ClusterRole` named `secret-reader` on clusters with the label `env: prod`,
 you can use the following configuration:
 ```yaml
 apiVersion: placement.kubernetes-fleet.io/v1alpha1
@@ -150,6 +150,14 @@ spec:
             value:
               {"cluster-name":"${MEMBER-CLUSTER-NAME}"}
 ```
+
+> Note: To add a new label to the existing labels, please use the below configuration:
+> ```yaml
+>  - op: add
+>    path: /metadata/labels/new-label
+>    value: "new-value"
+> ```
+
 The `ClusterResourceOverride` object above will add a label `cluster-name` with the value of the `memberCluster` name to the `ClusterRole` named `secret-reader` on clusters with the label `env: prod`.
 
 ##### Example: Remove Verbs
--- a/docs/howtos/cluster-resource-override.md
+++ b/docs/howtos/cluster-resource-override.md
@@ -49,7 +49,7 @@ spec:
       name: secret-reader
 ```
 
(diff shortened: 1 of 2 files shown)


docs/howtos/resource-override.md
```diff
@@ -186,6 +231,39 @@ The `jsonPatchOverrides` field supports the following fields:
...
+       version: v1
+       name: my-deployment
+  policy:
+    overrideRules:
+      - clusterSelector:
+          clusterSelectorTerms:
+            - labelSelector:
+                matchLabels:
+                  env: prod
+        jsonPatchOverrides:
```
jwtty, line 260: Can we also have an example for delete? thanks!

zhiying-lin: it's in line 165

docs/howtos/resource-override.md
```diff
@@ -149,12 +162,44 @@ with the image `nginx:1.20.0` on all clusters with the label `env: prod`.
...
+  policy:
+    overrideRules:
+      - clusterSelector:
+          clusterSelectorTerms:
+            - labelSelector:
+                matchLabels:
+                  env: test
+        overrideType: Delete
+```
+
```
ryanzhang-oss, line 190: why not move this example into the Override Type section?

ryanzhang-oss: or move the OverrideType section earlier or remove it?   Ideally, the OverrideType section should be the parent of JSON Patch Override and have a delete type section. Currently, it's at the same level as JSON Patch and no corresponding Delete section.

zhiying-lin: i think the current logic is to give concrete examples at the very beginning of the policy section, and then go deep into each fields. it should be ok?

docs/howtos/resource-override.md
```diff
@@ -10,12 +10,17 @@ These changes can include updates to container images, environment variables, re
...
+- **Placement**: This specifies which placement the override is applied to.
 - **Resource Selectors**: These specify the set of resources selected for overriding.
 - **Policy**: This specifies the policy to be applied to the selected resources.
 
 
 The following sections discuss these components in depth.
 
+## Placement
+
+To configure which placement the override is applied to, you can use the name of `ResourcePlacement`.
```
jwtty, line 22: ClusterResourcePlacement or do we support namespaced ResourcePlacement?

docs/howtos/cluster-resource-override.md
```diff
@@ -179,6 +120,8 @@ The jsonPatchOverrides field supports the following fields:
     * There is a list of reserved variables that will be replaced by the actual values:
       * `${MEMBER-CLUSTER-NAME}`:  this will be replaced by the name of the `memberCluster` that represents this cluster.
 
+##### Example: Override Labels
+    
 For example, to add a label to the `ClusterRole` named `secret-reader` on clusters with the label `env: prod`,
```
jwtty, line 125: I wonder if it's better to say "to set the labels of the `ClusterRole` named `secret-reader`..." because below example resets labels field. Or use `/metadata/labels/cluster-name` instead with a caveat about empty labels.

jwtty (approved):

PR #1038, opened 2025-02-11
feat: handle the reportDiff Condition in the rollout controller 

ryanzhang-oss:
1. Special handle the reportDiff Condition in the rollout controller
2. Special handle the  reportDiff Condition in the eviction controller.

Fixes #

I have:


pkg/controllers/rollout/controller.go
```diff
@@ -358,7 +358,7 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
 		case fleetv1beta1.BindingStateUnscheduled:
 			if bindingutils.HasBindingFailed(binding) {
 				klog.V(3).InfoS("Found a failed to be ready unscheduled binding", "clusterResourcePlacement", crpKObj, "binding", bindingKObj)
-			} else {
+			} else if !bindingutils.IsBindingReportDiff(binding) {
```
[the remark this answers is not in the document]
michaelawyu: Since `canBeReadyBinding` is used only when calculating the `maxNumberToAdd`, it would mean that for N target clusters, there will be at max. 2N (+maxSurge) clusters at any time being selected, as long as the `DiffReported` condition is populated for all N existing bindings.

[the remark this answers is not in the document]
michaelawyu: This has no side effect per se, as `ReportDiff` mode does not make any change in member clusters; it is just that the behavior is a bit inconsistent with what would happen when CSA/SSA apply strategy is used.

pkg/controllers/rollout/controller.go
```diff
@@ -407,7 +407,7 @@ func (r *Reconciler) pickBindingsToRoll(ctx context.Context, allBindings []*flee
 			if bindingutils.HasBindingFailed(binding) {
 				klog.V(3).InfoS("Found a failed to be ready bound binding", "clusterResourcePlacement", crpKObj, "binding", bindingKObj)
 				bindingFailed = true
-			} else {
+			} else if !bindingutils.IsBindingReportDiff(binding) {
```
michaelawyu, line 410: A similar pattern happens here as well:

Since `HasBindingFailed` will only return `True` when either `Applied` or `Available` condition is set to False and fresh, any binding in `ReportDiff` mode will never be considered to be failing.

michaelawyu: So `ReportDiff` bindings are always updated per the `maxUnavailable` setting, even if the `DiffReported` condition has turned False (an error has occurred).

michaelawyu: Not to say that it is a wrong behavior though. Again, it is just an inconsistency.

michaelawyu (commented):
Added a few comments, PTAL 🙏

michaelawyu:
Hi Ryan! I've gotten another minor concern for the new logic: at this moment we are only comparing the conditions -> for regular cases this works greatly, but in situations where apply strategy types are switched (ReportDiff <-> CSA/SSA), there are many intermediate states that are going to get difficult to reason about.

pkg/controllers/rollout/controller.go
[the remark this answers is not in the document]
ryanzhang-oss: yes, you are right. The reason that I added this is to not let ReportDiff to block the rollout based on max surge as we are not actually applying the resources on the clusters. AFAIK, the main reason for max surge is to prevent too many copies of a resource to exist.

ryanzhang-oss:
> 
> Since the CRP is always the source of truth, would it be better (possible) for us to always reference that?

Agree, the intermediate state is confusing.  I wonder how can we reference CRP as the truth?

michaelawyu (commented):
LGTM ;) 

Have some minor comments, will add them later; approve now to unblock the progress.

pkg/controllers/rollout/controller_test.go
```diff
@@ -732,6 +732,94 @@ func TestIsBindingReady(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			readyTimeCutOff: now,
+			wantReady:       false,
+			wantWaitTime:    -1,
+		},
+
+		"binding diff report false on different generation with successful current apply should be ready": {
```
michaelawyu, line 796: Hi Ryan! A nit: this should be available?

pkg/controllers/rollout/controller_test.go
```diff
@@ -732,6 +732,94 @@ func TestIsBindingReady(t *testing.T) {
...
+							LastTransitionTime: metav1.NewTime(now.Add(-5 * time.Minute)),
+							ObservedGeneration: 9, //previous generation
+						},
+					},
+				},
+			},
+			readyTimeCutOff: now,
+			wantReady:       true,
+			wantWaitTime:    0,
+		},
```
michaelawyu, line 822: We probably also need a `diffReported is current but available is stale` case.

pkg/utils/binding/binding_test.go
```diff
@@ -315,3 +315,97 @@ func TestHasBindingFailed(t *testing.T) {
 		})
 	}
 }
+
```
michaelawyu, line 318: A nit (when I was resolving conflicts): the case on line 248 is essentially the same as the case on line 278.

michaelawyu (approved):

Arvindthiru (approved):

PR #1039, opened 2025-02-11
feat: tweak the condition behaviors in the new work applier

michaelawyu:

This PR tweaks how the new work applier sets conditions for the Work objects and their manifests.

I have:

- [x] Unit tests
- [x] Integration tests

### DIFF
--- a/pkg/controllers/workapplier/controller.go
+++ b/pkg/controllers/workapplier/controller.go
@@ -111,22 +111,24 @@ func NewReconciler(
 	}
 }
 
-const (
-	allManifestsAppliedMessage          = "All the specified manifests have been applied"
-	allAppliedObjectAvailableMessage    = "All of the applied manifests are available"
-	someAppliedObjectUntrackableMessage = "Some of the applied manifests cannot be tracked for availability"
-
-	notAllManifestsAppliedReason         = "FailedToApplyAllManifests"
-	notAllManifestsAppliedMessage        = "Failed to apply all the specified manifests (%d of %d manifests are applied)"
-	notAllAppliedObjectsAvailableReason  = "NotAllAppliedObjectAreAvailable"
-	notAllAppliedObjectsAvailableMessage = "Not all of the applied manifests are available (%d of %d manifests are available)"
-
-	someObjectsHaveDiffs = "Found configurations diffs on some objects (%d of %d objects have diffs)"
-)
-
 var (
-	// Some exported reasons. Currently only the untrackable reason is being actively used.
-	WorkNotTrackableReason = string(ManifestProcessingAvailabilityResultTypeNotTrackable)
+	// Some exported reasons for Work object conditions. Currently only the untrackable reason is being actively used.
+	WorkNotAllManfestsTrackableReason     = "SomeManifestsAreNotAvailabilityTrackable"
+	WorkAllManifestsAppliedReason         = "AllManifestsApplied"
+	WorkAllManifestsAvailableReason       = "AllManifestsAvailable"
+	WorkAllManifestsDiffReportedReason    = "AllManifestsDiffReported"
+	WorkNotAllManifestsAppliedReason      = "SomeManifestsAreNotApplied"
+	WorkNotAllManifestsAvailableReason    = "SomeManifestsAreNotAvailable"
+	WorkNotAllManifestsDiffReportedReason = "SomeManifestsHaveNotReportedDiff"
+
+	// Some condition messages for Work object conditions.
+	allManifestsAppliedMessage           = "All the specified manifests have been applied"
+	allManifestsHaveReportedDiffMessage  = "All the specified manifests have reported diff"
+	allAppliedObjectAvailableMessage     = "All of the applied manifests are available"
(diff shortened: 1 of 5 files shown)


pkg/controllers/workapplier/status.go
```diff
@@ -151,24 +148,27 @@ func (r *Reconciler) refreshWorkStatus(
 		if bundle.availabilityResTyp == ManifestProcessingAvailabilityResultTypeNotTrackable {
 			untrackableAppliedObjectsCount++
 		}
+		if isManifestObjectDiffReported(bundle.reportDiffResTyp) {
+			diffReportedObjectsCount++
+		}
```
ryanzhang-oss, line 153: Why Is there no UT test change related to this change that count "both diff and noDiff as reported"

pkg/controllers/workapplier/status.go
```diff
@@ -151,24 +148,27 @@ func (r *Reconciler) refreshWorkStatus(
...
+				appliedManifestsCount, availableAppliedObjectsCount, untrackableAppliedObjectsCount, diffReportedObjectsCount, manifestCount))
 	}
 
 	if work.Status.Conditions == nil {
 		work.Status.Conditions = []metav1.Condition{}
 	}
 	setWorkAppliedCondition(work, manifestCount, appliedManifestsCount)
-	setWorkAvailableCondition(work, manifestCount, availableAppliedObjectsCount, untrackableAppliedObjectsCount)
-	setWorkDiffReportedCondition(work, manifestCount, diffedObjectsCount)
+	setWorkAvailableCondition(work, manifestCount, appliedManifestsCount, availableAppliedObjectsCount, untrackableAppliedObjectsCount)
```
ryanzhang-oss, line 170: Nit: what if the appliedCondition (say false) does not agree with appliedManifestsCount (say == manifestCount)? I think it's slightly better to use the output of the setWorkAppliedCondition to determine the appliedCondition than using "appliedManifestsCount" as they are two source of truth

pkg/controllers/workapplier/status.go
```diff
@@ -268,7 +272,13 @@ func setManifestAppliedCondition(
...
 	}
 
-	meta.SetStatusCondition(&manifestCond.Conditions, *appliedCond)
+	if appliedCond != nil {
+		meta.SetStatusCondition(&manifestCond.Conditions, *appliedCond)
+	} else {
+		// As the conditions are ported back; removal must be performed if the DiffReported
+		// condition is not set.
+		meta.RemoveStatusCondition(&manifestCond.Conditions, fleetv1beta1.WorkConditionTypeApplied)
+	}
```
ryanzhang-oss, line 281: I don't see any test changes that is related to most of the changes in this file

ryanzhang-oss (approved):

pkg/controllers/workapplier/status.go
```diff
@@ -412,33 +421,41 @@ func setWorkAppliedCondition(
...
+	manifestCount, appliedManifestCount, availableManifestCount, untrackableAppliedObjectsCount int,
 ) {
 	var availableCond *metav1.Condition
 	switch {
 	case work.Spec.ApplyStrategy != nil && work.Spec.ApplyStrategy.Type == fleetv1beta1.ApplyStrategyTypeReportDiff:
 		// ReportDiff mode is on; no apply op has been performed, and consequently
 		// Fleet will not update the Available condition.
-		return
+	case appliedManifestCount != manifestCount:
+		// Not all manifests have been applied; skip updating the Available condition.
```
ryanzhang-oss, line 452: same here, no test change needed?

pkg/controllers/workapplier/controller.go
```diff
@@ -111,22 +111,24 @@ func NewReconciler(
...
-	notAllAppliedObjectsAvailableMessage = "Not all of the applied manifests are available (%d of %d manifests are available)"
-
-	someObjectsHaveDiffs = "Found configurations diffs on some objects (%d of %d objects have diffs)"
-)
-
 var (
-	// Some exported reasons. Currently only the untrackable reason is being actively used.
-	WorkNotTrackableReason = string(ManifestProcessingAvailabilityResultTypeNotTrackable)
+	// Some exported reasons for Work object conditions. Currently only the untrackable reason is being actively used.
+	WorkNotAllManfestsTrackableReason     = "SomeManifestsAreNotAvailabilityTrackable"
```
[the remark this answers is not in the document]
michaelawyu: Fixed 👌

pkg/controllers/workapplier/status.go
michaelawyu: Additional UTs added.

pkg/controllers/workapplier/status.go
michaelawyu: In the prev. commit there are a number of integration tests that come with the always-set `Available` condition, and it's removed in this PR -> but it's not explicit/covered enough, will add more test cases.

pkg/controllers/workapplier/status.go
michaelawyu: Will add more tests. Before there was the ReportDiff mode UT that covers this case (`Applied` -> `DiffReported`, no left over conditions); can always use more of course.

PR #1044, opened 2025-02-14
fix: remove uncached reader in eviction controller

Arvindthiru:

Fixes #

I have:

Removed the uncached reader from eviction controller, was originally added to account for cache miss when attempting to get PDB in the controller, E2Es were failing

Steps taken to remedy issue,

- Moved the creation of PDB to BeforeAll block to give more time for cache to be populated
- Added FlakeAttempts decorator to tests which run in parallel and requires the creation of PDB to account for cache miss in the controller

These steps aim to optimistically ensure that the E2Es are retried even on a cache and hopefully the cache is populated on the retry to successfully get PDB and block eviction

### DIFF
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -229,8 +229,7 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 			}
 			klog.Info("Setting up cluster resource placement eviction controller")
 			if err := (&clusterresourceplacementeviction.Reconciler{
-				Client:         mgr.GetClient(),
-				UncachedReader: mgr.GetAPIReader(),
+				Client: mgr.GetClient(),
 			}).SetupWithManager(mgr); err != nil {
 				klog.ErrorS(err, "Unable to set up cluster resource placement eviction controller")
 				return err
--- a/pkg/controllers/clusterresourceplacementeviction/controller.go
+++ b/pkg/controllers/clusterresourceplacementeviction/controller.go
@@ -30,8 +30,6 @@ import (
 // Reconciler reconciles a ClusterResourcePlacementEviction object.
 type Reconciler struct {
 	client.Client
-	// UncachedReader is only used to read disruption budget objects directly from the API server to ensure we can enforce the disruption budget for eviction.
-	UncachedReader client.Reader
 }
 
 // Reconcile triggers a single eviction reconcile round.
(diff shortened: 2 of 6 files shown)


michaelawyu:
Hi Arvind! To summarize some of the approaches we have gone through (specific to the test part):

a) use the `FlakeAttempts=N` decorator on the container node; optionally we could use a random suffix on the resources we create to simplify the whole flow

    Please be aware of how Ginkgo handles the `FlakeAttempts` decorator on ordered nodes:

    ```
    In brief, Ginkgo generally guarantees that BeforeAll and AfterAll node closures only run once - but FlakeAttempts can modify this behavior. If a failure occurs within a subject node in an Ordered container (i.e. in an It) then Ginkgo will rerun that It but not the BeforeAll or AfterAll. However, if a failure occurs in a BeforeAll, Ginkgo will immediately run the AfterAll (to clean up) and then rerun the BeforeAll.
    ```
b) use the `wait (k8s.io/apimachinery/pkg/util/wait)` package (there are similar alternatives) to achieve the nested `Eventually` block

    As mentioned in the doc, Ginkgo will handle the Eventually func calls correctly so with right timeout/interval settings this option should work just fine. It's also possible to control the timeout dynamically if we prefer not setting the timeout/interval explicitly (or let the inner block decide when to give up -> this would be functionally very similar to the `FlakeAttempts` decorator).

michaelawyu:
Hi Arvind! To summarize some of the approaches we have gone through (specific to the test part):

a) use the `FlakeAttempts=N` decorator on the container node; optionally we could use a random suffix on the resources we create to simplify the whole flow

    Please be aware of how Ginkgo handles the `FlakeAttempts` decorator on ordered nodes:

    ```
    In brief, Ginkgo generally guarantees that BeforeAll and AfterAll node closures only run once - but FlakeAttempts can modify this behavior. If a failure occurs within a subject node in an Ordered container (i.e. in an It) then Ginkgo will rerun that It but not the BeforeAll or AfterAll. However, if a failure occurs in a BeforeAll, Ginkgo will immediately run the AfterAll (to clean up) and then rerun the BeforeAll.
    ```
b) use the `wait (k8s.io/apimachinery/pkg/util/wait)` package (there are similar alternatives) to achieve the nested `Eventually` block

    As mentioned in the doc, Ginkgo will handle the Eventually func calls correctly so with right timeout/interval settings this option should work just fine. It's also possible to control the timeout dynamically if we prefer not setting the timeout/interval explicitly (or let the inner block decide when to give up -> this would be functionally very similar to the `FlakeAttempts` decorator).

PR #1056, opened 2025-02-25
Fix: fix the cro example

ryanzhang-oss:

A new map node in the Json is different from adding a kv pair inside a map.

Fixes #

I have:

### DIFF
--- a/examples/test-cro1.yaml
+++ b/examples/test-cro1.yaml
@@ -19,5 +19,5 @@ spec:
                   env: canary
         jsonPatchOverrides:
           - op: add
-            path: /metadata/labels/new-label
-            value: "new-value"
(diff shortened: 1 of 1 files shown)


examples/test-cro1.yaml
```diff
@@ -19,5 +19,5 @@ spec:
                   env: canary
         jsonPatchOverrides:
           - op: add
-            path: /metadata/labels/new-label
-            value: "new-value"
\ No newline at end of file
+            path: /metadata/labels
+            value: {"new-label": "new-value"}
```
zhiying-lin, line 23: hi ryan, this example is to add labels when the list is empty.

We need an example: add a new label when the list is not empty.
The original example is supposed to work :(

PR #1058, opened 2025-02-26
docs: add comment to the override examples

zhiying-lin:
- add documentation for the override examples
- add a UT

Fixes #

I have:

### DIFF
--- a/pkg/controllers/workgenerator/override_test.go
+++ b/pkg/controllers/workgenerator/override_test.go
@@ -2115,6 +2115,37 @@ func TestApplyJSONPatchOverride(t *testing.T) {
 			},
 		},
 		{
+			name: "reset the labels",
+			deployment: appsv1.Deployment{
+				TypeMeta: deploymentType,
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "deployment-name",
+					Namespace: "deployment-namespace",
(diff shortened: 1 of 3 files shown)


pkg/controllers/workgenerator/override_test.go
```diff
@@ -2114,6 +2114,37 @@ func TestApplyJSONPatchOverride(t *testing.T) {
...
+					Namespace: "deployment-namespace",
+					Labels: map[string]string{
+						"app": "nginx-1",
+						"key": "value",
+					},
+				},
+			},
+			overrides: []placementv1alpha1.JSONPatchOverride{
+				{
+					Operator: placementv1alpha1.JSONPatchOverrideOpAdd,
```
jwtty, line 2132: Just curious, will it have similar effect to use `JSONPatchOverrideOpReplace`?

ryanzhang-oss: it seems this is more of a "replace" semantic, the add op is kinda weird

zhiying-lin: added another test and comment.

The main diff between "Replace" and "Add" is that add works even the path is missing.

jwtty (approved):

PR #1059, opened 2025-02-26
feat: add metric for crp completion

britaniar:
I have: add prometheus metric to show CRP completion which will be used for alerting and monitoring.

image
Metric includes label for CRP name. If value is "1", CRP rollout is completed. Otherwise, if value is "0", the CRP rollout is incomplete.

Once CRP is deleted, the metric will also disappear. Below I deleted the CRP "test-crp", which is no longer on the list below.
image

image

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -101,7 +101,7 @@ func (r *Reconciler) handleDelete(ctx context.Context, crp *fleetv1beta1.Cluster
 	}
 	klog.V(2).InfoS("Removed crp-cleanup finalizer", "clusterResourcePlacement", crpKObj)
 	r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementCleanupFinalizerRemoved", "Deleted the snapshots and removed the placement cleanup finalizer")
-	metrics.FleetPlacementCount.Delete(prometheus.Labels{"name": crp.Name, "status": controller.LabelComplete})
+	metrics.FleetPlacementStatus.Delete(prometheus.Labels{"name": crp.Name})
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -99,6 +101,7 @@ func (r *Reconciler) handleDelete(ctx context.Context, crp *fleetv1beta1.Cluster
 	}
 	klog.V(2).InfoS("Removed crp-cleanup finalizer", "clusterResourcePlacement", crpKObj)
 	r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementCleanupFinalizerRemoved", "Deleted the snapshots and removed the placement cleanup finalizer")
+	metrics.FleetPlacementStatus.Delete(prometheus.Labels{"name": crp.Name})
```
ryanzhang-oss, line 104: I think that there is a way to test metrics output

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -219,10 +222,12 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			klog.V(2).InfoS("Placement has finished the rollout process and reached the desired status", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 			r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Placement has finished the rollout process and reached the desired status")
 		}
+		metrics.FleetPlacementStatus.WithLabelValues(crp.Name).Set(1)
```
ryanzhang-oss, line 225: Do we want to repeatedly record this or just like the event, only record it once per completion?

pkg/controllers/clusterresourceplacement/controller.go
britaniar: The output is up on top.

britaniar: It would show up in Kusto as: 
{
	"cloud": "<cloud>",
	"cluster_id": "<ccp-namespace>",
	"name": "<crp-name>",
	"environment": "<environment",
	"job": "fleet-hub-agent",
	"pod": "<fleet-hub-agent-pod>",
	"scrape_location": "<location>"
}

pkg/controllers/clusterresourceplacement/controller.go
britaniar: In this case it would be used like an event (in which we would look for when it changed from 0 -> 1, or 1 -> 0). If we wanted more information like how many times it succeeded we could increment the metric only when it completes and do nothing when it doesn't (in this case we would probably want to use a counter instead).

britaniar: I was thinking the former as we want to see the number of completed CRPs and its completed percentage amongst all CRPs. This would also give us a view of which CRPs are not completed.

zhiying-lin: we probably need to add more values to distinguish the incomplete types, for example, not scheduled, or failed to applied, or unavailable, or in the processing, etc.

In addition, we need to set the incomplete state when the CRP hits some internal error (ex, fail to update status), since we're using gauge type. if we don't reset, the CRP will be always treated as completed.

ryanzhang-oss (approved):
please add a follow up to add "edge" metrics

PR #1060, opened 2025-02-28
util: tools to drain and uncordon cluster

Arvindthiru:

Adding new tools to drain and uncordon cluster

Fixes #

I have:

If drain fails we automatically uncordon cluster


tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	if err := clusterv1beta1.AddToScheme(scheme); err != nil {
+		log.Fatalf("failed to add custom APIs (cluster) to the runtime scheme: %v", err)
+	}
+	if err := placementv1beta1.AddToScheme(scheme); err != nil {
+		log.Fatalf("failed to add custom APIs (placement) to the runtime scheme: %v", err)
+	}
+
+	clusterConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
```
ryanzhang-oss, line 45: what is this for?

tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,145 @@
...
+			memberCluster = &memberClusterList.Items[i]
+			break
+		}
+	}
+
+	if memberCluster == nil {
+		log.Fatalf("failed to find member cluster %s to cordon", *clusterName)
+	}
+
+	var crbList placementv1beta1.ClusterResourceBindingList
```
ryanzhang-oss, line 78: why do we need to list crb instead of CRP directly?

tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+
+	if len(crpNameMap) == 0 {
+		log.Fatalf("failed to find any CRP which has propagated resource to cluster %s", *clusterName)
+	}
+
+	// retry until update succeeds or non conflict error occurs.
+	for {
+		err := updateMemberClusterWithTaint(ctx, hubClient, *clusterName)
```
ryanzhang-oss, line 99: retry with backoff?

tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,145 @@
...
+			Spec: placementv1beta1.PlacementEvictionSpec{
+				PlacementName: crpName,
+				ClusterName:   *clusterName,
+			},
+		}
+		if err := hubClient.Create(ctx, &eviction); err != nil {
+			log.Fatalf("failed to create CRP eviction for CRP %s", crpName)
+		}
+	}
+}
```
ryanzhang-oss, line 121: wait and get the result back

tools/cordon-cluster/main/main.go
Arvindthiru: clusterConfig -> restConfig -> hubClient, similar setup to our E2E

tools/README.md
ryanzhang-oss, line 1: This should be moved inside cordon dir so we can have other tools in the futrue

tools/cordon-cluster/main/main.go
ryanzhang-oss, line 1: why having a dir named "main" with only a file named "main" inside?

tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+		}
+	}
+
+	if len(crpNameMap) == 0 {
+		log.Fatalf("failed to find any CRP which has propagated resource to cluster %s", *clusterName)
+	}
+
+	// add taint to member cluster to ensure resources aren't scheduled on it.
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
ryanzhang-oss, line 82: This needs to be in the beginning to prevent any future placement.

tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,176 @@
...
+	for i := range crpList.Items {
+		for j := range crpList.Items[i].Status.PlacementStatuses {
+			if crpList.Items[i].Status.PlacementStatuses[j].ClusterName == *clusterName {
+				crpNameMap[crpList.Items[i].Name] = true
+				break
+			}
+		}
+	}
+
+	if len(crpNameMap) == 0 {
```
ryanzhang-oss, line 77: this is not a fatal error, user have no idea if there is placement or not.

tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,176 @@
...
+	var crpList placementv1beta1.ClusterResourcePlacementList
+	if err = hubClient.List(ctx, &crpList); err != nil {
+		log.Fatalf("failed to list cluster resource placements: %v", err)
+	}
+
+	// find all unique CRP names for which eviction needs to occur.
+	crpNameMap := make(map[string]bool)
+	for i := range crpList.Items {
+		for j := range crpList.Items[i].Status.PlacementStatuses {
+			if crpList.Items[i].Status.PlacementStatuses[j].ClusterName == *clusterName {
```
ryanzhang-oss, line 70: This doesn't mean the binding exists, what happens if it doesn't exist. What about binding is not bound?

tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,176 @@
...
+			if err := hubClient.Get(ctx, types.NamespacedName{Name: evictionName}, &eviction); err != nil {
+				return false, err
+			}
+			validCondition := eviction.GetCondition(string(placementv1beta1.PlacementEvictionConditionTypeValid))
+			if condition.IsConditionStatusFalse(validCondition, eviction.GetGeneration()) {
+				return true, nil
+			}
+			executedCondition := eviction.GetCondition(string(placementv1beta1.PlacementEvictionConditionTypeExecuted))
+			if executedCondition != nil {
+				return true, nil
```
ryanzhang-oss, line 153:
```suggestion
if eviction.GetCondition(string(placementv1beta1.PlacementEvictionConditionTypeExecuted)) != nil {
				return true, nil
```

tools/cordon-cluster/main/main.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+			return false, nil
+		})
+
+		if err != nil {
+			log.Fatalf("failed to wait for evictions to reach terminal state: %v", err)
+		}
+	}
+
+	log.Printf("Issued evictions to cordon member cluster %s, please verify to ensure evictions were successfully executed", *clusterName)
```
ryanzhang-oss, line 163: that's it?

tools/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+CURRENT   NAME               CLUSTER            AUTHINFO                                            NAMESPACE         
+*         hub                hub                clusterUser_clusterResourceGroup_hub   
+```
+
+The command internally creates `ClusterResourcePlacementEviction` objects for all the `ClusterResourcePlacement` 
+objects that have propagated resources to the member cluster that needs to be cordoned. And it also adds a `Taint` the 
+`MemberCluster` resource of the member cluster to prevent any new resources from being propagated to the member cluster.
+
+## Uncordon Member Cluster connected to a fleet
```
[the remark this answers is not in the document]
ryanzhang-oss: my bad, kubectl uncordon is just removing taint

tools/draincluster/main.go
```diff
@@ -0,0 +1,281 @@
...
+		for i := range d.ClusterResourcePlacementResourcesMap[crpName] {
+			resourceIdentifier := d.ClusterResourcePlacementResourcesMap[crpName][i]
+			log.Printf("evicted resource %s propagated by CRP %s", fmt.Sprintf(resourceIdentifierKeyFormat, resourceIdentifier.Group, resourceIdentifier.Version, resourceIdentifier.Kind, resourceIdentifier.Name, resourceIdentifier.Namespace), crpName)
+		}
+	}
+
+	return isDrainSuccessful, nil
+}
+
+func resourcePropagatedByCRP(selectedResources []placementv1beta1.ResourceIdentifier, failedPlacements []placementv1beta1.FailedResourcePlacement) []placementv1beta1.ResourceIdentifier {
```
Arvindthiru, line 248: Just realized, we can get propagated resources from the work object instead

tools/draincluster/main.go
```diff
@@ -0,0 +1,282 @@
...
+			if !evictionutils.IsPlacementPresent(targetBinding) {
+				return false, fmt.Errorf("placement is not present in cluster %s for CRP %s, "+
+					"please retry drain once resources are propagated to the cluster", d.ClusterName, crp.Name)
+			}
+			// get all successfully applied resources for the CRP.
+			resourcesPropagated, err := resourcesPropagatedByCRP(ctx, d.hubClient, crp.Name)
+			if err != nil {
+				return false, fmt.Errorf("failed to get resources propagated by CRP %s: %w", crp.Name, err)
+			}
+			d.ClusterResourcePlacementResourcesMap[crp.Name] = resourcesPropagated
```
ryanzhang-oss, line 164: can we just record the cluster scoped resources?

michaelawyu:
Hi Arvind! Below are some of the points I've been thinking about since our earlier discussion; hopefully they can be of use/relevance to you:

michaelawyu:
> Hi Arvind! Below are some of the points I've been thinking about since our earlier discussion; hopefully they can be of use/relevance to you:

a) The `kubectl drain` command functions similar to what we have in this PR right now, which first marks a node as unschedulable (cordoned) then attempts to evict all pods (if applicable -> certain types of pods will be exempted). The tool is not fully consistent; there still leaves a (albeit quite short) time window when a pod can be assigned to a node during the draining process -> this can happen if the draining runs when the scheduler is about to finish scheduling a pod. In this regard, however, K8s has an advantage over us: in K8s we have first the pods, then the scheduling decisions (whereas in Fleet, bindings are provisioned after the scheduling is completed); so if strong consistency is desired (i.e., must drain everything applicable), `kubectl` can just watch for unassigned pods in addition to assigned pods, and wait until scheduling is complete on these unassigned pods (successful or not, either way's OK).

michaelawyu:
b) as discussed earlier in our sync, if we would like to implement stronger consistency in the draining tool, options would include

1) use the `NoExecute` semantics -> this is not yet implemented
2) set up the rollout controller to reject bindings associated with a unschedulable cluster -> the concerns are mostly that *) this incurs coupling/logic leaking in our code + *) since we have multiple rollout controller options now it would not feel right to make this case a part of our rollout contract 
3) set up the scheduler to be able to ACK cordoning/draining attempts, so that we know for sure no bindings would be created any more -> this option also incurs concerns regarding coupling/logic leaking

There might not be a very easily approachable solution here.

michaelawyu:
c) With that being said, it might make more sense for us to evaluate how important the stronger consistency is to our users, esp. considering that `kubectl` does not give the same guarantee either (though in theory it can implement this much more easily).

michaelawyu:
d) A side note is that, `kubectl drain/cordon` marks a cluster as unschedulable by setting its node spec (there is a `schedulable` boolean value) -> it does not use taints (consequently no toleration can be set up). Of course this might not necessarily apply to our case

tools/utils/common.go
```diff
@@ -0,0 +1,74 @@
...
+	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	kubeConfigPath = os.Getenv("KUBECONFIG")
+	CordonTaint    = clusterv1beta1.Taint{
```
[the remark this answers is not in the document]
Arvindthiru: https://github.com/Azure/fleet/pull/1060#issuecomment-2734454161

tools/utils/common.go
```diff
@@ -0,0 +1,74 @@
...
+		&clientcmd.ConfigOverrides{
+			CurrentContext: clusterContext,
+		})
+
+	restConfig, err := clusterConfig.ClientConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	hubClient, err := client.New(restConfig, client.Options{Scheme: scheme})
```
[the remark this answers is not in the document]
Arvindthiru: Currently it doesn't use cache https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/client.go#L82-L89

tools/draincluster/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+The command adds a `Taint` the `MemberCluster` resource of the member cluster to prevent any new resources from being 
+propagated to the member cluster. Then it creates `ClusterResourcePlacementEviction` objects for all the 
+`ClusterResourcePlacement` objects that have propagated resources to the member cluster.
+
+>> **Note**: The `draincluster` tool is a best-effort mechanism at the moment, and it may not remove all the resources that 
+> have been propagated to the member cluster. To ensure that drain is successful the users can ensure that no new 
+> changes are being propagated to the member cluster when running the tool by any `ClusterResourcePlacement` resource 
+> because at the time of draining the member cluster, a `ClusterResourcePlacement` could be propagating resources to the 
+> member cluster these resources may not be removed by the tool.
```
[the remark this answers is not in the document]
Arvindthiru: https://kubernetes.io/images/docs/kubectl_drain.svg, updated doc as discussed offline

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,244 @@
...
+		if executedCondition == nil || executedCondition.Status == metav1.ConditionFalse {
+			isDrainSuccessful = false
+			log.Printf("eviction %s was not executed successfully for CRP %s", evictionName, crpName)
+			continue
+		}
+		log.Printf("eviction %s was executed successfully for CRP %s", evictionName, crpName)
+		// log each resource evicted by CRP.
+		for i := range h.ClusterResourcePlacementResourcesMap[crpName] {
+			resourceIdentifier := h.ClusterResourcePlacementResourcesMap[crpName][i]
+			log.Printf("evicted resource %s propagated by CRP %s", fmt.Sprintf(resourceIdentifierKeyFormat, resourceIdentifier.Group, resourceIdentifier.Version, resourceIdentifier.Kind, resourceIdentifier.Name, resourceIdentifier.Namespace), crpName)
```
michaelawyu, line 123: Hi Arvind! The format param order seems to be wrong? It should be GVK -> namespace -> name.

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,244 @@
...
+		return false, fmt.Errorf("failed to cordon member cluster %s: %w", h.ClusterName, err)
+	}
+
+	if err := h.fetchClusterResourcePlacementToEvict(ctx); err != nil {
+		return false, err
+	}
+
+	if len(h.ClusterResourcePlacementResourcesMap) == 0 {
+		log.Printf("There are currently no resources propagated to %s from fleet using ClusterResourcePlacement resources", h.ClusterName)
+		return true, nil
```
michaelawyu, line 50: Hi Arvind! Now the cluster is cordoned for no reason; at least we need to remind user to uncordon this.

michaelawyu: A (much) better way is to put the check front I assume.

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,244 @@
...
+		if err != nil {
+			return false, fmt.Errorf("failed to wait for evictions to reach terminal state: %w", err)
+		}
+	}
+
+	isDrainSuccessful := true
+	for crpName := range h.ClusterResourcePlacementResourcesMap {
+		evictionName := fmt.Sprintf(drainEvictionNameFormat, crpName, h.ClusterName)
+		eviction := placementv1beta1.ClusterResourcePlacementEviction{}
+		if err := h.HubClient.Get(ctx, types.NamespacedName{Name: evictionName}, &eviction); err != nil {
```
michaelawyu, line 101: Just a nit: this GET can be saved with some refactoring.

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,244 @@
...
+		if validCondition != nil && validCondition.Status == metav1.ConditionFalse {
+			// check to see if CRP is missing or CRP is being deleted or CRB is missing.
+			if validCondition.Reason == condition.EvictionInvalidMissingCRPMessage ||
+				validCondition.Reason == condition.EvictionInvalidDeletingCRPMessage ||
+				validCondition.Reason == condition.EvictionInvalidMissingCRBMessage {
+				continue
+			}
+		}
+		executedCondition := eviction.GetCondition(string(placementv1beta1.PlacementEvictionConditionTypeExecuted))
+		if executedCondition == nil || executedCondition.Status == metav1.ConditionFalse {
```
michaelawyu, line 114: This part heavily relies on a few of our coding assumptions here (e.g., valid -> executed, valid = false then executed is not set, valid is never unknown, and executed is never unknown).

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,244 @@
...
+	if len(h.ClusterResourcePlacementResourcesMap) == 0 {
+		log.Printf("There are currently no resources propagated to %s from fleet using ClusterResourcePlacement resources", h.ClusterName)
+		return true, nil
+	}
+
+	// create eviction objects for all <crpName, targetCluster>.
+	for crpName := range h.ClusterResourcePlacementResourcesMap {
+		evictionName := fmt.Sprintf(drainEvictionNameFormat, crpName, h.ClusterName)
+
+		if err := removeExistingEviction(ctx, h.HubClient, evictionName); err != nil {
```
michaelawyu, line 57: I have some reservations in this part; eviction objects are left in the cluster after its creation/execution as a track record, now it is being erased every time the tool is called.

michaelawyu: I understand that this part is needed for conflict resolution, but would it be better if we just use a random suffix in name? (considering that the eviction controller can handle this situation)

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,244 @@
...
+		if err := h.HubClient.Get(ctx, types.NamespacedName{Name: evictionName}, &eviction); err != nil {
+			return false, fmt.Errorf("failed to get eviction %s: %w", evictionName, err)
+		}
+		validCondition := eviction.GetCondition(string(placementv1beta1.PlacementEvictionConditionTypeValid))
+		if validCondition != nil && validCondition.Status == metav1.ConditionFalse {
+			// check to see if CRP is missing or CRP is being deleted or CRB is missing.
+			if validCondition.Reason == condition.EvictionInvalidMissingCRPMessage ||
+				validCondition.Reason == condition.EvictionInvalidDeletingCRPMessage ||
+				validCondition.Reason == condition.EvictionInvalidMissingCRBMessage {
+				continue
```
michaelawyu, line 110: Hi Arvind! If I understand it correctly, in these situations the drain is considered successfully but no output is given?

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,244 @@
...
+				}
+				if !evictionutils.IsPlacementPresent(targetBinding) {
+					// need to wait until placement is present on member cluster.
+					return false, nil
+				}
+				return true, nil
+			})
+
+			if err != nil {
+				return fmt.Errorf("failed to wait for placement to be present on member cluster: %w", err)
```
michaelawyu, line 183: Hi Arvind! This is a bit weird too if I read it correctly: if a binding no the cluster cannot get ready (i.e., rollout is stuck), the cluster cannot be drained?

tools/draincluster/main.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if isDrainSuccessful {
+		log.Printf("drain was successful for cluster %s", *clusterName)
+	} else {
+		log.Printf("drain was not successful for cluster %s", *clusterName)
+	}
+
+	log.Printf("retrying drain to ensure all resources propagated from hub cluster are evicted")
+
+	// reset ClusterResourcePlacementResourcesMap for retry.
```
michaelawyu, line 68: Hi Arvind! Is this retry really necessary? If it is added as a consistency guard, it's not 100% bullet proof. And surely doing the same thing twice is confusing.

tools/uncordoncluster/uncordon/uncordon.go
```diff
@@ -0,0 +1,54 @@
...
+		for i := range mc.Spec.Taints {
+			taint := mc.Spec.Taints[i]
+			if taint == toolsutils.CordonTaint {
+				cordonTaintIndex = i
+				break
+			}
+		}
+
+		if cordonTaintIndex >= 0 {
+			mc.Spec.Taints = append(mc.Spec.Taints[:cordonTaintIndex], mc.Spec.Taints[cordonTaintIndex+1:]...)
```
michaelawyu, line 47: A nit: it might be much more straightforward to just reconstruct the array.

michaelawyu (commented):
Added a number of comments, PTAL

tools/draincluster/drain/drain.go
Arvindthiru: We use https://github.com/Azure/fleet/pull/1060/files#diff-6e3b2712c782f83d3836cdf31f77c57fc50d0aa4573f4ffa3147d9aacbbfd981R16 util to safeguard ourselves

tools/draincluster/drain/drain.go
michaelawyu: Hi Arvind! As previously discussed, let's add a log output here mentioning that the cluster has been cordoned with a taint.

tools/draincluster/drain/drain.go
michaelawyu: Hi Arvind! There might be some misunderstanding here; I was trying to suggest that we probably should add a log here mentioning that the CRP is already gone, just for better visibility.

tools/draincluster/drain/drain.go
michaelawyu: Hi Arvind! It's not a blocker per se but I still believe that we should reply on API semantics for validations in this regard.

michaelawyu: Plus, even with the function you mentioned, unknown Valid condition can still pass the check, so are unknown Executed conditions.

michaelawyu (commented):
Hi Arvind! Apologies, I just need a little bit more time to review the code for one more time. Will add approval or more comments later if you do not mind.

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+			if !crpNameMap[crpName] {
+				crpNameMap[crpName] = true
+			}
+		}
+	}
+
+	crpNames := make([]string, len(crpNameMap))
+	i := 0
+	for crpName := range crpNameMap {
```
michaelawyu, line 171: A nit: this cast from map to string array seems redundant (in the sense that the map itself can be used directly anyway); we could also use apimachinery Set if necessary.

tools/draincluster/drain/drain.go
```diff
@@ -0,0 +1,212 @@
...
+	crpNameMap := make(map[string]bool)
+	// find all unique CRP names for which eviction needs to occur.
+	for i := range crbList.Items {
+		crb := crbList.Items[i]
+		if crb.Spec.TargetCluster == h.ClusterName && crb.DeletionTimestamp == nil {
+			crpName, ok := crb.GetLabels()[placementv1beta1.CRPTrackingLabel]
+			if !ok {
+				return []string{}, fmt.Errorf("failed to get CRP name from binding %s", crb.Name)
+			}
+			if !crpNameMap[crpName] {
```
michaelawyu, line 163: A nit: an unnecessary check.

michaelawyu (commented):
Added a few other nits; otherwise LGTM.

tools/draincluster/drain/drain.go
Arvindthiru: Added reminder here https://github.com/Azure/fleet/pull/1060/commits/1885458165bc404dbc21df83c3e61dbca3913312#diff-470da8ac4426d765d81fc981231b957157d9af879602ea08dc24a786311a13aaR77

tools/draincluster/drain/drain.go
Arvindthiru: Added TODO will handle in a follow up PR

michaelawyu (approved):
LGTM ;)

PR #1062, opened 2025-02-28
fix: handle override delete on demand

ryanzhang-oss:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/rollout/controller.go
+++ b/pkg/controllers/rollout/controller.go
@@ -149,7 +149,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 
 	// fill out all the default values for CRP just in case the mutation webhook is not enabled.
 	defaulter.SetDefaultsClusterResourcePlacement(&crp)
-
+	// Note: there is a corner case that an override is in-between snapshots (the old one is marked as not the latest while the new one is not created yet)
+	// This will result in one of the override is removed by the rollout controller so the first instance of the updated cluster can experience
+	// a complete removal of the override effect following by applying the new override effect.
+	// This is hard to avoid as the system is eventual consistent and we would recommend users to use updateRun
 	matchedCRO, matchedRO, err := overrider.FetchAllMatchingOverridesForResourceSnapshot(ctx, r.Client, r.InformerManager, crp.Name, latestResourceSnapshot)
 	if err != nil {
 		klog.ErrorS(err, "Failed to find all matching overrides for the clusterResourcePlacement", "clusterResourcePlacement", crpName)
@@ -709,6 +712,10 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
 				klog.V(2).InfoS("Handling a clusterResourceOverrideSnapshot create event", "clusterResourceOverrideSnapshot", klog.KObj(e.Object))
 				handleClusterResourceOverrideSnapshot(e.Object, q)
 			},
+			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.RateLimitingInterface) {
+				klog.V(2).InfoS("Handling a clusterResourceOverrideSnapshot delete event", "clusterResourceOverrideSnapshot", klog.KObj(e.Object))
+				handleClusterResourceOverrideSnapshot(e.Object, q)
+			},
 			GenericFunc: func(ctx context.Context, e event.GenericEvent, q workqueue.RateLimitingInterface) {
 				klog.V(2).InfoS("Handling a clusterResourceOverrideSnapshot generic event", "clusterResourceOverrideSnapshot", klog.KObj(e.Object))
 				handleClusterResourceOverrideSnapshot(e.Object, q)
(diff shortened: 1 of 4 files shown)


pkg/controllers/rollout/controller.go
```diff
@@ -149,7 +149,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 
 	// fill out all the default values for CRP just in case the mutation webhook is not enabled.
 	defaulter.SetDefaultsClusterResourcePlacement(&crp)
-
+	// Note: there is a corner case that an override is in-between snapshots (the old one is marked as not the latest while the new one is not created yet)
+	// This will result in one of the override is removed by the rollout controller so the first instance of the updated cluster can experience
+	// a complete removal of the override effect following by applying the new override effect.
+	// This is hard to avoid as the system is eventual consistent and we would recommend users to use updateRun
```
zhiying-lin, line 155: For the updateRun, it has the similar issue since cx cannot pick the overrideSnapshot when creating updateRun.
For both cases, cx needs to check the override snapshots before rolling out the new changes and updating the overrides.

Failing this call when it hits this middle state could help the ux.

pkg/controllers/rollout/controller.go
```diff
@@ -709,6 +712,10 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
 				klog.V(2).InfoS("Handling a clusterResourceOverrideSnapshot create event", "clusterResourceOverrideSnapshot", klog.KObj(e.Object))
 				handleClusterResourceOverrideSnapshot(e.Object, q)
 			},
+			DeleteFunc: func(ctx context.Context, e event.DeleteEvent, q workqueue.RateLimitingInterface) {
```
zhiying-lin, line 715: trigger it when cx deletes the resourceOverride or clusterResourceOverride?

The current fix relies on the assumption and current implementation, when we delete the latest snapshot, it means cx is deleting override itself. If we prefer the current one, better to add a comment to record this.

another tiny diff is, if controller fails to delete resourceOverride for some reason, the override revert changes will be applied when the latestSnapshot is deleted.

test/e2e/placement_ro_test.go
```diff
@@ -125,7 +125,7 @@ var _ = Describe("creating resourceOverride (selecting all clusters) to override
 		}
 		crpStatusUpdatedActual := crpStatusWithOverrideUpdatedActual(workResourceIdentifiers(), allMemberClusterNames, "0", nil, wantRONames)
 		// use the long duration to wait until the rollout is finished.
```
zhiying-lin, line 127: remove comment?

test/e2e/actuals_test.go
```diff
@@ -103,6 +103,21 @@ func validateOverrideAnnotationOfConfigMapOnCluster(cluster *framework.Cluster,
...
+func validateConfigMapNoAnnotationKeyOnCluster(cluster *framework.Cluster, key string) error {
+	workNamespaceName := fmt.Sprintf(workNamespaceNameTemplate, GinkgoParallelProcess())
+	appConfigMapName := fmt.Sprintf(appConfigMapNameTemplate, GinkgoParallelProcess())
+
+	configMap := &corev1.ConfigMap{}
+	if err := cluster.KubeClient.Get(ctx, types.NamespacedName{Namespace: workNamespaceName, Name: appConfigMapName}, configMap); err != nil {
+		return err
+	}
+
+	if len(configMap.Annotations) == 0 || configMap.Annotations[key] == "" {
```
zhiying-lin, line 115:
```suggestion
        if _, exist :=  configMap.Annotations[key]; !exist {
```
just in case, annotation value is empty but key is there.

test/e2e/placement_cro_test.go
```diff
@@ -144,9 +143,24 @@ var _ = Describe("creating clusterResourceOverride (selecting all clusters) to o
...
+	It("delete the cro attached to this CRP", func() {
+		cleanupClusterResourceOverride(croName)
+	})
+
+	It("should update CRP status on demand to not select any override", func() {
+		crpStatusUpdatedActual := crpStatusWithOverrideUpdatedActual(workResourceIdentifiers(), allMemberClusterNames, "0", nil, nil)
+		Eventually(crpStatusUpdatedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to update CRP %s status as expected", crpName)
+	})
+
+	It("should not have annotations on the configmap", func() {
```
zhiying-lin, line 156: make sure the namespace itself does not have annotations?

pkg/controllers/rollout/controller.go
ryanzhang-oss: I will add a TODO then

pkg/controllers/rollout/controller.go
ryanzhang-oss: Although the diff is small, I thought about it a bit more and I think it's better to watch CRO/RO directly. I was a bit concerned that the rollout controller was watching too many types of resources but I think it should be okey.

test/e2e/placement_cro_test.go
```diff
@@ -144,9 +143,25 @@ var _ = Describe("creating clusterResourceOverride (selecting all clusters) to o
 		want := map[string]string{croTestAnnotationKey: croTestAnnotationValue1}
 		checkIfOverrideAnnotationsOnAllMemberClusters(true, want)
 	})
+
+	It("delete the cro attached to this CRP", func() {
```
zhiying-lin, line 147: do you want to move to the AfterAll to make it consistent with ro test?

zhiying-lin (approved):

test/e2e/placement_cro_test.go
ryanzhang-oss: I kept it this way to try both types. The original goal is to add that in all BeforeAll but I can't put all the context inside one describe.

PR #1069, opened 2025-03-06
feat: CRP Eviction completion metric

Arvindthiru:

image

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterresourceplacementeviction/controller.go
+++ b/pkg/controllers/clusterresourceplacementeviction/controller.go
@@ -70,11 +70,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 		return runtime.Result{}, err
 	}
 	if !validationResult.isValid {
-		err = r.updateEvictionStatus(ctx, &eviction)
-		if err != nil {
+		if err = r.updateEvictionStatus(ctx, &eviction); err != nil {
 			internalError = err
+			return runtime.Result{}, err
 		}
-		return runtime.Result{}, err
 	}
 
 	markEvictionValid(&eviction)
@@ -88,7 +87,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	if err != nil {
 		internalError = err
 	}
-	return runtime.Result{}, r.updateEvictionStatus(ctx, &eviction)
+	return runtime.Result{}, err
 }
 
 // validateEviction performs validation for eviction object's spec and returns a wrapped validation result.
--- a/pkg/controllers/clusterresourceplacementeviction/controller.go
+++ b/pkg/controllers/clusterresourceplacementeviction/controller.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -24,6 +25,7 @@ import (
 	bindingutils "go.goms.io/fleet/pkg/utils/binding"
 	"go.goms.io/fleet/pkg/utils/condition"
 	"go.goms.io/fleet/pkg/utils/controller"
+	"go.goms.io/fleet/pkg/utils/controller/metrics"
 	"go.goms.io/fleet/pkg/utils/defaulter"
 )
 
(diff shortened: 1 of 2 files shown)


britaniar (approved):

britaniar (dismissed):

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -348,7 +369,14 @@ func (r *Reconciler) SetupWithManager(mgr runtime.Manager) error {
 	return runtime.NewControllerManagedBy(mgr).Named("clusterresourceplacementeviction-controller").
 		WithOptions(ctrl.Options{MaxConcurrentReconciles: 1}). // max concurrent reconciles is currently set to 1 for concurrency control.
 		For(&placementv1beta1.ClusterResourcePlacementEviction{}).
-		Complete(r)
+		WithEventFilter(predicate.Funcs{
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				// delete complete status metric for eviction and skip reconciliation.
+				klog.V(2).InfoS("ClusterResourcePlacementEviction is being deleted", "clusterResourcePlacementEviction", e.Object.GetName())
+				metrics.FleetEvictionStatus.DeletePartialMatch(prometheus.Labels{"name": e.Object.GetName()})
```
ryanzhang-oss, line 376:
```suggestion
			count :=	metrics.FleetEvictionStatus.DeletePartialMatch(prometheus.Labels{"name": e.Object.GetName()})
klog.V(2).InfoS("ClusterResourcePlacementEviction is being deleted", "clusterResourcePlacementEviction", e.Object.GetName(), "count", count)
				```

Comment (ryanzhang-oss) [reply pkg/controllers/clusterresourceplacementeviction/controller.go:376]: is the count always 1?  Why do we need to delete it?

File: pkg/controllers/clusterresourceplacementeviction/controller.go [hunk 2]
```diff
@@ -39,36 +43,53 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	startTime := time.Now()
 	evictionName := req.NamespacedName.Name
 	klog.V(2).InfoS("ClusterResourcePlacementEviction reconciliation starts", "clusterResourcePlacementEviction", evictionName)
+	var internalError error
```
ryanzhang-oss, line 46: looks like we used this as a boolean?

pkg/controllers/clusterresourceplacementeviction/controller.go
Arvindthiru: The count should ideally be 1 always.

we log metric for complete and incomplete evictions,

For example,

When eviction is incomplete we log this metric and keep updating it as long as it stays incomplete
fleet_workload_eviction_complete{isCompleted="false",name="test-eviction-1"} 99 -> 100

Once eviction completes, we delete metric with isCompleted=false and create new metric with isCompleted=true
fleet_workload_eviction_complete{isCompleted="false",name="test-eviction-1"} 99 -> 100 (Deleted)
fleet_workload_eviction_complete{isCompleted="true",name="test-eviction-1"} 102

Then once we delete eviction we no longer care about the metric which has the same name, but the metric could have isCompleted set to true/false hence we do a partial match.

pkg/utils/controller/metrics/metrics.go
```diff
@@ -54,6 +54,13 @@ var (
 		Name: "fleet_workload_placement_complete",
 		Help: "Placement complete status ",
 	}, []string{"name"})
+
+	// FleetEvictionStatus is prometheus metrics which holds the
+	// status of eviction completion.
+	FleetEvictionStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "fleet_workload_eviction_complete",
+		Help: "Eviction complete status ",
+	}, []string{"name", "isCompleted"})
```
ryanzhang-oss, line 63: I think it will be very helpful to have another label about if it's a user error

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -39,36 +43,58 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
...
 	}()
 
 	var eviction placementv1beta1.ClusterResourcePlacementEviction
 	if err := r.Client.Get(ctx, req.NamespacedName, &eviction); err != nil {
+		internalError = true
 		klog.ErrorS(err, "Failed to get cluster resource placement eviction", "clusterResourcePlacementEviction", evictionName)
 		return runtime.Result{}, client.IgnoreNotFound(err)
 	}
 
 	if isEvictionInTerminalState(&eviction) {
```
ryanzhang-oss, line 62: if the previous update succeeded then why would we reconcile again to come here?

Arvindthiru: Due to the status update on eviciton we reconile again, but moved the metric emission to when we originally update status instead

ryanzhang-oss (approved):

PR #1071, opened 2025-03-07
feat: apply the secret/configmap before workloads

ryanzhang-oss:

1. Reorder the apply list to apply secret/configMap first
2. Remove the parallelism from the workapplier

Fixes #

I have:


pkg/controllers/clusterresourceplacement/resource_selector_test.go
```diff
@@ -881,18 +952,36 @@ func TestSortResource(t *testing.T) {
...
 
 	for testName, tt := range tests {
 		t.Run(testName, func(t *testing.T) {
-			sortResources(tt.resources)
-
-			// Check that the returned resources match the expected resources
-			diff := cmp.Diff(tt.want, tt.resources)
-			if diff != "" {
-				t.Errorf("sortResources() mismatch (-want +got):\n%s", diff)
+			// run many times to make sure it's stable
```
michaelawyu, line 977: Hi Ryan! `sort.Slice` does not guarantee stableness IIRC; shall we switch to `sort.SliceStable` instead?

michaelawyu (commented):
Added some comments, PTAL

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -101,30 +101,38 @@ func (r *Reconciler) gatherSelectedResource(placement string, selectors []fleetv
...
+		// compare group/version;kind for the rest of type of resources
+		gvkComp := strings.Compare(gvk1, gvk2)
 		if gvkComp == 0 {
 			// same gvk, compare namespace/name
 			return strings.Compare(fmt.Sprintf("%s/%s", obj1.GetNamespace(), obj1.GetName()),
 				fmt.Sprintf("%s/%s", obj2.GetNamespace(), obj2.GetName())) > 0
 		}
-		return gvkComp > 0
+		// sort by the cluster scoped priority resource types first
+		if gvk1 == utils.NamespaceMetaGVK.String() || gvk2 == utils.NamespaceMetaGVK.String() {
```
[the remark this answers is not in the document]
ryanzhang-oss: > Hi Ryan! Since for the prioritized types the `gvkComp` results won't be needed at all, would it make more sense to put the object type check at the beginning instead? These checks need to be performed anyway,

not really, if the gvk is not checked first to make sure o1 and o2 don't have the same gvk, each block below needs to check again

pkg/controllers/clusterresourceplacement/resource_selector_test.go
ryanzhang-oss: the difference between SliceStable is only if there are identical objects in the array which is not possible in our case as a ETCD guarantee.

test/e2e/placement_selecting_resources_test.go
```diff
@@ -1185,6 +1186,145 @@ var _ = Describe("validating CRP when selected resources cross the 1MB limit", O
 	})
 })
 
+// This test verifies that resources are selected and ordered correctly according to the sortResources logic
+var _ = Describe("creating CRP and checking selected resources order", Ordered, func() {
+	crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+	nsName := fmt.Sprintf("test-ns-order-%d", GinkgoParallelProcess())
```
michaelawyu, line 1192: Hi Ryan! Just a nit: the NS part could use the shared `appNamespace()` utility/`workNamespaceNameTemplate` variable/`workResourceSelector()` utility to simplify some of the setup.

pkg/controllers/clusterresourceplacement/resource_selector.go
```diff
@@ -101,30 +101,38 @@ func (r *Reconciler) gatherSelectedResource(placement string, selectors []fleetv
 
 	return resources, nil
 }
+
```
michaelawyu, line 104: Hi Ryan! Do we want to enforce this order for enveloped resources as well?

michaelawyu (commented):
Added some comments, PTAL

test/e2e/placement_selecting_resources_test.go
ryanzhang-oss: actully I found that e2e has to use the appNamespace() as the clean up assumed that. Bummer.

michaelawyu (approved):
LGTM

PR #1072, opened 2025-03-08
feat: check format of enveloped objects and block cross namespace placement

ryanzhang-oss:

1. check the format of enveloped objects 
2. block cross namespace placement

Fixes #

I have:


pkg/controllers/workgenerator/controller.go
```diff
@@ -1245,11 +1245,20 @@ func extractResFromConfigMap(uConfigMap *unstructured.Unstructured) ([]fleetv1be
...
-	for _, value := range configMap.Data {
+	for key, value := range configMap.Data {
+		// so we need to check the GVK and annotation of the selected resource
 		content, jsonErr := yaml.ToJSON([]byte(value))
 		if jsonErr != nil {
 			return nil, jsonErr
 		}
+		var uManifest unstructured.Unstructured
+		if unMarshallErr := uManifest.UnmarshalJSON(content); unMarshallErr != nil {
+			klog.ErrorS(unMarshallErr, "manifest has invalid content", "manifestKey", key, "envelopResource", klog.KObj(uConfigMap))
```
jwtty, line 1256:
```suggestion
			klog.ErrorS(unMarshallErr, "manifest has invalid content", "manifestKey", key, "envelopeResource", klog.KObj(uConfigMap))
```

pkg/controllers/workgenerator/controller.go
```diff
@@ -1245,11 +1245,20 @@ func extractResFromConfigMap(uConfigMap *unstructured.Unstructured) ([]fleetv1be
...
+			return nil, fmt.Errorf("the object with manifest key `%s` in evenlop config `%s` is malformatted, err: %w", key, klog.KObj(uConfigMap), unMarshallErr)
```
jwtty, line 1257:
```suggestion
			return nil, fmt.Errorf("the object with manifest key `%s` in envelope config `%s` is malformatted, err: %w", key, klog.KObj(uConfigMap), unMarshallErr)
```

pkg/controllers/workgenerator/controller.go
```diff
@@ -1245,11 +1245,20 @@ func extractResFromConfigMap(uConfigMap *unstructured.Unstructured) ([]fleetv1be
...
+		}
+		if len(uManifest.GetNamespace()) != 0 && uManifest.GetNamespace() != configMap.Namespace {
+			return nil, fmt.Errorf("the namespaced object `%s` in evenlop config `%s` is placed in a different namespace `%s` ", uManifest.GetName(), klog.KObj(uConfigMap), uManifest.GetNamespace())
```
jwtty, line 1260:
```suggestion
			return nil, fmt.Errorf("the namespaced object `%s` in envelope config `%s` is placed in a different namespace `%s` ", uManifest.GetName(), klog.KObj(uConfigMap), uManifest.GetNamespace())
```

jwtty (commented):
Other than the nits, lgtm.

pkg/controllers/workgenerator/controller.go
zhiying-lin, line 1260:
```suggestion
			return nil, fmt.Errorf("placing a namespaced object `%s` using evenlop config `%s` in a different namespace `%s`  is not allowed", uManifest.GetName(), klog.KObj(uConfigMap), uManifest.GetNamespace())
```

pkg/controllers/workgenerator/controller_test.go
```diff
@@ -148,6 +147,150 @@ func TestGetWorkNamePrefixFromSnapshotName(t *testing.T) {
...
+			wantErr: false,
+		},
+	}
+
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			got, err := extractResFromConfigMap(tt.uConfigMap)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("extractResFromConfigMap() error = %v, wantErr %v", err, tt.wantErr)
+				return
```
zhiying-lin, line 285:
```suggestion
				t.Fatalf("extractResFromConfigMap() error = %v, wantErr %v", err, tt.wantErr)
```
can use fatalf instead

pkg/controllers/workgenerator/controller.go
```diff
@@ -1245,11 +1245,20 @@ func extractResFromConfigMap(uConfigMap *unstructured.Unstructured) ([]fleetv1be
 		return nil, err
 	}
 	// the list order is not stable as the map traverse is random
-	for _, value := range configMap.Data {
```
zhiying-lin, line 1248: not sure what happens when configMap.Data is empty.

pkg/controllers/workgenerator/controller.go
```diff
@@ -1245,11 +1245,20 @@ func extractResFromConfigMap(uConfigMap *unstructured.Unstructured) ([]fleetv1be
...
+		// so we need to check the GVK and annotation of the selected resource
 		content, jsonErr := yaml.ToJSON([]byte(value))
 		if jsonErr != nil {
 			return nil, jsonErr
 		}
+		var uManifest unstructured.Unstructured
+		if unMarshallErr := uManifest.UnmarshalJSON(content); unMarshallErr != nil {
+			klog.ErrorS(unMarshallErr, "manifest has invalid content", "manifestKey", key, "envelopResource", klog.KObj(uConfigMap))
+			return nil, fmt.Errorf("the object with manifest key `%s` in evenlop config `%s` is malformatted, err: %w", key, klog.KObj(uConfigMap), unMarshallErr)
+		}
```
michaelawyu, line 1258: Hi Ryan! Just a nit here: would it be good if we add a comment here mentioning that namespaced objects that do not have a namespace value specified would pass the check here but fail at the work applier end?

jwtty: So they will not have assigned the same namespace as the configMap?

ryanzhang-oss: we won't assign the namespace since we don't know if it's a namespace scoped or cluster scoped object

pkg/controllers/workgenerator/controller.go
ryanzhang-oss: it's okey for now and there is a test for this case.

jwtty (approved):
LGTM, thank you.

PR #1074, opened 2025-03-11
feat: support arbitrary key on the cluster as the value of override

ryanzhang-oss:

Allow user to use the value of an arbitrary key on the cluster as the override
example 
${MEMBER-CLUSTER-LABEL-KEY-fleet.azure.com/location-region_public} 

would be replaced by the value of the key fleet.azure.com/location-region_public on the cluster

Fixes #

I have:

### DIFF
--- a/apis/placement/v1alpha1/common.go
+++ b/apis/placement/v1alpha1/common.go
@@ -18,10 +18,12 @@ const (
 	// OverrideClusterNameVariable is the reserved variable in the override value that will be replaced by the actual cluster name.
 	OverrideClusterNameVariable = "${MEMBER-CLUSTER-NAME}"
 
-	// OverrideClusterLabelKeyVariablePrefix is the reserved variable prefix in the override expression.
-	// The string will be replaced by the actual label value associated with the key following the prefix.
+	// OverrideClusterLabelKeyVariablePrefix is a reserved variable in the override expression.
+	// We use this variable to find the associated the key following the prefix.
 	// The key name ends with a "}" character (but not include it).
-	// The key name must be a valid label name and case-sensitive.
-	// For example, if the key name is "kube-fleet.io/region", the variable will be "${MEMBER-CLUSTER-LABEL-KEY-kube-fleet.io/region}"
+	// The key name must be a valid Kubernetes label name and case-sensitive.
+	// The content of the string containing this variable will be replaced by the actual label value on the member cluster.
+	// For example, if the string is "${MEMBER-CLUSTER-LABEL-KEY-kube-fleet.io/region}" then the key name is "kube-fleet.io/region".
+	// If there is a label "kube-fleet.io/region": "us-west-1" on the member cluster, this string will be replaced by "us-west-1".
 	OverrideClusterLabelKeyVariablePrefix = "${MEMBER-CLUSTER-LABEL-KEY-"
 )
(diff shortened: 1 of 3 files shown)


pkg/controllers/workgenerator/override_test.go
```diff
@@ -2496,3 +2759,88 @@ func TestApplyJSONPatchOverride(t *testing.T) {
 		})
 	}
 }
+
+func Test_replaceClusterLabelKeyVariables(t *testing.T) {
```
zhiying-lin, line 2763:
```suggestion
func TestReplaceClusterLabelKeyVariables(t *testing.T) {
```
nit

apis/placement/v1alpha1/common.go
```diff
@@ -17,4 +17,11 @@ const (
 
 	// OverrideClusterNameVariable is the reserved variable in the override value that will be replaced by the actual cluster name.
 	OverrideClusterNameVariable = "${MEMBER-CLUSTER-NAME}"
+
+	// OverrideClusterLabelKeyVariablePrefix is the reserved variable prefix in the override expression.
+	// The string will be replaced by the actual label value with the given the name of the key following the prefix.
```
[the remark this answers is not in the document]
zhiying-lin: need to add a comment to indicate that the key is case-sensitive.

pkg/controllers/workgenerator/override_test.go
```diff
@@ -2496,3 +2759,88 @@ func TestApplyJSONPatchOverride(t *testing.T) {
...
+		},
+		"ClusterLabelKey Variable replaced": {
+			cluster: &clusterv1beta1.MemberCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
+						"region": "us-west-1",
+					},
+				},
+			},
+			input:    "The cluster is in ${MEMBER-CLUSTER-LABEL-KEY-region}",
```
zhiying-lin, line 2789: what if the input is "The cluster is in ${MEMBER-CLUSTER-LABEL-KEY-}"

ryanzhang-oss: then the key is empty which means it will throw an error, case added.

pkg/controllers/workgenerator/override_test.go
```diff
@@ -2819,6 +2819,17 @@ func Test_replaceClusterLabelKeyVariables(t *testing.T) {
 			input:     "The cluster is in ${MEMBER-CLUSTER-LABEL-KEY-region}",
 			expectErr: true,
 		},
+		"ClusterLabelKey Variable key case not match": {
```
zhiying-lin, line 2822: what if the key is ${member-cluster-label-key-region}?

apis/placement/v1alpha1/common.go
```diff
@@ -17,4 +17,11 @@ const (
 
 	// OverrideClusterNameVariable is the reserved variable in the override value that will be replaced by the actual cluster name.
 	OverrideClusterNameVariable = "${MEMBER-CLUSTER-NAME}"
+
+	// OverrideClusterLabelKeyVariablePrefix is the reserved variable prefix in the override expression.
+	// The string will be replaced by the actual label value associated with the key following the prefix.
+	// The key name ends with a "}" character (but not include it).
+	// The key name must be a valid label name and case-sensitive.
```
zhiying-lin, line 24:
```suggestion
	// The key name must be a valid member cluster label name and case-sensitive.
```
indicate it's member cluster label?

ryanzhang-oss: I think the emphasis here is on its format, I added more verbiage.

jwtty (approved):
LGTM

PR #1077, opened 2025-03-12
feat: Add placement status metrics

britaniar:

Fixes #

I have: added a metric to emit crp condition status' along with a test.

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -1049,7 +1049,6 @@ func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
 		if cond != nil {
 			status = string(cond.Status)
 		}
-		metrics.FleetPlacementStatusLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name, "generation": strconv.FormatInt(crp.Generation, 10), "conditionType": string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType)})
 		metrics.FleetPlacementStatusLastTimeStampSeconds.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType), status).SetToCurrentTime()
 		return
 	}
@@ -1062,7 +1061,6 @@ func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
 			if cond != nil {
 				status = string(cond.Status)
 			}
-			metrics.FleetPlacementStatusLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name, "generation": strconv.FormatInt(crp.Generation, 10), "conditionType": string(condType.ClusterResourcePlacementConditionType())})
 			metrics.FleetPlacementStatusLastTimeStampSeconds.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(condType.ClusterResourcePlacementConditionType()), status).SetToCurrentTime()
 			return
 		}
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -1049,6 +1049,7 @@ func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
 		if cond != nil {
 			status = string(cond.Status)
 		}
+		metrics.FleetPlacementStatusLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name, "generation": strconv.FormatInt(crp.Generation, 10), "conditionType": string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType)})
 		metrics.FleetPlacementStatusLastTimeStampSeconds.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType), status).SetToCurrentTime()
 		return
 	}
@@ -1061,6 +1062,7 @@ func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
 			if cond != nil {
 				status = string(cond.Status)
 			}
+			metrics.FleetPlacementStatusLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name, "generation": strconv.FormatInt(crp.Generation, 10), "conditionType": string(condType.ClusterResourcePlacementConditionType())})
 			metrics.FleetPlacementStatusLastTimeStampSeconds.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(condType.ClusterResourcePlacementConditionType()), status).SetToCurrentTime()
 			return
 		}
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -226,13 +226,11 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			klog.V(2).InfoS("Placement has finished the rollout process and reached the desired status", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 			r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Placement has finished the rollout process and reached the desired status")
 		}
-		metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 		metrics.FleetPlacementComplete.WithLabelValues(crp.Name, "true").SetToCurrentTime()
 		// We don't need to requeue any request now by watching the binding changes
 		return ctrl.Result{}, nil
 	}
 
-	metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 	metrics.FleetPlacementComplete.WithLabelValues(crp.Name, "false").SetToCurrentTime()
 
 	if !isClusterScheduled {
@@ -1046,19 +1044,19 @@ func isCRPScheduled(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
 	// Check CRP Scheduled condition
 	cond := crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType))
-	if cond != nil {
+	if !condition.IsConditionStatusTrue(cond, crp.Generation) && cond != nil {
 		metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), cond.Type, string(cond.Status)).SetToCurrentTime()
+		return
 	}
 	// Check CRP expected conditions
 	expectedCondTypes := determineExpectedCRPAndResourcePlacementStatusCondType(crp)
 	for _, i := range expectedCondTypes {
 		cond := crp.GetCondition(string(i.ClusterResourcePlacementConditionType()))
-		if cond != nil { // emit if found
+		if !condition.IsConditionStatusTrue(cond, crp.Generation) && cond != nil { // emit the first not true condition
 			metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 			metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), cond.Type, string(cond.Status)).SetToCurrentTime()
-		} else {
-			return // return if not found
+			return
 		}
 	}
 }
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -101,7 +101,8 @@ func (r *Reconciler) handleDelete(ctx context.Context, crp *fleetv1beta1.Cluster
 	}
 	klog.V(2).InfoS("Removed crp-cleanup finalizer", "clusterResourcePlacement", crpKObj)
 	r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementCleanupFinalizerRemoved", "Deleted the snapshots and removed the placement cleanup finalizer")
-	metrics.FleetPlacementStatus.Delete(prometheus.Labels{"name": crp.Name})
+	metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+	metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 	return ctrl.Result{}, nil
 }
 
@@ -177,6 +178,8 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			klog.ErrorS(updateErr, "Failed to update the status", "clusterResourcePlacement", crpKObj)
 			return ctrl.Result{}, controller.NewUpdateIgnoreConflictError(updateErr)
 		}
+		metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), scheduleCondition.Type, string(scheduleCondition.Status)).SetToCurrentTime()
 		return ctrl.Result{}, err
 	}
 
@@ -214,6 +217,7 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			r.Recorder.Event(crp, corev1.EventTypeNormal, i.EventReasonForTrue(), i.EventMessageForTrue())
 		}
 	}
+	emitPlacementStatusMetric(crp)
 
 	// Rollout is considered to be completed when all the expected condition types are set to the
 	// True status.
@@ -222,12 +226,15 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			klog.V(2).InfoS("Placement has finished the rollout process and reached the desired status", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 			r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Placement has finished the rollout process and reached the desired status")
 		}
-		metrics.FleetPlacementStatus.WithLabelValues(crp.Name).Set(1)
+		metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+		metrics.FleetPlacementComplete.WithLabelValues(crp.Name, "true").SetToCurrentTime()
 		// We don't need to requeue any request now by watching the binding changes
 		return ctrl.Result{}, nil
 	}
 
-	metrics.FleetPlacementStatus.WithLabelValues(crp.Name).Set(0)
+	metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+	metrics.FleetPlacementComplete.WithLabelValues(crp.Name, "false").SetToCurrentTime()
+
 	if !isClusterScheduled {
 		// Note:
 		// If the scheduledCondition is failed, it means the placement requirement cannot be satisfied fully. For example,
@@ -1034,4 +1041,24 @@ func isRolloutCompleted(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 
 func isCRPScheduled(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 	return condition.IsConditionStatusTrue(crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType)), crp.Generation)
+}
+
+func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
+	// Check CRP Scheduled condition
+	cond := crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType))
+	if cond != nil {
+		metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), cond.Type, string(cond.Status)).SetToCurrentTime()
+	}
+	// Check CRP expected conditions
+	expectedCondTypes := determineExpectedCRPAndResourcePlacementStatusCondType(crp)
+	for _, i := range expectedCondTypes {
+		cond := crp.GetCondition(string(i.ClusterResourcePlacementConditionType()))
+		if cond != nil { // emit if found
+			metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+			metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), cond.Type, string(cond.Status)).SetToCurrentTime()
+		} else {
+			return // return if not found
+		}
+	}
 }
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -94,6 +94,8 @@ func (r *Reconciler) handleDelete(ctx context.Context, crp *fleetv1beta1.Cluster
 		return ctrl.Result{}, err
 	}
 
+	metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+	metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 	controllerutil.RemoveFinalizer(crp, fleetv1beta1.ClusterResourcePlacementCleanupFinalizer)
 	if err := r.Client.Update(ctx, crp); err != nil {
 		klog.ErrorS(err, "Failed to remove crp finalizer", "clusterResourcePlacement", crpKObj)
@@ -101,8 +103,6 @@ func (r *Reconciler) handleDelete(ctx context.Context, crp *fleetv1beta1.Cluster
 	}
 	klog.V(2).InfoS("Removed crp-cleanup finalizer", "clusterResourcePlacement", crpKObj)
 	r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementCleanupFinalizerRemoved", "Deleted the snapshots and removed the placement cleanup finalizer")
-	metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
-	metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 	return ctrl.Result{}, nil
 }
 
@@ -1042,20 +1042,27 @@ func isCRPScheduled(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 }
 
 func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
+	metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 	// Check CRP Scheduled condition
+	status := "nil"
 	cond := crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType))
-	if !condition.IsConditionStatusTrue(cond, crp.Generation) && cond != nil {
-		metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
-		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), cond.Type, string(cond.Status)).SetToCurrentTime()
+	if !condition.IsConditionStatusTrue(cond, crp.Generation) {
+		if cond != nil {
+			status = string(cond.Status)
+		}
+		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType), status).SetToCurrentTime()
 		return
 	}
+
 	// Check CRP expected conditions
 	expectedCondTypes := determineExpectedCRPAndResourcePlacementStatusCondType(crp)
-	for _, i := range expectedCondTypes {
-		cond := crp.GetCondition(string(i.ClusterResourcePlacementConditionType()))
-		if !condition.IsConditionStatusTrue(cond, crp.Generation) && cond != nil { // emit the first not true condition
-			metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
-			metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), cond.Type, string(cond.Status)).SetToCurrentTime()
+	for _, condType := range expectedCondTypes {
+		cond := crp.GetCondition(string(condType.ClusterResourcePlacementConditionType()))
+		if !condition.IsConditionStatusTrue(cond, crp.Generation) {
+			if cond != nil {
+				status = string(cond.Status)
+			}
+			metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(condType.ClusterResourcePlacementConditionType()), status).SetToCurrentTime()
 			return
 		}
 	}
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -94,8 +94,8 @@ func (r *Reconciler) handleDelete(ctx context.Context, crp *fleetv1beta1.Cluster
 		return ctrl.Result{}, err
 	}
 
-	metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
-	metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+	metrics.FleetPlacementCompleteLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+	metrics.FleetPlacementStatusLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 	controllerutil.RemoveFinalizer(crp, fleetv1beta1.ClusterResourcePlacementCleanupFinalizer)
 	if err := r.Client.Update(ctx, crp); err != nil {
 		klog.ErrorS(err, "Failed to remove crp finalizer", "clusterResourcePlacement", crpKObj)
@@ -178,8 +178,8 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			klog.ErrorS(updateErr, "Failed to update the status", "clusterResourcePlacement", crpKObj)
 			return ctrl.Result{}, controller.NewUpdateIgnoreConflictError(updateErr)
 		}
-		metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
-		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), scheduleCondition.Type, string(scheduleCondition.Status)).SetToCurrentTime()
+		metrics.FleetPlacementStatusLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+		metrics.FleetPlacementStatusLastTimeStampSeconds.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), scheduleCondition.Type, string(scheduleCondition.Status)).SetToCurrentTime()
 		return ctrl.Result{}, err
 	}
 
@@ -226,12 +226,12 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			klog.V(2).InfoS("Placement has finished the rollout process and reached the desired status", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 			r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Placement has finished the rollout process and reached the desired status")
 		}
-		metrics.FleetPlacementComplete.WithLabelValues(crp.Name, "true").SetToCurrentTime()
+		metrics.FleetPlacementCompleteLastTimeStampSeconds.WithLabelValues(crp.Name, "true").SetToCurrentTime()
 		// We don't need to requeue any request now by watching the binding changes
 		return ctrl.Result{}, nil
 	}
 
-	metrics.FleetPlacementComplete.WithLabelValues(crp.Name, "false").SetToCurrentTime()
+	metrics.FleetPlacementCompleteLastTimeStampSeconds.WithLabelValues(crp.Name, "false").SetToCurrentTime()
 
 	if !isClusterScheduled {
 		// Note:
@@ -1042,7 +1042,6 @@ func isCRPScheduled(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 }
 
 func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
-	metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
 	// Check CRP Scheduled condition
 	status := "nil"
 	cond := crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType))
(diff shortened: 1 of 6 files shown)


zhiying-lin (commented):
Should we emit the last condition if the crp is not completed? since your metrics type is gauge.  WE only need to know the current status.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -222,12 +223,12 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			klog.V(2).InfoS("Placement has finished the rollout process and reached the desired status", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 			r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Placement has finished the rollout process and reached the desired status")
 		}
-		metrics.FleetPlacementStatus.WithLabelValues(crp.Name).Set(1)
+		metrics.FleetPlacementComplete.WithLabelValues(crp.Name).SetToCurrentTime()
```
zhiying-lin, line 226: i remember i never see people emit the cpu metrics with the timestamp. Could you please investigate how people create the gauge metrics?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -1025,13 +1026,26 @@ func isRolloutCompleted(crp *fleetv1beta1.ClusterResourcePlacement) bool {
...
 		}
 	}
 	return true
 }
 
 func isCRPScheduled(crp *fleetv1beta1.ClusterResourcePlacement) bool {
-	return condition.IsConditionStatusTrue(crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType)), crp.Generation)
+	cond := crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType))
+	if cond != nil {
+		emitPlacementStatusMetric(cond, crp)
```
zhiying-lin, line 1043: there are multiple ways to create the status metrics,

for example, 
the label is, crp name, generation, value is status for that generation.

could you please investigate how others emit metrics?

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: I talked to Arvind about this too as my reaction was exactly the same "I  never see people emit metrics with the timestamp" but it turned out there is actually a reason for this

ryanzhang-oss:
please see https://github.com/Azure/fleet/pull/1083/files#diff-fe7cf80f71bd1a1911f8703979980085428e3e1c81175394c3ea3f7e7dc350e6 to add tests

pkg/controllers/clusterresourceplacement/controller.go
ryanzhang-oss: the timestamp is there to help to setup alerts for "snapshot" instead of including all the history

pkg/controllers/clusterresourceplacement/controller.go
jwtty: I wonder if including the crp name can result in cardinality being too large?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -1035,3 +1042,23 @@ func isRolloutCompleted(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 func isCRPScheduled(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 	return condition.IsConditionStatusTrue(crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType)), crp.Generation)
 }
+
+func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
+	// Check CRP Scheduled condition
+	cond := crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType))
+	if cond != nil {
+		metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), cond.Type, string(cond.Status)).SetToCurrentTime()
```
britaniar, line 1051: Any thought on setting it as `SetToCurrentTime()` or `Set(float(LastTransitionTime.Unixnano()))`

ryanzhang-oss: it seems that SetToCurrentTime is a built-in? If so, I would prefer that

britaniar: Okay, I will leave as is then.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -222,12 +226,15 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 			klog.V(2).InfoS("Placement has finished the rollout process and reached the desired status", "clusterResourcePlacement", crpKObj, "generation", crp.Generation)
 			r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Placement has finished the rollout process and reached the desired status")
 		}
-		metrics.FleetPlacementStatus.WithLabelValues(crp.Name).Set(1)
+		metrics.FleetPlacementComplete.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
```
jwtty, line 229: Do we really need to delete the label? I think metrics with the same label will be override? Keeping old labels can also help debugging? We can know when crp was incomplete and now complete.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -9,10 +9,14 @@ import (
 	"strconv"
 	"time"
 
+	"go.goms.io/fleet/pkg/utils/controller/metrics"
```
jwtty, line 12: Please reorder this.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -1035,3 +1042,23 @@ func isRolloutCompleted(crp *fleetv1beta1.ClusterResourcePlacement) bool {
...
+	if cond != nil {
+		metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), cond.Type, string(cond.Status)).SetToCurrentTime()
+	}
+	// Check CRP expected conditions
+	expectedCondTypes := determineExpectedCRPAndResourcePlacementStatusCondType(crp)
+	for _, i := range expectedCondTypes {
+		cond := crp.GetCondition(string(i.ClusterResourcePlacementConditionType()))
+		if cond != nil { // emit if found
+			metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
```
jwtty, line 1058: Will this cause previously emitted metrics all deleted? A crp with `RolloutStarted`, `Overriden`, and `Available` 3 conditions, will end up having just the `Available` metrics?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -304,13 +338,22 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
 			}
-			gotCRP = retrieveAndValidateClusterResourcePlacement(testCRPName, wantCRP)
+			gotCRP = retrieveAndValidateClusterResourcePlacement(crp.Name, wantCRP)
 		})
 
 		AfterEach(func() {
 			By("Deleting crp")
 			Expect(k8sClient.Delete(ctx, gotCRP)).Should(Succeed())
 			retrieveAndValidateCRPDeletion(gotCRP)
+			By("Deleting clusterResourceBindings")
```
jwtty, line 348: Suppose these bindings can be deleted automatically with CRP deletion?

pkg/controllers/clusterresourceplacement/controller_test.go
```diff
@@ -3086,31 +3089,37 @@ func TestIsRolloutComplete(t *testing.T) {
 					Status:             metav1.ConditionTrue,
 					Type:               string(fleetv1beta1.ClusterResourcePlacementAppliedConditionType),
 					ObservedGeneration: crpGeneration,
+					LastTransitionTime: metav1.NewTime(currentTime.Add(time.Second * 5)),
```
jwtty, line 3092: Do we still need this change?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: They are manually created because the controller is not connected.

pkg/controllers/clusterresourceplacement/controller.go
britaniar: Yes, zhiying previously commented on how we only need the last condition.

pkg/controllers/clusterresourceplacement/controller.go
britaniar: We have to delete or else we won't have 1 metric emitted. Anytime a label is changed a new metric is emitted and scraped. For example:

When the CRP is reconciling the scraped metrics will be:
fleet_workload_placement_complete{name: <crp-name>, isCompleted: false} <current-time-1>
It will get overwritten every time.

Once the CRP completes and if we do not delete the previous instance the scraped metrics will be:
fleet_workload_placement_complete{name: <crp-name>, isCompleted: false} <current-time-1>
fleet_workload_placement_complete{name: <crp-name>, isCompleted: true} <current-time-2>

Which ideally we don't want, so would be able to "filter" out if we delete any previous instance of the metric with the same CRP name.

pkg/utils/controller/metrics/metrics.go
```diff
@@ -50,10 +50,17 @@ var (
...
-	FleetPlacementStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+	//FleetPlacementComplete is a prometheus metric which keeps track if the placement is complete
+	FleetPlacementComplete = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Name: "fleet_workload_placement_complete",
 		Help: "Placement complete status ",
-	}, []string{"name"})
+	}, []string{"name", "isCompleted"})
+
+	// FleetPlacementStatus is a prometheus metric which keeps track of the placement status
+	FleetPlacementStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
britaniar: Yes, the first one keeps track of completed. The second is only if we need to see the last condition to why it may have failed.

pkg/controllers/clusterresourceplacement/controller_test.go
britaniar: Oh, these changes are needed in this file anymore. Will revert back.

pkg/controllers/clusterresourceplacement/controller.go
jwtty: But I feel keeping both is fine? Indeed, we will only have 2 metrics per crp.

pkg/controllers/clusterresourceplacement/controller.go
zhiying-lin: it's not an efficient way though. can you skip the previous condition and add only the last one? instead of delete and then add.

we need the first !true condition (nil/unknown/false)

jwtty: I guess Zhiying means the first missing/unknown/false condition?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -198,23 +202,54 @@ func createSynchronizedClusterResourceBinding(cluster string, policySnapshot *pl
...
+		ObservedGeneration: binding.Generation,
+	}
+	meta.SetStatusCondition(&binding.Status.Conditions, cond)
+	Expect(k8sClient.Status().Update(ctx, binding)).Should(Succeed(), "Failed to update the binding status")
 }
 
 var _ = Describe("Test ClusterResourcePlacement Controller", func() {
 	Context("When creating new pickAll ClusterResourcePlacement", func() {
+		crpName := fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())
+		var customRegistry *prometheus.Registry
```
jwtty, line 230: This could be moved to below var list.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -343,7 +387,64 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
 				},
 			}
-			retrieveAndValidateClusterResourcePlacement(testCRPName, wantCRP)
+			retrieveAndValidateClusterResourcePlacement(crpName, wantCRP)
+
+			By("Ensure placement complete metric was emitted with isCompleted False")
+			checkPlacementCompleteMetric(customRegistry, crpName, false, 1)
+
+			By("Ensure placement status metric was emitted with Scheduled True")
+			checkPlacementStatusMetric(customRegistry, crp, string(placementv1beta1.ClusterResourcePlacementScheduledConditionType), string(corev1.ConditionUnknown))
```
jwtty, line 396: CRP scheduled condition is true, why the metrics is unknown here?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +640,239 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+				},
+				{
+					ClusterName: member2Name,
+					Selected:    true,
+					Reason:      "valid",
+				},
+			}
+			Expect(k8sClient.Status().Update(ctx, gotPolicySnapshot)).Should(Succeed(), "Failed to update the policy snapshot status")
+
+			By("By creating clusterResourceBinding on member-1")
```
jwtty, line 696:
```suggestion
			By("By creating a synchronized clusterResourceBinding on member-1")
```

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +640,239 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+								},
+							},
+						},
+					},
+				},
+			}
+			gotCRP = retrieveAndValidateClusterResourcePlacement(crpName, wantCRP)
+
+			By("Ensure placement complete metric was emitted with isCompleted True")
+			checkPlacementCompleteMetric(customRegistry, crpName, true, 2)
```
jwtty, line 820: What happens for status metric this case?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: https://github.com/Azure/fleet/pull/1077#discussion_r2028007983

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: There is a flakiness that happens because depending on when I check the metric the last !true condition could be workSychronized or Applied. Considering we are using the PlacementStatus metric mostly for incomplete metrics I thought it would be okay to skip the check.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -65,7 +70,7 @@ var (
 	crpOwnerReference = metav1.OwnerReference{
 		APIVersion:         fleetAPIVersion,
 		Kind:               "ClusterResourcePlacement",
-		Name:               testCRPName,
+		Name:               fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess()),
```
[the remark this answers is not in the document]
britaniar: I need this change to match the CRP names. If it runs in parallel then the metrics will conflict because of the same CRP name. With the CRP name change to `fmt.Sprintf(crpNameTemplate, GinkgoParallelProcess())`, this also needs to be updated to match.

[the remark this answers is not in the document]
britaniar: Oh, I think I made this change because of another conflict of how I arranged the test beforehand. I reverted back.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
jwtty: A bit confused: since crp's condition is Available, why would we see a !true metric? Should we remove the metric when crp becomes complete? Or we say if crp complete metric is set to true, do not bother the status metric at all?

pkg/utils/controller/metrics/metrics.go
```diff
@@ -50,10 +50,17 @@ var (
 		Help: "Number of currently used workers per controller",
 	}, []string{"controller"})
 
+	//FleetPlacementComplete is a prometheus metric which keeps track if the placement is complete
+	FleetPlacementComplete = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "fleet_workload_placement_complete_timestamp_second",
```
jwtty, line 55:
```suggestion
		Name: "fleet_workload_placement_complete_last_timestamp_seconds",
```

pkg/utils/controller/metrics/metrics.go
```diff
@@ -50,10 +50,17 @@ var (
...
+		Help: "Timestamp in seconds of the current completion status of crp. The 'isCompleted' label indicates whether CRP completion is true or false.",
+	}, []string{"name", "isCompleted"})
+
+	// FleetPlacementStatus is a prometheus metric which keeps track of the last placement status
 	FleetPlacementStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "fleet_workload_placement_complete",
-		Help: "Placement complete status ",
-	}, []string{"name"})
+		Name: "fleet_workload_last_placement_status_timestamp_second",
```
jwtty, line 61:
```suggestion
		Name: "fleet_workload_placement_status_last_timestamp_seconds",
```

jwtty (commented):
Other than the nits, LGTM

jwtty (approved):
LGTM

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -1035,3 +1040,30 @@ func isRolloutCompleted(crp *fleetv1beta1.ClusterResourcePlacement) bool {
...
+			status = string(cond.Status)
+		}
+		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType), status).SetToCurrentTime()
+		return
+	}
+
+	// Check CRP expected conditions
+	expectedCondTypes := determineExpectedCRPAndResourcePlacementStatusCondType(crp)
+	for _, condType := range expectedCondTypes {
+		cond := crp.GetCondition(string(condType.ClusterResourcePlacementConditionType()))
```
zhiying-lin, line 1060:
```suggestion
		cond = crp.GetCondition(string(condType.ClusterResourcePlacementConditionType()))
```
duplicate definition

pkg/utils/controller/metrics/metrics.go
```diff
@@ -50,10 +50,17 @@ var (
 		Help: "Number of currently used workers per controller",
 	}, []string{"controller"})
 
+	//FleetPlacementComplete is a prometheus metric which keeps track if the placement is complete
```
zhiying-lin, line 53:
```suggestion
	// FleetPlacementComplete is a prometheus metric which keeps track if the placement is complete.
```

pkg/utils/controller/metrics/metrics.go
```diff
@@ -50,10 +50,17 @@ var (
...
+	FleetPlacementComplete = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
zhiying-lin, line 54:
```suggestion
	FleetPlacementCompleteLastTimeStampSeconds = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
to improve the readability

pkg/utils/controller/metrics/metrics.go
```diff
@@ -50,10 +50,17 @@ var (
...
+		Name: "fleet_workload_placement_complete_last_timestamp_seconds",
+		Help: "Timestamp in seconds of the current completion status of crp. The 'isCompleted' label indicates whether CRP completion is true or false.",
+	}, []string{"name", "isCompleted"})
+
+	// FleetPlacementStatus is a prometheus metric which keeps track of the last placement status
```
zhiying-lin, line 59:
```suggestion
	// FleetPlacementStatus is a prometheus metric which keeps track of the last placement status.
```

pkg/utils/controller/metrics/metrics.go
```diff
@@ -50,10 +50,17 @@ var (
...
 	FleetPlacementStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
zhiying-lin, line 60:
```suggestion
	FleetPlacementStatusLastTimeStampSeconds = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -1035,3 +1040,30 @@ func isRolloutCompleted(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 func isCRPScheduled(crp *fleetv1beta1.ClusterResourcePlacement) bool {
 	return condition.IsConditionStatusTrue(crp.GetCondition(string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType)), crp.Generation)
 }
+
+func emitPlacementStatusMetric(crp *fleetv1beta1.ClusterResourcePlacement) {
+	metrics.FleetPlacementStatus.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
```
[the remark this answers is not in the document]
zhiying-lin: can you add a test to make sure crp should have multiple entries for different generations?

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -1035,3 +1040,30 @@ func isRolloutCompleted(crp *fleetv1beta1.ClusterResourcePlacement) bool {
...
+	if !condition.IsConditionStatusTrue(cond, crp.Generation) {
+		if cond != nil {
+			status = string(cond.Status)
+		}
+		metrics.FleetPlacementStatus.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(fleetv1beta1.ClusterResourcePlacementScheduledConditionType), status).SetToCurrentTime()
+		return
+	}
+
+	// Check CRP expected conditions
+	expectedCondTypes := determineExpectedCRPAndResourcePlacementStatusCondType(crp)
```
jwtty, line 1058: Can we also insert metrics test for reportDiff type?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -855,22 +1238,41 @@ func checkPlacementStatusMetric(registry *prometheus.Registry, crp *placementv1b
...
-		case "generation":
-			Expect(label.GetValue()).Should(Equal(strconv.FormatInt(crp.Generation, 10)))
-		case "name":
-			Expect(label.GetValue()).Should(Equal(crp.Name))
-		case "conditionType":
-			Expect(label.GetValue()).Should(Equal(condType))
-		case "status":
-			Expect(label.GetValue()).Should(Equal(status))
+	// Iterate over the labels and compare with expected values from statuses map
+	for _, emittedMetric := range placementStatusMetrics {
```
zhiying-lin, line 1243: Can you sort the placementStatusMetrics + labels here to make the result predictable? and then build the want + cmp.Diff to compare the results.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -855,22 +1238,41 @@ func checkPlacementStatusMetric(registry *prometheus.Registry, crp *placementv1b
...
+			case "generation":
+				Expect(label.GetValue()).Should(Equal(strconv.FormatInt(crp.Generation, 10)))
+			case "name":
+				Expect(label.GetValue()).Should(Equal(crp.Name))
+			case "conditionType":
+				condType = label.GetValue()
+				// Retrieve possible statuses from the map directly
+				expectedStatuses, statusExists = statuses[condType]
+				// Ensure that the conditionType exists in the map
+				Expect(statusExists).Should(BeTrue(), fmt.Sprintf("status for conditionType '%s' is missing in the conditionTypeStatuses map", condType))
```
zhiying-lin, line 1263: the current validation is not accurate.

We want the data to be exactly the one we want instead of using exists.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -440,8 +460,14 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
-			By("Ensure placement status metric was emitted with Scheduled False")
-			checkPlacementStatusMetric(customRegistry, crp, string(placementv1beta1.ClusterResourcePlacementScheduledConditionType), string(corev1.ConditionFalse))
+			By("Ensure placement status metric was emitted")
+			status := map[string][]string{
+				string(placementv1beta1.ClusterResourcePlacementScheduledConditionType): {
+					string(corev1.ConditionUnknown),
+					string(corev1.ConditionFalse),
+				},
+			}
+			checkPlacementStatusMetric(customRegistry, crp, status)
```
zhiying-lin, line 470: in this case, we want scheduled as false, right? why we allow the unknown?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -440,8 +564,22 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+			By("Ensure placement status metric was emitted")
+			status := map[string][]string{
+				string(placementv1beta1.ClusterResourcePlacementScheduledConditionType): {
+					string(corev1.ConditionUnknown),
+				},
+				string(placementv1beta1.ClusterResourcePlacementRolloutStartedConditionType): {
+					string(corev1.ConditionUnknown),
+				},
+			}
+			checkPlacementStatusMetric(customRegistry, crp, status)
```
zhiying-lin, line 579: in this case, we expect the metric only has scheduled unknown condition, right?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -855,22 +1238,41 @@ func checkPlacementStatusMetric(registry *prometheus.Registry, crp *placementv1b
...
+		// Initialize variables for each metric iteration
+		var condType string
+		var expectedStatuses []string
+		var statusExists bool
+
+		// Iterate over the labels and compare with expected values from conditionType
+		for _, label := range metricLabels {
+			switch label.GetName() {
+			case "generation":
+				Expect(label.GetValue()).Should(Equal(strconv.FormatInt(crp.Generation, 10)))
```
zhiying-lin, line 1255: the tests with different generation are missing?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: This is what I was saying earlier. If we do not delete it will show all the statuses it goes through so we could have multiple for the same type.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: Correct, I said I still needed to add that.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: In this case, since the CRP status has unknown and we check for the first !true, the first reconcile will see scheduled as unknown and then the second one will see rolloutStarted as unknown so we should have both.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: This just checking that the condition type exists in the map and there is a value attached to it. The next case for status checks to make sure the actual status is correct.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: If there is only one condition what timestamp am I comparing?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: I have updated the validation check

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -534,6 +662,596 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+			}
+			if label.GetName() == "name" {
+				Expect(label.GetValue()).Should(Equal(crpName))
+			}
+		}
+		Expect(emittedMetric.GetGauge().GetValue()).ShouldNot(Equal(float64(0)))
+	}
+}
+
+func checkPlacementStatusMetric(registry *prometheus.Registry, crp *placementv1beta1.ClusterResourcePlacement, statuses map[string]string) {
```
zhiying-lin, line 1187: please refer to my sample code, https://github.com/Azure/fleet/pull/1110/commits/7abce3f11d9c2376428bde9414cc5eed450da04a

pkg/controllers/clusterresourceplacement/controller_integration_test.go
zhiying-lin: The cmp package is comparing your want & got. you can use current time as the want, and allow some margins. but i'm ok if you only check value > 0 or not

pkg/controllers/clusterresourceplacement/controller_integration_test.go
zhiying-lin: Actually it would be very quite common, since crp can be triggered by resource without changing the crp generation. We should keep these data points

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: I have added two test cases

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +734,1117 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+								},
+							},
+						},
+					},
+				},
+			}
+			gotCRP = retrieveAndValidateClusterResourcePlacement(testCRPName, wantCRP)
+
+			By("Ensure placement complete metric was emitted with isCompleted True")
+			checkPlacementCompleteMetric(customRegistry, testCRPName, true, 2)
```
zhiying-lin, line 1317: we also need to validate the crp status metrics

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +734,1117 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+	Expect(err).Should(Succeed())
+	var placementStatusMetrics []*prometheusclientmodel.Metric
+	for _, mf := range metricFamilies {
+		if mf.GetName() == "fleet_workload_placement_status_last_timestamp_seconds" {
+			placementStatusMetrics = mf.GetMetric()
+		}
+	}
+
+	// Check the number of emitted metrics
+	Expect(len(placementStatusMetrics)).Should(Equal(len(wantMetrics)))
```
zhiying-lin, line 1846: this one is not needed, it should be covered by cmp.Diff

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -222,12 +226,13 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
...
 			r.Recorder.Event(crp, corev1.EventTypeNormal, "PlacementRolloutCompleted", "Placement has finished the rollout process and reached the desired status")
 		}
-		metrics.FleetPlacementStatus.WithLabelValues(crp.Name).Set(1)
+		metrics.FleetPlacementCompleteLastTimeStampSeconds.WithLabelValues(crp.Name, "true").SetToCurrentTime()
 		// We don't need to requeue any request now by watching the binding changes
 		return ctrl.Result{}, nil
 	}
 
-	metrics.FleetPlacementStatus.WithLabelValues(crp.Name).Set(0)
+	metrics.FleetPlacementCompleteLastTimeStampSeconds.WithLabelValues(crp.Name, "false").SetToCurrentTime()
```
zhiying-lin, line 234: this metrics looks redundant now, we can check the Status, if  Status is reporDiff == true || available == true, then it's completed.

We can merge these two metrics.

pkg/utils/test_util.go
```diff
@@ -32,6 +34,19 @@ var (
...
+
+	MetricsCmpOptions = []cmp.Option{
+		cmpopts.SortSlices(func(a, b *prometheusclientmodel.Metric) bool {
+			return a.GetGauge().GetValue() < b.GetGauge().GetValue() // sort by time
+		}),
+		cmpopts.SortSlices(func(a, b *prometheusclientmodel.LabelPair) bool {
+			return *a.Name < *b.Name // Sort by label name
+		}),
+		cmp.Comparer(func(a, b *prometheusclientmodel.Gauge) bool {
+			return (a.GetValue() > 0) == (b.GetValue() > 0)
```
zhiying-lin, line 46: we need to handle nil, otherwise it will panic

for example,
```
cmpopts.SortSlices(func(a, b *prometheusclientmodel.LabelPair) bool {
			if a.Name == nil || b.Name == nil {
                            return a.Name == nil
                        }
                        return *a.Name < *b.Name // Sort by label
		}),
```

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: For complete metrics, the test is very flaky depending how fast the conditions update since we don't have all the controllers working properly in the integration test. Originally we wanted the placementstatus to record the status for incomplete crp to see where they stopped.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -1035,3 +1040,29 @@ func isRolloutCompleted(crp *fleetv1beta1.ClusterResourcePlacement) bool {
...
+	for _, condType := range expectedCondTypes {
+		cond = crp.GetCondition(string(condType.ClusterResourcePlacementConditionType()))
+		if !condition.IsConditionStatusTrue(cond, crp.Generation) {
+			if cond != nil {
+				status = string(cond.Status)
+			}
+			metrics.FleetPlacementStatusLastTimeStampSeconds.WithLabelValues(crp.Name, strconv.FormatInt(crp.Generation, 10), string(condType.ClusterResourcePlacementConditionType()), status).SetToCurrentTime()
+			return
+		}
+	}
```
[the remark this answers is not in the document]
britaniar: I personally prefer option 1. The completed metrics makes it a lot easier to query whether the CRP is completed or not for dashboards. If in the future, not sure if we will, we add more final conditions (like Available and ReportDiff) then we will have to continue changing the alert to include these cases. The second metric was just to provide more information in the alert, while the first helps with dashboard and easier to filter out through the logs.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -344,6 +401,138 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
 				},
 			}
 			retrieveAndValidateClusterResourcePlacement(testCRPName, wantCRP)
+
+			By("Ensure placement complete metric was emitted with isCompleted False")
+			wantCompleteMetrics := []*prometheusclientmodel.Metric{
```
jwtty, line 406: This incomplete metric can be refactored into a function. There's code duplication.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -344,6 +401,138 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+						Value: ptr.To(float64(time.Now().UnixNano()) / 1e9),
+					},
+				},
+			}
+			checkPlacementCompleteMetric(customRegistry, wantCompleteMetrics)
+
+			By("Ensure placement status metric was emitted")
+			// A metric for the different reconcilations for all the !true statuses as the CRP reconciles.
+			// In this case we have 2 metrics for 1 condition type as Scheduled goes from `Unknown` to `False`.
+			wantStatusMetrics := []*prometheusclientmodel.Metric{
```
jwtty, line 511: Looks like every CRP now has a scheduled unknown metric emitted. Will this happen in prod too?

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +765,1077 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+					Value: ptr.To(float64(time.Now().UnixNano()) / 1e9),
+				},
+			})
+			checkPlacementStatusMetric(customRegistry, wantStatusMetrics)
 		})
-	})
-})
+
+		It("Emit metrics when CRP spec updates with different generations", func() {
+			By("By updating clusterSchedulingPolicySnapshot status to schedule success")
```
jwtty, line 789:
```suggestion
			By("Update clusterSchedulingPolicySnapshot status to schedule success")
```

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +765,1077 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+				},
+				{
+					ClusterName: member2Name,
+					Selected:    true,
+					Reason:      "valid",
+				},
+			}
+			Expect(k8sClient.Status().Update(ctx, gotPolicySnapshot)).Should(Succeed(), "Failed to update the policy snapshot status")
+
+			By("By creating clusterResourceBinding on member-1")
```
jwtty, line 812:
```suggestion
			By("Create clusterResourceBinding on member-1")
```

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +765,1077 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+			member1Binding = createOverriddenClusterResourceBinding(member1Name, gotPolicySnapshot, gotResourceSnapshot)
+
+			By("By validating the CRP status")
```
jwtty, line 815:
```suggestion
			By("Validate the CRP status")
```

pkg/controllers/clusterresourceplacement/controller_integration_test.go
jwtty: I actually prefer using "Creating", but I see none of the other code use "ing".

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +765,1077 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+						{Name: ptr.To("status"), Value: ptr.To(string(corev1.ConditionUnknown))},
+					},
+					Gauge: &prometheusclientmodel.Gauge{
+						Value: ptr.To(float64(time.Now().UnixNano()) / 1e9),
+					},
+				},
+			}
+			checkPlacementStatusMetric(customRegistry, wantStatusMetrics)
+
+			By("By creating a synchronized clusterResourceBinding on member-2")
```
jwtty, line 923:
```suggestion
			By("Create a synchronized clusterResourceBinding on member-2")
```

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +765,1077 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+					Version: "v1",
+					Kind:    "Namespace",
+					LabelSelector: &metav1.LabelSelector{
+						MatchLabels: map[string]string{"region": "west"},
+					},
+				})
+			Expect(k8sClient.Update(ctx, crp)).Should(Succeed(), "Failed to update crp")
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: testCRPName}, crp)).Should(BeNil(), "Get() clusterResourcePlacement mismatch")
+
+			By("By validating the CRP status with new spec")
```
jwtty, line 1048:
```suggestion
			By("Validate the CRP status with new spec")
```

pkg/controllers/clusterresourceplacement/controller_integration_test.go
```diff
@@ -533,7 +765,1077 @@ var _ = Describe("Test ClusterResourcePlacement Controller", func() {
...
+
+			wantCRP = &placementv1beta1.ClusterResourcePlacement{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       testCRPName,
+					Finalizers: []string{placementv1beta1.ClusterResourcePlacementCleanupFinalizer},
+				},
+				Spec: crp.Spec,
+				Status: placementv1beta1.ClusterResourcePlacementStatus{
+					ObservedResourceIndex: "0",
+					Conditions: []metav1.Condition{
```
jwtty, line 1166: I think this piece of code can also be refactored.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
zhiying-lin: i see, could you please add a comment there to clarify why we don't validate the status metrics here.

pkg/controllers/clusterresourceplacement/controller.go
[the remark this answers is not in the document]
zhiying-lin: agree with Wantong, they seem coupled, not sure how to join them but with different timestamp.

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -94,14 +94,15 @@ func (r *Reconciler) handleDelete(ctx context.Context, crp *fleetv1beta1.Cluster
 		return ctrl.Result{}, err
 	}
 
+	metrics.FleetPlacementCompleteLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
+	metrics.FleetPlacementStatusLastTimeStampSeconds.DeletePartialMatch(prometheus.Labels{"name": crp.Name})
```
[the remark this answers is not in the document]
jwtty: I asked to move these earlier. If the pod stops when finalizer is removed but before metrics are removed, they would be left never deleted.

test/utils/metrics/metrics.go
```diff
@@ -0,0 +1,30 @@
...
+var (
+	MetricsCmpOptions = []cmp.Option{
+		cmpopts.SortSlices(func(a, b *prometheusclientmodel.Metric) bool {
+			return a.GetGauge().GetValue() < b.GetGauge().GetValue() // sort by time
+		}),
+		cmpopts.SortSlices(func(a, b *prometheusclientmodel.LabelPair) bool {
+			if a.Name == nil || b.Name == nil {
+				return a.Name == nil
+			}
+			return *a.Name < *b.Name // Sort by label
```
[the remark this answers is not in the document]
jwtty: Yea, I just don't know whether this PR or #1107 will merge first. The latter one can always reuse this util.

michaelawyu:
Hi Britania! I am closing this PR as part of the CNCF repo migration process; please consider moving (re-creating) this PR in the new repo once [the sync PR](https://github.com/kubefleet-dev/kubefleet/pull/10) is merged. If there's any question/concern, please let me know. Thanks 🙏

pkg/controllers/clusterresourceplacement/controller.go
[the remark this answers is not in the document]
britaniar: Talked to Ryan. We will couple them.

pkg/controllers/clusterresourceplacement/controller_integration_test.go
britaniar: I check for the scheduled condition as well.

PR #1079, opened 2025-03-13
fix: Skip redundant noop status updates in crp

### LINKED ISSUE #940: [BUG] hub-agent making noop /status updates on every reconciliation
### **Describe the bug**

I'm observing that every time hub-agent restarts, it makes a ton of updates to `clusterrresourceplacement/status` for no reason. Given the default worker count is configured to be `1`, this starves the new items on the work queue on large clusters with lots of work to do for 10-20 minutes.

Similarly the resync period on the hub-agent is configured to be 5 minutes by default (way too aggressive IMO) and it exacerbates the frequency of the problem:

https://github.com/Azure/fleet/blob/6b81bdbbd9cc4187ba5c43cfbef0101fac22c655/cmd/hubagent/options/options.go#L66-L67

### **Environment**
Please provide the following: 
- Hub cluster details: hub-agent v0.8.5
- Member cluster details:  member-agent v0.8.5

### **To Reproduce**

- Restart controller (trigger a rolling update)
- Tail logs from the leader
- Observe it prints log statements to re-reconcile everything

      1030 17:55:03.104691       1 framework/framework.go:1310] "No change in scheduling decisions and condition, and the observed CRP generation remains the same" clusterSchedulingPolicySnapshot="lep-...
      1030 17:55:03.522789       1 framework/framework.go:1310] "No change in scheduling decisions and condition, and the observed CRP generation remains the same" clusterSchedulingPolicySnapshot="depo...
      1030 17:55:03.930300       1 framework/framework.go:1310] "No change in scheduling decisions and condition, and the observed CRP generation remains the same" clusterSchedulingPolicySnapshot="tran...
      1030 17:55:04.400769       1 framework/framework.go:1310] "No change in scheduling decisions and condition, and the observed CRP generation remains the same" clusterSchedulingPolicySnapshot="job-...
      1030 17:55:04.812419       1 framework/framework.go:1310] "No change in scheduling decisions and condition, and the observed CRP generation remains the same" clusterSchedulingPolicySnapshot="deci...
      1030 17:55:05.282259       1 framework/framework.go:1310] "No change in scheduling decisions and condition, and the observed CRP generation remains the same" clusterSchedulingPolicySnapshot="auto...
      1030 17:55:05.690881       1 framework/framework.go:1310] "No change in scheduling decisions and condition, and the observed CRP generation remains the same" clusterSchedulingPolicySnapshot="samp...
      1030 17:55:06.145613       1 framework/framework.go:1310] "No change in scheduling decisions and condition, and the observed CRP generation remains the same" clusterSchedulingPolicySnapshot="anti...

- Look at API Server audit logs, observe that it's making updates to `clusterrresourceplacement/status` for every `clusterSchedulingPolicySnapshot` object despite there should be no changes to the object (timestamps are matching):

    Image

- I'm not seeing any updated timestamps etc in `clusterresourceplacement/status` to warrant this `/status` update:

      status:
        conditions:
        - lastTransitionTime: "2024-09-19T00:06:37Z" # this is not today
          message: found all the clusters needed as specified by the scheduling policy
          observedGeneration: 1
          reason: SchedulingPolicyFulfilled
          status: "True"
          type: Scheduled
        observedCRPGeneration: 1
        targetClusters:
        - clusterName: redacted
          reason: picked by scheduling policy
          selected: true

In this part of the code we can clearly see the status update call is made unconditionally:
https://github.com/Azure/fleet/blob/cb9a7a0b305ae8a2875fa9e3ea9fb9f8d134b27b/pkg/controllers/clusterresourceplacement/controller.go#L193-L201

### **Expected behavior**

The controller should do `apiequality.Semantic.DeepEqual(old.status,new.status)`, and skip updating the status on the API when there's no reason to make this API call.

This would ensure the full resyncs and controller startup can happen very fast, and reduce the load on the API Server.

### **Screenshots**

Attached above

### **Additional context**

N/A

ahmetb:

Compare cached object to calculated status and if nothing has changed, skip the status update. When the cache catches up, we'll get another event in the workqueue anyway and we can catch the drift.

Otherwise this is causing a non-stop stream of crp/status updates when you have thousands of namespaces and each takes time to update status even though the update was actually noop, which causes workqueue buildup.

Fixes #940.

> linter seems to be broken at the go version 1.23.5?
>
>	```
>	ERRO Running error: 1 error occurred:
>	        * can't run linter goanalysis_metalinter: goanalysis_metalinter: buildir: package "slices" (isInitialPkg: false, needAnalyzeSource: true): Cannot range over: func(yield func(E) bool)
>	...
>	ERRO [linters_context/goanalysis] SA5012: panic during analysis: interface conversion: interface {} is nil, not *buildir.IR, goroutine 12724 [running]:
>	runtime/debug.Stack()
>	...
>	```

N/A –relying on existing tests.

Context at #940.

### DIFF
--- a/pkg/controllers/clusterresourceplacement/controller.go
+++ b/pkg/controllers/clusterresourceplacement/controller.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	corev1 "k8s.io/api/core/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -198,11 +199,13 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		return ctrl.Result{}, err
 	}
 
-	if err := r.Client.Status().Update(ctx, crp); err != nil {
-		klog.ErrorS(err, "Failed to update the status", "clusterResourcePlacement", crpKObj)
-		return ctrl.Result{}, err
+	if !apiequality.Semantic.DeepEqual(oldCRP.Status, crp.Status) {
+		if err := r.Client.Status().Update(ctx, crp); err != nil {
+			klog.ErrorS(err, "Failed to update the status", "clusterResourcePlacement", crpKObj)
+			return ctrl.Result{}, err
+		}
+		klog.V(2).InfoS("Updated the clusterResourcePlacement status", "clusterResourcePlacement", crpKObj)
 	}
-	klog.V(2).InfoS("Updated the clusterResourcePlacement status", "clusterResourcePlacement", crpKObj)
 
 	// We skip checking the last resource condition (available) because it will be covered by checking isRolloutCompleted func.
 	for i := condition.RolloutStartedCondition; i < condition.TotalCondition-1; i++ {


ArchanaAnand0212:
test failure seems not to be related to the change
there need to `err != nil ` [here](https://github.com/Azure/fleet/blob/9a376db6ee569a7f37b21a29b5ffeb6495c4fbed/test/e2e/actuals_test.go#L978)

```
[FAILED] Timed out after 10.001s.
  Failed to remove work resources from member cluster kind-cluster-1
  Expected success, but got an error:
      <*fmt.wrapError | 0xc000689020>: 
      work namespace application-3 still exists or an unexpected error occurred: %!w(<nil>)
      {
          msg: "work namespace application-3 still exists or an unexpected error occurred: %!w(<nil>)",
          err: nil,
      }
```
Other PR has similar failure.

ryanzhang-oss:
Thank you for the POC. Unfortunately, the CRP controller is not a controller-runtime controller (we need a customized one in order to accommodate our special case to watch every object in the cluster) so the reconcile loop won't be triggered by its status change. We don't want to reconcile based on status either.  With that side, we can brainstorm some ideas to avoid the update storm upon restart.

ryanzhang-oss:
Thank you for the POC. Unfortunately, the CRP controller is not a controller-runtime controller (we need a customized one in order to accommodate our special case to watch every object in the cluster) so the reconcile loop won't be triggered by its status change. We don't want to reconcile based on status either.  With that side, we can definitely brainstorm some ideas to avoid the update storm upon restart.

ryanzhang-oss:
> test failure seems not to be related to the change there need to `err != nil ` [here](https://github.com/Azure/fleet/blob/9a376db6ee569a7f37b21a29b5ffeb6495c4fbed/test/e2e/actuals_test.go#L978)
> 
> ```
> [FAILED] Timed out after 10.001s.
>   Failed to remove work resources from member cluster kind-cluster-1
>   Expected success, but got an error:
>       <*fmt.wrapError | 0xc000689020>: 
>       work namespace application-3 still exists or an unexpected error occurred: %!w(<nil>)
>       {
>           msg: "work namespace application-3 still exists or an unexpected error occurred: %!w(<nil>)",
>           err: nil,
>       }
> ```
> 
> Other PR has similar failure.

yeah, this is not related to this PR. We have a PR to fix the flaky e2e

pkg/controllers/clusterresourceplacement/controller.go
```diff
@@ -198,11 +199,13 @@ func (r *Reconciler) handleUpdate(ctx context.Context, crp *fleetv1beta1.Cluster
 		return ctrl.Result{}, err
 	}
 
-	if err := r.Client.Status().Update(ctx, crp); err != nil {
-		klog.ErrorS(err, "Failed to update the status", "clusterResourcePlacement", crpKObj)
-		return ctrl.Result{}, err
+	if !apiequality.Semantic.DeepEqual(oldCRP.Status, crp.Status) {
```
ryanzhang-oss, line 202: unfortuantely, this is not safe as we don't reconcile on status change

ahmetb: I still don't quite understand:

- who else do you expect will update the status of this resource and make it out of sync?
- why can't a periodic resync that you already have course-correct drifts eventually?

ryanzhang-oss: - Only the CRP controller changes its status but the reconcile is triggered by many other objects's status change. 
- if we rely on periodic resync to course-correct then it means the resync period needs to be reasonably short (say a min) which is something we agreed not a good practice.

We should still try to optimize the startup sequence though

michaelawyu:
Hi Ahmet (Long time no see ✨)! I am closing this PR as Fleet has been accepted as a CNCF sandbox project, and per our agreement with CNCF we will be moving to a CNCF hosted repo for future development; please consider moving (re-creating) this PR in [the new repo](https://github.com/kubefleet-dev/kubefleet) once [the sync PR](https://github.com/kubefleet-dev/kubefleet/pull/10) is merged. If there's any question/concern, please let me know. Thanks 🙏

PR #1083, opened 2025-03-16
test: add tests for eviction complete metric

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/pkg/controllers/clusterresourceplacementeviction/controller_test.go
+++ b/pkg/controllers/clusterresourceplacementeviction/controller_test.go
@@ -13,6 +13,8 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/prometheus/client_golang/prometheus"
+	prometheusclientmodel "github.com/prometheus/client_model/go"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -22,6 +24,7 @@ import (
 
 	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	"go.goms.io/fleet/pkg/utils/condition"
+	"go.goms.io/fleet/pkg/utils/controller/metrics"
 	"go.goms.io/fleet/pkg/utils/defaulter"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -365,7 +365,7 @@ func markEvictionNotExecuted(eviction *placementv1beta1.ClusterResourcePlacement
 }
 
 func emitEvictionCompleteMetric(eviction *placementv1beta1.ClusterResourcePlacementEviction) {
-	metrics.FleetEvictionStatus.Delete(prometheus.Labels{"name": eviction.Name, "isCompleted": "false"})
+	metrics.FleetEvictionStatus.Delete(prometheus.Labels{"name": eviction.Name, "isCompleted": "false", "isValid": "false"})
```
ryanzhang-oss, line 368: why only delete the false one?  Actually, why delete it anyway?

pkg/controllers/clusterresourceplacementeviction/controller_test.go
```diff
@@ -1469,6 +1476,195 @@ func TestIsEvictionAllowed(t *testing.T) {
...
+			name:       "failed reconcile - emit incomplete eviction metric",
+			request:    controllerruntime.Request{NamespacedName: types.NamespacedName{Name: "test-eviction"}},
+			isValid:    "false",
+			isComplete: "false",
+		},
+	}
+
+	for _, tc := range tests {
+		// Create a test registry
+		customRegistry := prometheus.NewRegistry()
```
ryanzhang-oss, line 1604: now you have a way to test he prometheus metics, why there is only one case?

pkg/controllers/clusterresourceplacementeviction/controller.go
Arvindthiru: I wanted to only emit a single line of completion status metric for each eviction. 

Sample workflow,
hitting metrics endpoint gives us:
fleet_workload_completion_status(name: test-eviction, isCompleted: false, isValid: false) - 1 (eviction is incomplete, controller ran into an internal error)

Then when it eventually completes I only wanted to see the metric with isCompleted set to true,
hitting metrics endpoint gives us:
fleet_workload_completion_status(name: test-eviction, isCompleted: true, isValid: false) - 4 (eviction is complete, eviction is invalid)

If we don't delete we would end up emitting two metrics for the same eviction,
hitting metrics endpoint gives us:
fleet_workload_completion_status(name: test-eviction, isCompleted: false, isValid: false) - 1
fleet_workload_completion_status(name: test-eviction, isCompleted: true, isValid: false) - 4

pkg/controllers/clusterresourceplacementeviction/controller_test.go
Arvindthiru: Added ITs for complete metric cases

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -365,7 +365,7 @@ func markEvictionNotExecuted(eviction *placementv1beta1.ClusterResourcePlacement
 }
 
 func emitEvictionCompleteMetric(eviction *placementv1beta1.ClusterResourcePlacementEviction) {
-	metrics.FleetEvictionStatus.Delete(prometheus.Labels{"name": eviction.Name, "isCompleted": "false"})
+	metrics.FleetEvictionStatus.Delete(prometheus.Labels{"name": eviction.Name, "isCompleted": "false", "isValid": "unknown"})
```
ryanzhang-oss, line 368: why not delete partial?

pkg/controllers/clusterresourceplacementeviction/controller_intergration_test.go
```diff
@@ -38,6 +41,99 @@ const (
...
+		// Reset metrics before each test
+		metrics.FleetEvictionStatus.Reset()
+		// emit incomplete eviction metric to simulate eviction failed once.
+		metrics.FleetEvictionStatus.WithLabelValues(evictionName, "false", "unknown").SetToCurrentTime()
+	})
+
+	AfterEach(func() {
+		ensureAllBindingsAreRemoved(crpName)
+		ensureEvictionRemoved(evictionName)
+		ensureCRPRemoved(crpName)
```
ryanzhang-oss, line 62: no need to remove registry or GC customRegistry?

pkg/controllers/clusterresourceplacementeviction/controller_intergration_test.go
```diff
@@ -536,3 +632,26 @@ func ensureAllBindingsAreRemoved(crpName string) {
 		ensureCRBRemoved(bindingList.Items[i].Name)
 	}
 }
+
+func checkEvictionCompleteMetric(registry *prometheus.Registry, isValid, isComplete string) {
```
ryanzhang-oss, line 636: why not just add this to the existing IT, no need to add more IT cases?

pkg/controllers/clusterresourceplacementeviction/controller_intergration_test.go
Arvindthiru: The registry is a new object in each BeforeEach and it's scope is confined to the Describe block, ungerstering eviction metric as part of cleanup

pkg/controllers/clusterresourceplacementeviction/controller.go
Arvindthiru: We exactly know, the lable combination we need to remove, in this case we need to be targeting this particular metric.

But when eviction is deleted, we don't know the combination of labels because it could be either a complete or incomplete eviction metric

ryanzhang-oss: DeletePartial makes the code more robust and fits the semantics of "delete the last inComplete run"

pkg/controllers/clusterresourceplacementeviction/controller.go
```diff
@@ -365,7 +365,7 @@ func markEvictionNotExecuted(eviction *placementv1beta1.ClusterResourcePlacement
 }
 
 func emitEvictionCompleteMetric(eviction *placementv1beta1.ClusterResourcePlacementEviction) {
-	metrics.FleetEvictionStatus.Delete(prometheus.Labels{"name": eviction.Name, "isCompleted": "false"})
+	metrics.FleetEvictionStatus.DeletePartialMatch(prometheus.Labels{"name": eviction.GetName()})
```
ryanzhang-oss, line 368: The semantics is "delete the last inComplete run"

ryanzhang-oss (approved):

PR #1085, opened 2025-03-18
test: log ns deletion timestamp in e2e to debug resource cleanup flakiness

jwtty:

Resource cleanup relies on garbage collection of child resources when their owner (AppliedWork) is deleted. The controllers do not wait for the clean up to fully complete to return. Our e2e test is flaky when checking the resources are completely deleted as current 10s wait time is not enough. Sample failures:

https://github.com/Azure/fleet/actions/runs/13911855578/job/38927583698
https://github.com/Azure/fleet/actions/runs/13840189622/job/38725462003?pr=1079

This PR logs the ns deletion timestamp to help debug.

Fixes #

I have:

### DIFF
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -743,7 +743,7 @@ func checkIfRemovedWorkResourcesFromMemberClusters(clusters []*framework.Cluster
 		memberCluster := clusters[idx]
 
 		workResourcesRemovedActual := workNamespaceRemovedFromClusterActual(memberCluster)
-		Eventually(workResourcesRemovedActual, 3*eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove work resources from member cluster %s", memberCluster.ClusterName)
+		Eventually(workResourcesRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove work resources from member cluster %s", memberCluster.ClusterName)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


test/e2e/utils_test.go
```diff
@@ -743,7 +743,7 @@ func checkIfRemovedWorkResourcesFromMemberClusters(clusters []*framework.Cluster
 		memberCluster := clusters[idx]
 
 		workResourcesRemovedActual := workNamespaceRemovedFromClusterActual(memberCluster)
-		Eventually(workResourcesRemovedActual, eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove work resources from member cluster %s", memberCluster.ClusterName)
+		Eventually(workResourcesRemovedActual, 3*eventuallyDuration, eventuallyInterval).Should(Succeed(), "Failed to remove work resources from member cluster %s", memberCluster.ClusterName)
```
ryanzhang-oss, line 746: this seems to be a regression on the clean up. We should not blindly relax the timeout. Instead, we should fix the root cause which is member agent GC slow

ryanzhang-oss:
another case 
https://github.com/Azure/fleet/actions/runs/13930432233/job/38996353144?pr=1086

probably easy to repro locally, need to see what actually happened on the GC

jwtty:
> another case https://github.com/Azure/fleet/actions/runs/13930432233/job/38996353144?pr=1086
> 
> probably easy to repro locally, need to see what actually happened on the GC

This one should be related to the bug fixed in #1096. The test didn't delete the namespace on the hub cluster at all. This timeout happens when we indeed try to delete the namespace but it times out waiting for GC.

michaelawyu:
Hi Wantong! I am closing this PR as part of the CNCF repo migration process; please consider moving (re-creating) this PR in the new repo once [the sync PR](https://github.com/kubefleet-dev/kubefleet/pull/10) is merged. If there's any question/concern, please let me know. Thanks 🙏

PR #1092, opened 2025-03-21
feat: deny the member leave when member still has serviceExport

zhiying-lin:
deny the member leave when member still has serviceExport in the webhook

Fixes #

I have:

### DIFF
--- a/pkg/webhook/membercluster/membercluster_validating_webhook.go
+++ b/pkg/webhook/membercluster/membercluster_validating_webhook.go
@@ -5,12 +5,15 @@ import (
 	"fmt"
 	"net/http"
 
+	admissionv1 "k8s.io/api/admission/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
+	fleetnetworkingv1alpha1 "go.goms.io/fleet-networking/api/v1alpha1"
 	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
 	"go.goms.io/fleet/pkg/utils"
 	"go.goms.io/fleet/pkg/utils/validator"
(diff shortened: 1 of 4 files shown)


test/e2e/fleet_guard_rail_test.go
```diff
@@ -841,37 +843,48 @@ var _ = Describe("fleet guard rail networking E2Es", Serial, Ordered, func() {
...
+			Expect(impersonateHubClient.Delete(ctx, &ise)).Should(Succeed(), "failed to delete Internal Service Export")
 		})
 
 		It("should allow CREATE operation on Internal service import resource in fleet-member namespace for user in member cluster identity", func() {
 			ise := internalServiceImport(isiName, imcNamespace)
 			By("expecting successful CREATE of Internal Service Import")
 			Expect(impersonateHubClient.Create(ctx, &ise)).Should(Succeed())
+
+			By("deleting Internal Service Import")
+			Expect(impersonateHubClient.Delete(ctx, &ise)).Should(Succeed(), "failed to delete Internal Service Import")
```
michaelawyu, line 857: A nit: this is no longer an ISE I think; the variable was named wrong.

test/e2e/fleet_guard_rail_test.go
```diff
@@ -841,37 +843,48 @@ var _ = Describe("fleet guard rail networking E2Es", Serial, Ordered, func() {
...
 		})
 
 		It("should allow CREATE operation on Endpoint slice export resource in fleet-member namespace for user in member cluster identity", func() {
 			ise := endpointSliceExport(epName, imcNamespace)
 			By("expecting successful CREATE of Endpoint slice export")
 			Expect(impersonateHubClient.Create(ctx, &ise)).Should(Succeed())
+
+			By("deleting Endpoint Slice Export")
+			Expect(impersonateHubClient.Delete(ctx, &ise)).Should(Succeed(), "failed to delete EndpointSlice Export")
```
michaelawyu, line 866: Same for this one.

test/e2e/fleet_guard_rail_test.go
```diff
@@ -841,37 +843,48 @@ var _ = Describe("fleet guard rail networking E2Es", Serial, Ordered, func() {
 			ise := internalServiceExport(iseName, imcNamespace)
 			By("expecting successful CREATE of Internal Service Export")
 			Expect(impersonateHubClient.Create(ctx, &ise)).Should(Succeed())
+
+			By("deleting Internal Service Export")
+			Expect(impersonateHubClient.Delete(ctx, &ise)).Should(Succeed(), "failed to delete Internal Service Export")
```
michaelawyu, line 848: Hi Zhiying! Just a nit: this probably needs a note saying that if it's not handled here the removal of the member cluster object would be blocked.

michaelawyu: The case is a bit different from the other two, where they can be deleted in cascade.

test/e2e/join_and_leave_test.go
```diff
@@ -95,9 +100,86 @@ var _ = Describe("Test member cluster join and leave flow", Ordered, Serial, fun
 			}
 		})
 
-		It("Should be able to unjoin a cluster with crp still running", func() {
-			By("remove all the clusters without deleting the CRP")
+		It("create a dummy internalServiceExport in the reserved member namespace", func() {
```
michaelawyu, line 103: Just a nit: it's my personal opinion, but it might be clearer if we refactor this into two separate test cases?

pkg/webhook/membercluster/membercluster_validating_webhook.go
```diff
@@ -22,26 +25,47 @@ var (
...
 func Add(mgr manager.Manager) error {
 	hookServer := mgr.GetWebhookServer()
-	hookServer.Register(ValidationPath, &webhook.Admission{Handler: &memberClusterValidator{admission.NewDecoder(mgr.GetScheme())}})
+	hookServer.Register(ValidationPath, &webhook.Admission{Handler: &memberClusterValidator{client: mgr.GetClient(), decoder: admission.NewDecoder(mgr.GetScheme())}})
 	return nil
 }
 
 // Handle memberClusterValidator checks to see if member cluster has valid fields.
-func (v *memberClusterValidator) Handle(_ context.Context, req admission.Request) admission.Response {
+func (v *memberClusterValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
```
michaelawyu, line 40: Another naming nit: the added logic might no longer belong in a validator? It's more of a guard 😄

pkg/webhook/membercluster/membercluster_validating_webhook.go
```diff
@@ -22,26 +25,47 @@ var (
...
+	mcObjectName := types.NamespacedName{Name: req.Name, Namespace: req.Namespace}
+	klog.V(2).InfoS("Validating webhook handling member cluster", "operation", req.Operation, "memberCluster", mcObjectName)
+
+	if req.Operation == admissionv1.Delete { // Will reject the requests whenever the serviceExport is not deleted
```
michaelawyu, line 44: I have some random thoughts, these are not blocker per se, just something that occurred to me when reading the code:

pkg/webhook/membercluster/membercluster_validating_webhook.go
michaelawyu: That is, the logic seems to be more aligned with those in `fleetresourcehandler` pkg?

pkg/webhook/membercluster/membercluster_validating_webhook.go
michaelawyu: We already have a webhook that guards against deletion attempts on Azure Fleet Manager from users (or any source other than RP, AKS support, etc.). So technically speaking the added logic is added to prevent RP (in most cases) from triggering the deletion. How would these components interact I guess? The concern is mostly that users won't see the error message that reminds them to delete service exports.

michaelawyu: For upstream they have full control so we generally assume that protection is not 100% a must.

michaelawyu (commented):
Added some comments, PTAL.

pkg/webhook/membercluster/membercluster_validating_webhook.go
zhiying-lin: fleetresourcehandler pkg is aim to validate the CR permissions so far. feel this package is better?

pkg/webhook/membercluster/membercluster_validating_webhook.go
zhiying-lin: In the RP side, we should surface this error to the user. Will test that.

test/e2e/join_and_leave_test.go
zhiying-lin: prefer to keep the current one to make the case complex to ensure it works as expected :)

michaelawyu (approved):
LGTM

PR #1096, opened 2025-03-25
test: fix e2e namespace already exist issue

jwtty:
Fix an e2e bug, the test didn't clean up the resources from the hub cluster.
Also fix a typo in the argocd demo doc just found :)

Fixes #

I have:

### DIFF
--- a/test/e2e/enveloped_object_placement_test.go
+++ b/test/e2e/enveloped_object_placement_test.go
@@ -342,7 +342,7 @@ var _ = Describe("Process objects with generate name", Ordered, func() {
 		// Create the namespace with both name and generate name set.
 		ns := appNamespace()
 		ns.GenerateName = nsGenerateName
-		Expect(hubClient.Create(ctx, &ns)).To(Succeed(), "Failed to create namespace %s", ns.Namespace)
+		Expect(hubClient.Create(ctx, &ns)).To(Succeed(), "Failed to create namespace %s", ns.Name)
 
 		// Create an envelope config map.
 		cm := &corev1.ConfigMap{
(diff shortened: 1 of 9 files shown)


test/e2e/enveloped_object_placement_test.go
```diff
@@ -481,11 +481,8 @@ var _ = Describe("Process objects with generate name", Ordered, func() {
 	})
 
 	AfterAll(func() {
-		// Remove the CRP.
-		cleanupCRP(crpName)
-
-		// Clean the placed resources.
-		cleanWorkResourcesOnCluster(allMemberClusters[0])
+		By(fmt.Sprintf("deleting placement %s and related resources", crpName))
+		ensureCRPAndRelatedResourcesDeleted(crpName, allMemberClusters)
```
zhiying-lin, line 485: why it's related to this AfterAll func?

This CRP only selects one cluster. What's the diff?

jwtty: The namespace and configmap are created on the hub cluster. The original cleanup only deletes the CRP and ensure member cluster is cleaned up, not the hub. That's why the next test scheduled on the same thread trying to create a namespace reporting namespace already exists error.

ryanzhang-oss: I get that the original test was not right but how did it pass if we increase the wait time as in your another PR?

ryanzhang-oss (approved):

ryanzhang-oss (approved):

test/e2e/enveloped_object_placement_test.go
jwtty: This test was introduced 5 days ago: https://github.com/Azure/fleet/pull/1086. My another PR was created and tested before that.

PR #1102, opened 2025-04-01
fix: use fleet-/kube- prefix checks in fleet guard rail

Arvindthiru:

Currently the fleet guard rail prevents users from creating namespaces with fleet/kube prefix instead of fleet-/kube- prefix.

This PR fixes that issue and I moved some code around to reuse code.

Eg: fleettest, kubetest namespaces will be prevented from being created which is not ideal

Fixes #

I have:

### DIFF
--- a/pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
+++ b/pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
-	"strings"
 
 	admissionv1 "k8s.io/api/admission/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -24,11 +23,11 @@ import (
 
 const (
 	// ValidationPath is the webhook service path which admission requests are routed to for validating custom resource definition resources.
-	ValidationPath             = "/validate-fleetresourcehandler"
-	groupMatch                 = `^[^.]*\.(.*)`
-	fleetMemberNamespacePrefix = "fleet-member"
-	fleetNamespacePrefix       = "fleet"
-	kubeNamespacePrefix        = "kube"
+	ValidationPath                                = "/validate-fleetresourcehandler"
+	groupMatch                                    = `^[^.]*\.(.*)`
+	allowedMessageMemberCluster                   = "upstream member cluster resource is allowed to be created/deleted by any user"
+	allowedMessageNonReservedNamespace            = "namespace name doesn't begin with fleet-/kube- prefix so we allow all operations on this namespace"
+	allowedMessageFleetReservedNamespacedResource = "namespace name of resource object doesn't begin with fleet-/kube- prefix so we allow all operations on request objects in these namespace"
 )
 
 // Add registers the webhook for K8s built-in object types.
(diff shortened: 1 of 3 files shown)


pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook.go
```diff
@@ -24,11 +23,11 @@ import (
...
 const (
 	// ValidationPath is the webhook service path which admission requests are routed to for validating custom resource definition resources.
-	ValidationPath             = "/validate-fleetresourcehandler"
-	groupMatch                 = `^[^.]*\.(.*)`
-	fleetMemberNamespacePrefix = "fleet-member"
-	fleetNamespacePrefix       = "fleet"
-	kubeNamespacePrefix        = "kube"
+	ValidationPath                                = "/validate-fleetresourcehandler"
+	groupMatch                                    = `^[^.]*\.(.*)`
+	allowedMessageMemberCluster                   = "upstream member cluster resource is allowed to be created/deleted by any user"
```
michaelawyu, line 28: Hi Arvind! Just a nit: shall we put the format/rule strings and the messages into separate blocks?

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook_test.go
```diff
@@ -518,6 +518,26 @@ func TestHandleMemberCluster(t *testing.T) {
 		resourceValidator fleetResourceValidator
 		wantResponse      admission.Response
 	}{
+		"allow create upstream fleet MC by any user": {
```
michaelawyu, line 521: Hi Arvind! We probably also need a UT for verifying the deletion behavior and another for verifying the `FleetAnnotationPrefix` presence branch.

michaelawyu: I understand that we might have already covered some of these in E2Es (upstream/RP)?

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook_test.go
```diff
@@ -953,14 +973,47 @@ func TestHandleNamespace(t *testing.T) {
...
+						Username: "testUser",
+						Groups:   []string{"testGroup"},
+					},
+					RequestKind: &utils.NamespaceMetaGVK,
+					Operation:   admissionv1.Create,
+				},
+			},
+			wantResponse: admission.Allowed(allowedMessageNonReservedNamespace),
+		},
+		"allow any user to modify non-reserved namespace with fleet prefix": {
```
michaelawyu, line 990: Hi Arvind! Just another nit: the name's a bit confusing to me, would something like `allow any user to modify non-reserved namespace` or `allow any user to modify namespace without fleet- prefix` be more clear?

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook_test.go
```diff
@@ -953,14 +973,47 @@ func TestHandleNamespace(t *testing.T) {
...
+						Username: "testUser",
+						Groups:   []string{"testGroup"},
+					},
+					RequestKind: &utils.NamespaceMetaGVK,
+					Operation:   admissionv1.Update,
+				},
+			},
+			wantResponse: admission.Allowed(allowedMessageNonReservedNamespace),
+		},
+		"allow any user to modify non-reserved namespace with kube prefix": {
```
michaelawyu, line 1004: Same for this one I guess...

michaelawyu (commented):
A few minor comments, PTAL

pkg/webhook/fleetresourcehandler/fleetresourcehandler_webhook_test.go
Arvindthiru: The cases are defined as a E2E https://github.com/Azure/fleet/blob/main/test/e2e/fleet_guard_rail_test.go#L49-L488

Added this case as I noticed we were not checking the allowed Message anywhere

michaelawyu (approved):
LGTM ;)

PR #1107, opened 2025-04-07
feat: add updaterun metrics and update Progressing condition

jwtty:
1. When updateRun is stuck waiting for a cluster update to finish or waiting for after-stage tasks to complete, we mark Progressing condition to false and add corresponding reacon.
2. Add updateRun status metrics based on updateRun condition. The metric contains 3 labels, name, generation, and status. Status can be one of "progressing", "stuck", "waiting", "completed", and "failed". The value is the timestamp when the status metric is emitted.

Generated metrics look like:
```
fleet_workload_update_run_status_last_timestamp_seconds{condition="Progressing",generation="1",name="example-run",reason="UpdateRunStarted",status="True"} 1.744335068438203e+09
fleet_workload_update_run_status_last_timestamp_seconds{condition="Progressing",generation="1",name="example-run",reason="UpdateRunWaiting",status="False"} 1.7443350222811694e+09
fleet_workload_update_run_status_last_timestamp_seconds{condition="Succeeded",generation="1",name="example-run",reason="UpdateRunSucceeded",status="True"} 1.7443350684599657e+09
```

Fixes #

I have:

### DIFF
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -327,7 +327,6 @@ func emitUpdateRunStatusMetric(updateRun *placementv1beta1.ClusterStagedUpdateRu
 		return
 	}
 
-	// We should not reach here, as reconcile should NOT return when the updateRun is still initializing or initialized but not progressing.
-	klog.V(2).ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("updateRun does not have valid conditions when emitting updateRun status metric")),
-		"Invalid updateRun status", "updateRun", klog.KObj(updateRun))
+	// We should rarely reach here, it can only happen when updating updateRun status fails.
+	klog.V(2).InfoS("There's no valid status condition on updateRun, status updating failed possibly", "updateRun", klog.KObj(updateRun))
 }
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -31,7 +32,24 @@ import (
 	"go.goms.io/fleet/pkg/utils"
 	"go.goms.io/fleet/pkg/utils/condition"
 	"go.goms.io/fleet/pkg/utils/controller"
+	"go.goms.io/fleet/pkg/utils/controller/metrics"
 	"go.goms.io/fleet/pkg/utils/informer"
+)
+
+// updateRunMetricsStatus is the status string for update run metrics.
+type updateRunMetricsStatus string
+
+const (
+	// updateRunMetricsStatusInitializing represents the update run is initializing in the metrics.
+	updateRunMetricsStatusInitializing updateRunMetricsStatus = "initializing"
+	// updateRunMetricsStatusProgressing represents the update run is in progress in the metrics.
+	updateRunMetricsStatusProgressing updateRunMetricsStatus = "progressing"
+	// updateRunMetricsStatusStuck represents the update run is stuck waiting for a cluster to be updated in the metrics.
+	updateRunMetricsStatusStuck updateRunMetricsStatus = "stuck"
+	// updateRunMetricsStatusCompleted represents the update run is completed in the metrics.
+	updateRunMetricsStatusCompleted updateRunMetricsStatus = "completed"
+	// updateRunMetricsStatusFailed represents the update run is failed in the metrics.
+	updateRunMetricsStatusFailed updateRunMetricsStatus = "failed"
 )
 
 var (
@@ -93,6 +111,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 			klog.V(2).InfoS("The clusterStagedUpdateRun has failed to initialize", "errorMsg", initCond.Message, "clusterStagedUpdateRun", runObjRef)
 			return runtime.Result{}, nil
 		}
+		emitUpdateRunStatusMetric(&updateRun, updateRunMetricsStatusInitializing)
 		if toBeUpdatedBindings, toBeDeletedBindings, err = r.initialize(ctx, &updateRun); err != nil {
 			klog.ErrorS(err, "Failed to initialize the clusterStagedUpdateRun", "clusterStagedUpdateRun", runObjRef)
 			// errInitializedFailed cannot be retried.
@@ -130,6 +149,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 	}
 
 	// Execute the updateRun.
+	emitUpdateRunStatusMetric(&updateRun, updateRunMetricsStatusProgressing)
 	klog.V(2).InfoS("Continue to execute the clusterStagedUpdateRun", "updatingStageIndex", updatingStageIndex, "clusterStagedUpdateRun", runObjRef)
 	finished, waitTime, execErr := r.execute(ctx, &updateRun, updatingStageIndex, toBeUpdatedBindings, toBeDeletedBindings)
 	if errors.Is(execErr, errStagedUpdatedAborted) {
@@ -158,13 +178,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req runtime.Request) (runtim
 // We delete all the dependent resources, including approvalRequest objects, of the clusterStagedUpdateRun object.
 func (r *Reconciler) handleDelete(ctx context.Context, updateRun *placementv1beta1.ClusterStagedUpdateRun) (bool, time.Duration, error) {
 	runObjRef := klog.KObj(updateRun)
-	// delete all the associated approvalRequests.
+	// Delete all the associated approvalRequests.
 	approvalRequest := &placementv1beta1.ClusterApprovalRequest{}
 	if err := r.Client.DeleteAllOf(ctx, approvalRequest, client.MatchingLabels{placementv1beta1.TargetUpdateRunLabel: updateRun.GetName()}); err != nil {
 		klog.ErrorS(err, "Failed to delete all associated approvalRequests", "clusterStagedUpdateRun", runObjRef)
 		return false, 0, controller.NewAPIServerError(false, err)
 	}
 	klog.V(2).InfoS("Deleted all approvalRequests associated with the clusterStagedUpdateRun", "clusterStagedUpdateRun", runObjRef)
+
+	// Delete the update run status metric.
+	metrics.FleetUpdateRunStatusLastTimestampSeconds.DeletePartialMatch(prometheus.Labels{"name": updateRun.GetName()})
+
 	controllerutil.RemoveFinalizer(updateRun, placementv1beta1.ClusterStagedUpdateRunFinalizer)
 	if err := r.Client.Update(ctx, updateRun); err != nil {
 		klog.ErrorS(err, "Failed to remove updateRun finalizer", "clusterStagedUpdateRun", runObjRef)
@@ -204,6 +228,7 @@ func (r *Reconciler) recordUpdateRunSucceeded(ctx context.Context, updateRun *pl
 		// updateErr can be retried.
 		return controller.NewUpdateIgnoreConflictError(updateErr)
 	}
+	emitUpdateRunStatusMetric(updateRun, updateRunMetricsStatusCompleted)
 	return nil
 }
 
@@ -228,6 +253,7 @@ func (r *Reconciler) recordUpdateRunFailed(ctx context.Context, updateRun *place
 		// updateErr can be retried.
 		return controller.NewUpdateIgnoreConflictError(updateErr)
 	}
+	emitUpdateRunStatusMetric(updateRun, updateRunMetricsStatusFailed)
 	return nil
 }
 
@@ -289,4 +315,9 @@ func handleClusterApprovalRequest(oldObj, newObj client.Object, q workqueue.Type
 	q.Add(reconcile.Request{
 		NamespacedName: types.NamespacedName{Name: updateRun},
 	})
+}
+
+// emitUpdateRunStatusMetric emits the update run status metric.
+func emitUpdateRunStatusMetric(updateRun *placementv1beta1.ClusterStagedUpdateRun, status updateRunMetricsStatus) {
+	metrics.FleetUpdateRunStatusLastTimestampSeconds.WithLabelValues(updateRun.Name, string(status)).SetToCurrentTime()
 }
--- a/pkg/controllers/updaterun/controller.go
+++ b/pkg/controllers/updaterun/controller.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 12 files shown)


pkg/controllers/updaterun/controller.go
```diff
@@ -31,9 +32,26 @@ import (
...
+// updateRunMetricsStatus is the status string for update run metrics.
+type updateRunMetricsStatus string
+
+const (
+	// updateRunMetricsStatusInitializing represents the update run is initializing in the metrics.
+	updateRunMetricsStatusInitializing updateRunMetricsStatus = "initializing"
+	// updateRunMetricsStatusProgressing represents the update run is in progress in the metrics.
+	updateRunMetricsStatusProgressing updateRunMetricsStatus = "progressing"
+	// updateRunMetricsStatusStuck represents the update run is stuck waiting for a cluster to be updated in the metrics.
+	updateRunMetricsStatusStuck updateRunMetricsStatus = "stuck"
```
zhiying-lin, line 48: what's the diff between progressing & stuck?

Is it possible to make use of the updateRun condition itself? when we introduce the new condition later, we don't need to update the metrics again.

pkg/controllers/updaterun/execution.go
```diff
@@ -161,10 +164,17 @@ func (r *Reconciler) executeUpdatingStage(
...
+
 		finished, updateErr := checkClusterUpdateResult(binding, clusterStatus, updatingStageStatus, updateRun)
 		if finished {
 			finishedClusterCount++
 			continue
+		} else {
+			// If cluster update has been running for more than 1 minute, mark the update run as stuck.
+			if time.Since(clusterStartedCond.LastTransitionTime.Time) > updateRunStuckThreshold {
+				emitUpdateRunStatusMetric(updateRun, updateRunMetricsStatusStuck)
+				fmt.Println("!!!threshold:", updateRunStuckThreshold)
```
zhiying-lin, line 176: typo?

Please add a log instead

pkg/utils/controller/metrics/metrics.go
```diff
@@ -61,6 +61,13 @@ var (
 		Name: "fleet_workload_eviction_complete",
 		Help: "Eviction complete status ",
 	}, []string{"name", "isCompleted", "isValid"})
+
+	// FleetUpdateRunStatusLastTimestampSeconds is a prometheus metric which holds the
+	// last update timestamp of update run status in seconds.
+	FleetUpdateRunStatusLastTimestampSeconds = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "fleet_workload_update_run_status_last_timestamp_seconds",
+		Help: "Last update timestamp of update run status in seconds",
+	}, []string{"name", "status"})
```
zhiying-lin, line 70: include the generation? we may support updating updateRun in the future.

pkg/controllers/updaterun/controller.go
```diff
@@ -31,9 +32,26 @@ import (
...
+	// updateRunMetricsStatusCompleted represents the update run is completed in the metrics.
+	updateRunMetricsStatusCompleted updateRunMetricsStatus = "completed"
+	// updateRunMetricsStatusFailed represents the update run is failed in the metrics.
+	updateRunMetricsStatusFailed updateRunMetricsStatus = "failed"
```
zhiying-lin, line 52: it means init failed or run failed? better to distinguish them

pkg/controllers/updaterun/controller.go
```diff
@@ -31,9 +32,26 @@ import (
...
 	"go.goms.io/fleet/pkg/utils/controller"
+	"go.goms.io/fleet/pkg/utils/controller/metrics"
 	"go.goms.io/fleet/pkg/utils/informer"
 )
 
+// updateRunMetricsStatus is the status string for update run metrics.
+type updateRunMetricsStatus string
+
+const (
+	// updateRunMetricsStatusInitializing represents the update run is initializing in the metrics.
```
zhiying-lin, line 43: are these status in this order?

initializing -> processing -> stuck/completed/failed

pkg/controllers/updaterun/controller.go
```diff
@@ -31,9 +32,26 @@ import (
...
+	updateRunMetricsStatusInitializing updateRunMetricsStatus = "initializing"
+	// updateRunMetricsStatusProgressing represents the update run is in progress in the metrics.
+	updateRunMetricsStatusProgressing updateRunMetricsStatus = "progressing"
```
Arvindthiru, line 46: As per offline discussion, can we either have progressing/stuck metric at any given point where we delete,
- progressing -> stuck, only emit stuck metric, delete progressing metric
- stuck -> progressing, only emit progressing, delete stuck metric

Instead of both metric lines at the same time, unless there is a need to having them both at the same time

pkg/controllers/updaterun/execution.go
jwtty: oh, I was doing some local experiment and mistakenly pushed it. Removed. Thanks!

pkg/controllers/updaterun/controller.go
jwtty: The failed status is used for both. I feel it not necessary to have two status as `initializing->failed` happens for init failed while `initializing->processing->failed` happens for run failed. I'm not sure if we need separate alerts for init failed and run failed.

pkg/controllers/updaterun/controller.go
jwtty: You are correct.

pkg/controllers/updaterun/controller.go
jwtty: `stuck` status means the updateRun has been waiting for a cluster update to finish for too long. I set a threshold of 1 minute and it's adjustable. We do not have a corresponding condition in the updateRun. All updateRuns that pass initialization phase have a `progressing` condition.

pkg/utils/controller/metrics/metrics.go
jwtty: I'm not sure if we really need the generation in metrics. If there's something wrong, we usually leverage logs to debug not the metrics. When setting up alerts, I cannot think of a scenario where I need to filter or do something specific for the generation. And generation itself adds more cardinality.

pkg/controllers/updaterun/controller.go
jwtty: initializing -> failed
initializing -> progressing -> completed/failed
initializing -> progressing -> stuck (per @Arvindthiru's comment below, I think I can delete the progressing metric when stuck is emiited)
initializing -> progressing -> stuck -> progressing -> completed/failed

pkg/controllers/updaterun/controller.go
jwtty: updated to only emit either one.

pkg/controllers/updaterun/controller.go
jwtty: I've updated the comments.

pkg/utils/controller/metrics/metrics.go
jwtty: add generation in metrics.

pkg/controllers/updaterun/controller.go
zhiying-lin: i feel it's better to improve our condition instead of reinventing the status. You still need to explain how the conditions will be changed (i mean the state machine, eg, initializing->processing->failed) to the cx. 

"stuck status means the updateRun has been waiting for a cluster update to finish for too long. " Should we also include in the CRP condition? change the reason?

pkg/controllers/updaterun/controller.go
zhiying-lin: if we emit the stuck & processing in different time, better to keep both instead of deleting. That's another data point.

pkg/controllers/updaterun/controller.go
jwtty: Used conditions instead.

pkg/controllers/updaterun/controller_integration_test.go
```diff
@@ -212,6 +226,127 @@ var _ = Describe("Test the clusterStagedUpdateRun controller", func() {
...
+		Gauge: &prometheusclientmodel.Gauge{
+			Value: ptr.To(float64(time.Now().UnixNano()) / 1e9),
+		},
+	}
+}
+
+func generateWaitingMetric(updateRun *placementv1beta1.ClusterStagedUpdateRun) *prometheusclientmodel.Metric {
+	return &prometheusclientmodel.Metric{
+		Label: generateMetricsLabels(updateRun, string(placementv1beta1.StagedUpdateRunConditionProgressing),
+			string(metav1.ConditionFalse), string(condition.UpdateRunWaitingReason)),
```
zhiying-lin, line 300:
```suggestion
			string(metav1.ConditionFalse), condition.UpdateRunWaitingReason),
```

pkg/utils/test_util.go
```diff
@@ -32,6 +34,19 @@ var (
 
 	// IgnoreConditionLTTAndMessageFields is a cmpopts.IgnoreFields that ignores the LastTransitionTime and Message fields
 	IgnoreConditionLTTAndMessageFields = cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime", "Message")
+
```
zhiying-lin, line 37: please refer to https://github.com/Azure/fleet/pull/1077 for this util

pkg/controllers/updaterun/controller_integration_test.go
```diff
@@ -212,6 +226,127 @@ var _ = Describe("Test the clusterStagedUpdateRun controller", func() {
...
+	return &prometheusclientmodel.Metric{
+		Label: generateMetricsLabels(updateRun, string(placementv1beta1.StagedUpdateRunConditionSucceeded),
+			string(metav1.ConditionTrue), string(condition.UpdateRunSucceededReason)),
+		Gauge: &prometheusclientmodel.Gauge{
+			Value: ptr.To(float64(time.Now().UnixNano()) / 1e9),
+		},
+	}
+}
+
+func validateUpdateRunMetricsRemoved(registry *prometheus.Registry) {
```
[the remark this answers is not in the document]
jwtty: Thanks, it's true. Removed.

pkg/controllers/updaterun/controller_integration_test.go
jwtty: bad AI :)

pkg/controllers/updaterun/execution.go
```diff
@@ -161,17 +164,24 @@ func (r *Reconciler) executeUpdatingStage(
...
 			return 0, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
 		}
+
 		finished, updateErr := checkClusterUpdateResult(binding, clusterStatus, updatingStageStatus, updateRun)
 		if finished {
 			finishedClusterCount++
 			continue
+		} else {
+			// If cluster update has been running for more than 1 minute, mark the update run as stuck.
+			if time.Since(clusterStartedCond.LastTransitionTime.Time) > updateRunStuckThreshold {
```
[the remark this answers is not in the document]
jwtty: Added.

pkg/controllers/updaterun/controller.go
```diff
@@ -290,3 +300,34 @@ func handleClusterApprovalRequest(oldObj, newObj client.Object, q workqueue.Type
...
+
+	initializedCond := meta.FindStatusCondition(updateRun.Status.Conditions, string(placementv1beta1.StagedUpdateRunConditionInitialized))
+	if condition.IsConditionStatusTrue(initializedCond, generation) || condition.IsConditionStatusFalse(initializedCond, generation) {
+		metrics.FleetUpdateRunStatusLastTimestampSeconds.WithLabelValues(updateRun.Name, genStr,
+			string(placementv1beta1.StagedUpdateRunConditionInitialized), string(initializedCond.Status), initializedCond.Reason).SetToCurrentTime()
+		return
+	}
+
+	// We should not reach here, as reconcile should NOT return when the updateRun is still initializing or initialized but not progressing.
+	klog.V(2).ErrorS(controller.NewUnexpectedBehaviorError(fmt.Errorf("updateRun does not have valid conditions when emitting updateRun status metric")),
```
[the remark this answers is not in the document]
jwtty: Yeah, this can happen when status updating fails. I do not want to emit an unknown metric as this can mess up with the actual unknown status on the updateRun. I want to log first as status updating failure should be retried.

pkg/utils/test_util.go
jwtty: I used prometheus built-in Getxxx() functions. Nil condition is handled natively.

pkg/utils/metrics/metrics.go
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package metrics
```
zhiying-lin, line 6: could you please move to the test/utils folder to split it from production code?

zhiying-lin, line 6: add a package comment?

jwtty: My bad. Done

jwtty: added, thanks!

pkg/controllers/updaterun/execution.go
```diff
@@ -161,17 +164,26 @@ func (r *Reconciler) executeUpdatingStage(
...
 			return 0, fmt.Errorf("%w: %s", errStagedUpdatedAborted, unexpectedErr.Error())
 		}
+
 		finished, updateErr := checkClusterUpdateResult(binding, clusterStatus, updatingStageStatus, updateRun)
 		if finished {
 			finishedClusterCount++
 			continue
+		} else {
+			// If cluster update has been running for more than 1 minute, mark the update run as stuck.
+			timeElapsed := time.Since(clusterStartedCond.LastTransitionTime.Time)
```
ryanzhang-oss, line 174: checkClusterUpdateResult wait until all the resources become available which depends on the "timeToReady" setting by the user. Therefore, 'updateRunStuckThreshold' has to be the greater of 60 seconds or that number

pkg/controllers/updaterun/execution.go
```diff
@@ -33,6 +33,9 @@ var (
 	// stageUpdatingWaitTime is the time to wait before rechecking the stage update status.
 	// Put it as a variable for convenient testing.
 	stageUpdatingWaitTime = 60 * time.Second
+
+	// updateRunStuckThreshold is the time to wait on a single cluster update before marking update run as stuck.
+	updateRunStuckThreshold = 60 * time.Second
```
ryanzhang-oss, line 38: this seems to be a bit too aggressive, a normal deployment with many replicas can take a few mins to get ready

pkg/controllers/updaterun/execution.go
jwtty: I realized one thing: "timeToReady" is a rolling update specific. It's under `RollingUpdateConfig` and cannot be set when rolling type is set to External. The wait is done in the rollingUpdate controller, not in the workApplier. For untrackable resources, workApplier simply sets it as Available with reason set to "Availability not trackable".

pkg/controllers/updaterun/execution.go
jwtty: Let me set it as 5 minutes for now. It will not cause updateRun to stop. Just mark it in the status so users can investigate CRP and see if there's issue. If issue is fixed or recovered automatically, updateRun can continue and the status is removed. We also add metrics to track so that we could have a better understanding and treak it.

pkg/controllers/updaterun/execution.go
ryanzhang-oss: good point, does it mean that we need to add this to the update run API too or create some wait between clusters when there are untrackable resources?

zhiying-lin: i notice this problem when reviewing the code. Should we keep the default value as 1 min to keep it consistent with the crp config? later, we can support the customized value?

jwtty: Yea, looks like we need to do similar thing in updateRun. That will be a new feature.

PR #1141, opened 2025-05-28
feat: add init container to install fleet CRDs

Arvindthiru:

Fixes #

I have:

### DIFF
--- a/charts/hub-agent/templates/deployment.yaml
+++ b/charts/hub-agent/templates/deployment.yaml
@@ -16,15 +16,13 @@ spec:
     spec:
       serviceAccountName: {{ include "hub-agent.fullname" . }}-sa
       initContainers:
-        {{- if .Values.crdInstaller.enabled }}
+        {{- if and .Values.crdInstaller.enabled .Values.enableV1Beta1APIs }}
         - name: crd-installer
           image: "{{ .Values.crdInstaller.image.repository }}:{{ .Values.crdInstaller.image.tag }}"
           imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
           args:
             - --mode={{ .Values.crdInstaller.mode }}
             - --v={{ .Values.crdInstaller.logVerbosity }}
-            - --enablev1alpha1API={{ .Values.enableV1Alpha1APIs }}
-            - --enablev1beta1API={{ .Values.enableV1Beta1APIs }}
         {{- end }}
       containers:
         - name: {{ .Chart.Name }}
--- a/charts/hub-agent/templates/deployment.yaml
+++ b/charts/hub-agent/templates/deployment.yaml
@@ -21,7 +21,7 @@ spec:
           image: "{{ .Values.crdInstaller.image.repository }}:{{ .Values.crdInstaller.image.tag }}"
           imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
           args:
-            - --mode={{ .Values.crdInstaller.mode }}
+            - --mode=hub
             - --v={{ .Values.crdInstaller.logVerbosity }}
         {{- end }}
       containers:
--- a/charts/hub-agent/templates/deployment.yaml
+++ b/charts/hub-agent/templates/deployment.yaml
@@ -15,6 +15,19 @@ spec:
         {{- include "hub-agent.selectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "hub-agent.fullname" . }}-sa
+      initContainers:
+        {{- if .Values.crdInstaller.enabled }}
+        - name: crd-installer
+          image: "{{ .Values.crdInstaller.image.repository }}:{{ .Values.crdInstaller.image.tag }}"
+          imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
+          args:
+            - --mode={{ .Values.crdInstaller.mode }}
+            - --wait={{ .Values.crdInstaller.waitForCRDs }}
+            - --timeout={{ .Values.crdInstaller.timeout }}
+            - --v={{ .Values.crdInstaller.logVerbosity }}
+            - --enablev1alpha1API={{ .Values.enableV1Alpha1APIs }}
+            - --enablev1beta1API={{ .Values.enableV1Beta1APIs }}
+        {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
--- a/charts/hub-agent/templates/deployment.yaml
+++ b/charts/hub-agent/templates/deployment.yaml
@@ -22,8 +22,6 @@ spec:
           imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
           args:
             - --mode={{ .Values.crdInstaller.mode }}
-            - --wait={{ .Values.crdInstaller.waitForCRDs }}
-            - --timeout={{ .Values.crdInstaller.timeout }}
             - --v={{ .Values.crdInstaller.logVerbosity }}
             - --enablev1alpha1API={{ .Values.enableV1Alpha1APIs }}
             - --enablev1beta1API={{ .Values.enableV1Beta1APIs }}
--- a/charts/member-agent/templates/deployment.yaml
+++ b/charts/member-agent/templates/deployment.yaml
@@ -16,6 +16,19 @@ spec:
     spec:
       restartPolicy: Always
       serviceAccountName: {{ include "member-agent.fullname" . }}-sa
+      initContainers:
+        {{- if .Values.crdInstaller.enabled }}
+        - name: crd-installer
+          image: "{{ .Values.crdInstaller.image.repository }}:{{ .Values.crdInstaller.image.tag }}"
+          imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
+          args:
+            - --mode={{ .Values.crdInstaller.mode }}
+            - --wait={{ .Values.crdInstaller.waitForCRDs }}
+            - --timeout={{ .Values.crdInstaller.timeout }}
+            - --v={{ .Values.crdInstaller.logVerbosity }}
+            - --enablev1alpha1API={{ .Values.enableV1Alpha1APIs }}
+            - --enablev1beta1API={{ .Values.enableV1Beta1APIs }}
+        {{- end }}
       containers:
         - name: {{ include "member-agent.fullname" . }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
--- a/charts/member-agent/templates/deployment.yaml
+++ b/charts/member-agent/templates/deployment.yaml
@@ -23,8 +23,6 @@ spec:
           imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
           args:
             - --mode={{ .Values.crdInstaller.mode }}
-            - --wait={{ .Values.crdInstaller.waitForCRDs }}
-            - --timeout={{ .Values.crdInstaller.timeout }}
             - --v={{ .Values.crdInstaller.logVerbosity }}
             - --enablev1alpha1API={{ .Values.enableV1Alpha1APIs }}
             - --enablev1beta1API={{ .Values.enableV1Beta1APIs }}
--- a/charts/member-agent/templates/deployment.yaml
+++ b/charts/member-agent/templates/deployment.yaml
@@ -17,15 +17,13 @@ spec:
       restartPolicy: Always
       serviceAccountName: {{ include "member-agent.fullname" . }}-sa
       initContainers:
-        {{- if .Values.crdInstaller.enabled }}
+        {{- if and .Values.crdInstaller.enabled .Values.enableV1Beta1APIs }}
         - name: crd-installer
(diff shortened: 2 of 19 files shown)


cmd/crdinstaller/main.go
```diff
@@ -0,0 +1,255 @@
...
+		crdName := crd.GetName()
+		klog.V(2).Infof("Creating CRD: %s", crdName)
+
+		// CRD GVR
+		crdGVR := schema.GroupVersionResource{
+			Group:    "apiextensions.k8s.io",
+			Version:  "v1",
+			Resource: "customresourcedefinitions",
+		}
+		_, err = client.Resource(crdGVR).Create(ctx, &crd, metav1.CreateOptions{})
```
ryanzhang-oss, line 197: use controllerruntime.CreateOrUpdate?

ryanzhang-oss: like pkg/controllers/clusterinventory/clusterprofile/controller.go:129

test/e2e/v1alpha1/webhook_test.go
```diff
@@ -558,6 +558,9 @@ var _ = Describe("Fleet's CRD Resource Handler webhook tests", func() {
 				g.Expect(HubCluster.KubeClient.Get(ctx, types.NamespacedName{Name: "memberclusters.fleet.azure.com"}, &crd)).Should(Succeed())
 				By("update labels in CRD")
 				labels := crd.GetLabels()
+				if labels == nil {
+					labels = make(map[string]string)
+				}
```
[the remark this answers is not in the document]
Arvindthiru: We run into panics https://github.com/Azure/fleet/actions/runs/15329080713/job/43153121211 I'm not sure why this PR triggers this though

cmd/crdinstaller/main.go
```diff
@@ -0,0 +1,108 @@
...
+	client, err := client.New(config, client.Options{
+		Scheme: scheme,
+	})
+
+	if err != nil {
+		klog.Fatalf("Failed to create Kubernetes client: %v", err)
+	}
+
+	// Install CRDs from the fixed location.
+	const crdPath = "/workspace/config/crd/bases"
```
ryanzhang-oss, line 67: I think this is set by the deployment yaml? This sidecar will break down if there is a mismatch, why not pass this in as a flag?

cmd/crdinstaller/main.go
```diff
@@ -0,0 +1,108 @@
...
+	if err := installCRDs(ctx, client, crdPath, *mode); err != nil {
+		klog.Fatalf("Failed to install CRDs: %v", err)
```
ryanzhang-oss, line 69: will this fail pod start?

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,201 @@
...
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	// CRDInstallerLabelKey is the label key used to indicate that a CRD is managed by the installer.
+	CRDInstallerLabelKey = "crd-installer.kubefleet.io/managed"
```
ryanzhang-oss: is this kubefleet or azurefleet?

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,201 @@
...
+	// AddonManagerLabelKey is the label key used to indicate that a CRD is managed by the addon manager.
+	AddonManagerLabelKey = "addonmanager.kubernetes.io/mode"
+)
+
+var (
+	hubCRDs = map[string]bool{
```
ryanzhang-oss: how can we keep this list in sync when new CRDs are introduced?

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,201 @@
...
+	decoder := codecFactory.UniversalDeserializer()
+
+	// Decode YAML into a structured CRD object.
+	obj, gvk, err := decoder.Decode(crdBytes, nil, nil)
+	if err != nil {
+		return fmt.Errorf("failed to decode CRD from %s: %w", path, err)
+	}
+
+	// Type assertion to make sure we have a CRD.
+	crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
```
ryanzhang-oss: is the GVK returned not enough?

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,201 @@
...
+
+		// Add an additional ownership label to indicate this CRD is managed by the installer.
+		if existingCRD.Labels == nil {
+			existingCRD.Labels = make(map[string]string)
+		}
+		// Ensure the label for management by the installer is set.
+		_, ok := existingCRD.Labels[CRDInstallerLabelKey]
+		if !ok {
+			existingCRD.Labels[CRDInstallerLabelKey] = "true"
+		}
```
ryanzhang-oss: why not set it anyway

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,201 @@
...
+		return fmt.Errorf("failed to decode CRD from %s: %w", path, err)
+	}
+
+	// Type assertion to make sure we have a CRD.
+	crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
+	if !ok {
+		return fmt.Errorf("unexpected type from %s, expected CustomResourceDefinition but got %s", path, gvk)
+	}
+
+	isManagedByAddonManager, err := isCRDManagedByAddonManager(ctx, client, crd.Name)
```
ryanzhang-oss: why get the CRD again in that function?

cmd/crdinstaller/main.go
ryanzhang-oss: oh, I take it back, it's from the docker build file. I guess we just have to keep this magic value.

cmd/crdinstaller/utils/util.go
Arvindthiru: We have options,

- Anyone adding new CRDs to the repo should update this map
- We can create another directory where we store CRDs installed by CRD installer and generate new CRDs in the new directory moving forward instead of a single directory containing all version of CRDs

cmd/crdinstaller/utils/util.go
Arvindthiru: We need make this assertion to access the labels to check if it's owned by addon manager, we can reuse this to achieve the same

charts/hub-agent/templates/deployment.yaml
```diff
@@ -15,6 +15,15 @@ spec:
         {{- include "hub-agent.selectorLabels" . | nindent 8 }}
     spec:
       serviceAccountName: {{ include "hub-agent.fullname" . }}-sa
+      initContainers:
+        {{- if and .Values.crdInstaller.enabled .Values.enableV1Beta1APIs }}
+        - name: crd-installer
+          image: "{{ .Values.crdInstaller.image.repository }}:{{ .Values.crdInstaller.image.tag }}"
+          imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
+          args:
+            - --mode={{ .Values.crdInstaller.mode }}
```
ryanzhang-oss: shouldn't this always be hub mode?

cmd/crdinstaller/main.go
Arvindthiru: We decide this location in the crd-installer docker file https://github.com/Azure/fleet/pull/1141/files#diff-7b7f3f96ad8a32f8674cdd6156bd2230711e15126785b656564766653ef4ef44R24. Since it's fixed I don't see why need a separate flag for it.

charts/member-agent/templates/deployment.yaml
```diff
@@ -16,6 +16,15 @@ spec:
     spec:
       restartPolicy: Always
       serviceAccountName: {{ include "member-agent.fullname" . }}-sa
+      initContainers:
+        {{- if and .Values.crdInstaller.enabled .Values.enableV1Beta1APIs }}
+        - name: crd-installer
+          image: "{{ .Values.crdInstaller.image.repository }}:{{ .Values.crdInstaller.image.tag }}"
+          imagePullPolicy: {{ .Values.crdInstaller.image.pullPolicy }}
+          args:
+            - --mode={{ .Values.crdInstaller.mode }}
```
ryanzhang-oss: has to be memeber

cmd/crdinstaller/main.go
```diff
@@ -0,0 +1,108 @@
+/*
+Copyright 2025 The KubeFleet Authors.
```
ryanzhang-oss: is this still kubefleet author?

cmd/crdinstaller/main.go
Arvindthiru: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#understanding-init-containers kubelet repeatedly restarts init container until it succeeds

If the init container exits with error, it should restart by default https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy. I didn't see a restart policy specified in our agent deployment chart in RP assuming it's Always as default value

test/e2e/v1alpha1/webhook_test.go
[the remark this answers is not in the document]
ryanzhang-oss: my bad, I got the map thing wrong.

cmd/crdinstaller/utils/util.go
Arvindthiru: The CRD we get is the CRD that exists in the cluster versus the one we read from config/crd/bases

cmd/crdinstaller/utils/util.go
Arvindthiru: We need to convert obj -> crd to access crd.Name to get object existing on the cluster. We can use check to achieve both (type assertion and get object name)

ryanzhang-oss: what's that GVK?  is that not always CustomResourceDefinition? or that's the real GVK of that CRD (i.e.  crp.kube-fleet.io)?

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Walk through the CRD directory and collect filenames.
+	if err := filepath.WalkDir(crdPath, func(path string, d fs.DirEntry, err error) error {
+		// Handle errors from WalkDir.
+		if err != nil {
+			return err
+		}
+
+		// Skip directories.
+		if d.IsDir() {
```
ryanzhang-oss, line 92: Are we supposed to have nested dirs? If yes, we can't skip. If no, what should we do?

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,161 @@
...
+			return nil
+		}
+
+		// Only process yaml files.
+		if filepath.Ext(path) != ".yaml" && filepath.Ext(path) != ".yml" {
```
ryanzhang-oss: what other files we can see in the dir?

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,161 @@
...
+			return nil
+		}
+
+		// Process based on mode.
+		filename := filepath.Base(path)
+		isHubCRD := hubCRDs[filename]
+		isMemberCRD := memberCRDs[filename]
+
+		switch mode {
```
ryanzhang-oss: so currently only "Hub & IsHub" or  "member & IsMember" works. What happens if we created a new CRD but didn't put that in the hubCRDs map? It will be missed.

cmd/crdinstaller/utils/util.go
ryanzhang-oss: we already have a directory containing all the CRDs. Currently, if we have an extra CRD not in any of those maps, it will be ignored. I wonder if there is a way to get rid of these two maps and only rely on the CRD itself? i.e has a label on the CRD to indicate if it's member or hub or some other way.

cmd/crdinstaller/utils/util.go
Arvindthiru: WalkDir starts from the root and includes root https://pkg.go.dev/path/filepath#WalkDir

cmd/crdinstaller/utils/util.go
Arvindthiru: image It's CustomResourceDefinition

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		crd, err := GetCRDFromPath(crdpath, scheme)
+		if err != nil {
+			return err
+		}
+
+		// For hub mode, only collect CRDs that are in fleet-placement, fleet-cluster categories.
+		if mode == "hub" {
+			if hubCRD[crdFileName] {
```
ryanzhang-oss: sig-multcluster crd

cmd/crdinstaller/utils/util.go
```diff
@@ -0,0 +1,163 @@
...
+				// special case for hub CRD that is not in any category.
+				crdsToInstall = append(crdsToInstall, *crd)
+				return nil
+			}
+			categories := crd.Spec.Names.Categories
```
ryanzhang-oss: what we actually want is CRDs in the kubenetes-fleet.io

PR #1162, opened 2025-07-29
fix: Update Github Action to publish arc member agent helm charts to MCR

MuhammadAliFleet:

This PR updates the Github Action to publish arc member agents helm charts to MCR.

Fixes #

I have:


ryanzhang-oss:
so this pulls images from both fleet and fleet-networking then why is the action here? Is there an ARC repo? Do we foresee more development needed purely for Arc?

ryanzhang-oss:
so this pulls images from both fleet and fleet-networking then why is the action here? Is there an ARC repo? Do we foresee more development needed purely for Arc?

Ealianis:
> so this pulls images from both fleet and fleet-networking then why is the action here? Is there an ARC repo? Do we foresee more development needed purely for Arc?

Took conversation into intenral Teams chat.

Makefile
```diff
@@ -336,6 +337,17 @@ docker-build-crd-installer: docker-buildx-builder
...
+.PHONY: helm-package-arc-member-agent
+helm-package-arc-member-agent:
+	helm package charts/member-agent-arc/ \
+		--version $(ARC_MEMBER_AGENT_IMAGE_VERSION) \
+		--set memberagent.tag=$(MEMBER_AGENT_IMAGE_VERSION) \
+		--set mcscontrollermanager.tag=$(MCS_CONTROLLER_IMAGE_VERSION) \
+		--set membernetcontrollermanager.tag=$(MEMBER_NET_CONTROLLER_IMAGE_VERSION) \
+		--set refreshtoken.tag=$(REFRESH_TOKEN_IMAGE_VERSION)
+
+	helm push member-agent-$(ARC_MEMBER_AGENT_IMAGE_VERSION).tgz oci://$(REGISTRY)/$(ARC_MEMBER_AGENT_IMAGE_NAME)
```
nibooy: Is `$REGISTRY` the `ARC_REGISTRY_REPO`? 

Is there a reason to add the additional subpath `$(ARC_MEMBER_AGENT_IMAGE_NAME)` to the MCR path we push to?
Then, when we issue Put ExtensionType Version, `FullPathToHelmChart` would look like the following:
MCR-registry/Microsoft.FleetMember/arc-member-cluster-agents-helm-chart/member-agent-0.0.0.tgz

nibooy: Resolved in an internal Teams chat.

yes, in the build-publish-mcr.yml, we set the env Variable REGISTRY to the   ARC_REGISTRY_REPO: public/microsoft.fleetmember

It is common practice to use subpaths.

.github/workflows/build-publish-mcr.yml
```diff
@@ -37,6 +43,11 @@ jobs:
           fi
           echo "release_tag=$RELEASE_TAG" >> $GITHUB_OUTPUT
 
+          # Fetch the latest fleet-networking version
```
zhiying-lin, line 46: Should we move it to the https://github.com/Azure/fleet-networking/blob/main/.github/workflows/build-publish-mcr.yml and handle it independently?

there is no order guaranteed when we cut the release.

there will be a case, we cut the fleet first and then fleet-networking. Then your acr does not have the latest fleet-networking tag.

xmuhammad-xali: The fleet member agent and networking agent both are needed to package and push to MCR.

We can move to fleet-networking but then we would then need to fetch the azure fleet version and would end up in the same case as you mentioned.

We will update the release process to update the order of cutting the release to 
1. azure/fleet-networking agent 
2. azure/fleet agent

I spoke to Britania, and it seems to be that DP team cuts the azure/fleet-networking first as it does not require backporting from the kubefleet and is easier compared to azure/fleet agent release.

jwtty: Why merging fleet-networking and fleet agent into one chart?

jwtty: Suppose someone updates fleet-networking repo, how do they know they need to update here too?

zhiying-lin: yeah, i have the same question as Wantong, why we need to couple fleet and fleet-networking into one chart? why not doing it independently?

xmuhammad-xali: **Why merging fleet-networking and fleet agent into one chart?**

We are working on enabling Arc connected clusters to join fleets as members. We install the agents on Arc clusters as an **Arc Extension**.  The Arc Extension is simply a packaged helm chart pushed into MCR. Since we are installing the networking and fleet agents onto the Arc Clusters, they need to be packaged as one for Arc Extension. Please see this [wiki](https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/personalplayground/812352/Fleet-Agent-Arc-Extension-Release-Process) for more info.

**Suppose someone updates fleet-networking repo, how do they know they need to update here too?**

We will be updating the Fleet Release Manual to ensure that we always run the Github Action to package all updated agents and push to MCR.

zhiying-lin: could you please add a comment in this workflow to provide such context? so that we don't lose the context in the future.

Please update the release manual to make sure that when people bump the agent versions for the aks cluster, use the same version combination as  the arc helm charts. Otherwise, what's installed in the normal aks cluster could be different from arc clusters.

xmuhammad-xali: In the Github Action, if no tag is provided for agent versions we fetch the latest tag and push that to MCR. We are doing the same for the ARC helm charts. We use the bumped/updated tag agent version for both fleet agent for the aks cluster and arc helm chart

.github/workflows/build-publish-mcr.yml
```diff
@@ -76,3 +87,13 @@ jobs:
         env:
           CRD_INSTALLER_IMAGE_VERSION: ${{ needs.prepare-variables.outputs.release_tag }}
           REGISTRY: ${{ secrets.AZURE_REGISTRY }}/${{ env.REGISTRY_REPO}}
+      - name: Build and publish ARC member agent helm chart
+        run: |
+          make helm-package-arc-member-agent
+        env:
+          ARC_MEMBER_AGENT_IMAGE_VERSION: ${{ needs.prepare-variables.outputs.fleet_networking_version }}
```
britaniar: Shouldn't this be it's own version and not use fleet-networking?

xmuhammad-xali: Yes the test PR had all the updated changes. Just ported the changes to this PR.

.github/workflows/build-publish-mcr.yml
zhiying-lin: to clarify that, i'm talking about [the toggle bump PR](https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/12955309?_a=files&path=/toggles/global/fleet-networking-config.yaml), who creates the bump PR needs to make sure the ARC is using the same versions (fleet & fleet-networking) as the one defined in the toggle to keep them consistent.

.github/workflows/build-publish-mcr.yml
```diff
@@ -76,3 +87,13 @@ jobs:
         env:
           CRD_INSTALLER_IMAGE_VERSION: ${{ needs.prepare-variables.outputs.release_tag }}
           REGISTRY: ${{ secrets.AZURE_REGISTRY }}/${{ env.REGISTRY_REPO}}
+      - name: Build and publish ARC member agent helm chart
```
britaniar: >We are working on enabling Arc connected clusters to join fleets as members. We install the agents on Arc clusters as an Arc Extension. The Arc Extension is simply a packaged helm chart pushed into MCR. Since we are installing the networking and fleet agents onto the Arc Clusters, they need to be packaged as one for Arc Extension. 

think including this in the file as a comment would be good. Gives more context

.github/workflows/build-publish-mcr.yml
```diff
@@ -37,6 +43,11 @@ jobs:
           fi
           echo "release_tag=$RELEASE_TAG" >> $GITHUB_OUTPUT
 
+          # Fetch the latest fleet-networking version
+          FLEET_NETWORKING_VERSION="${FLEET_NETWORKING_VERSION:-$(curl "https://api.github.com/repos/Azure/fleet-networking/tags" | jq -r '.[0].name')}"
```
britaniar, line 48: Please include a comment here that fleet-networking image will keep cut and pushed to MCR first before retrieving this.

britaniar (commented):
Overall LGTM. Just minor comments. I assume you have tested installing the helm chart you pushed and the current set up will work from start to finish.

Also small: PR Title check is failing please update to use one of [these](https://github.com/Azure/fleet/blob/7a763d6f9ad755000a6c484ba75a23663e813cd5/.github/pr-title-config.json#L7)

charts/member-agent-arc/values.yaml
```diff
@@ -0,0 +1,38 @@
...
+  tag: ""
+
+refreshtoken:
+  repository: mcr.microsoft.com/aks/fleet/refresh-token
+  tag: ""
+
+logVerbosity: 5
+
+namespace: 
+  fleet-system
```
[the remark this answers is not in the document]
britaniar: Ditto

.github/workflows/build-publish-mcr.yml
```diff
@@ -76,3 +87,13 @@ jobs:
         env:
           CRD_INSTALLER_IMAGE_VERSION: ${{ needs.prepare-variables.outputs.release_tag }}
           REGISTRY: ${{ secrets.AZURE_REGISTRY }}/${{ env.REGISTRY_REPO}}
+      - name: Build and publish ARC member agent helm chart
+        run: |
+          make helm-package-arc-member-agent
+        env:
+          ARC_MEMBER_AGENT_IMAGE_VERSION: ${{ inputs.arcExtensionVersion }}
```
nibooy: **nit:** Should we call it `ARC_MEMBER_AGENT_HELMCHART_VERSION` to be accurate?

Makefile
```diff
@@ -12,6 +12,7 @@ HUB_AGENT_IMAGE_NAME ?= hub-agent
 MEMBER_AGENT_IMAGE_NAME ?= member-agent
 REFRESH_TOKEN_IMAGE_NAME ?= refresh-token
 CRD_INSTALLER_IMAGE_NAME ?= crd-installer
+ARC_MEMBER_AGENT_IMAGE_NAME = arc-member-cluster-agents-helm-chart
```
nibooy: **nit:** `ARC_MEMBER_AGENT_HELMCHART_NAME`?

britaniar (approved):

PR #1189, opened 2025-09-09
fix: update indentation for custom cert

MuhammadAliFleet:

This PR fixes indentation in the arc agent deployment for custom cert.

Fixes #

I have:


nibooy (approved):

jim-minter:
Good find @MuhammadAliFleet !  Please can we have a unit test that uses go templates with a set of values that causes as many parts of the file as possible to be activated, then validates that the result is yaml parseable, so that we hopefully avoid a mistake like this in the future?

^ is a bit of work to do (suggest you use AI) and I'm OK with it being a follow-up as long as it happens soon.  Please create a workitem for that?

PR #1190, opened 2025-09-10
fix: add validation for CRP and PickFixed cluster names

nwnt:

As a result from the latest bug bash, we want to introduce validations for:
  - the CRP resource name, which currently only validates the length but not the characters in the name.
  - the cluster names for PickFixed policy.

I have:

Run unit tests and add new test cases for negative validation scenarios.

### DIFF
--- a/pkg/utils/validator/clusterresourceplacement.go
+++ b/pkg/utils/validator/clusterresourceplacement.go
@@ -194,9 +194,12 @@ func validatePolicyForPickFixedPlacementType(policy *placementv1beta1.PlacementP
 	}
 	uniqueClusterNames := make(map[string]bool)
 	for _, name := range policy.ClusterNames {
-		nameErr := validation.IsDNS1123Label(name)
+		nameErr := validation.IsDNS1123Subdomain(name)
 		if nameErr != nil {
 			allErr = append(allErr, fmt.Errorf("PickFixed cluster name %s is not a valid member name: %s", name, strings.Join(nameErr, "; ")))
+		}
+		if len(name) > validation.DNS1035LabelMaxLength {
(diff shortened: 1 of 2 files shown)


pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -101,8 +101,9 @@ func ValidateClusterResourcePlacementAlpha(clusterResourcePlacement *fleetv1alph
 func ValidateClusterResourcePlacement(clusterResourcePlacement *placementv1beta1.ClusterResourcePlacement) error {
 	allErr := make([]error, 0)
 
-	if len(clusterResourcePlacement.Name) > validation.DNS1035LabelMaxLength {
-		allErr = append(allErr, fmt.Errorf("the name field cannot have length exceeding %d", validation.DNS1035LabelMaxLength))
+	crpNameErr := validation.IsDNS1035Label(clusterResourcePlacement.Name)
```
ryanzhang-oss: this is not correct, we are not enforcing IsDNS1035Label for the CRP name. We are only enforcing the length of the name.

nwnt: removed this part.

pkg/utils/validator/clusterresourceplacement.go
```diff
@@ -194,6 +194,10 @@ func validatePolicyForPickFixedPlacementType(policy *placementv1beta1.PlacementP
 	}
 	uniqueClusterNames := make(map[string]bool)
 	for _, name := range policy.ClusterNames {
+		nameErr := validation.IsDNS1123Label(name)
```
ryanzhang-oss: actually it should be IsDNS1123Subdomain and length(name) < 64

nwnt: changed to what you suggested now.

PR #1195, opened 2025-09-16
feat: Add ARM64 image support

michaelawyu:

This PR updates the Makefile/Dockerfiles/GitHub Action pipeline setup to allow building ARM64 images.

I have:

N/A

### DIFF
--- a/docker/hub-agent.Dockerfile
+++ b/docker/hub-agent.Dockerfile
@@ -1,6 +1,9 @@
 # Build the hubagent binary
 FROM mcr.microsoft.com/oss/go/microsoft/golang:1.24.6 AS builder
 
+ARG GOOS=linux
+ARG GOARCH=amd64
+
 WORKDIR /workspace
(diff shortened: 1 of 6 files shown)


docker/hub-agent.Dockerfile
```diff
@@ -17,7 +20,8 @@ COPY pkg/ pkg/
 ARG TARGETARCH
```
britaniar: You removed this line in the other files but not this one. Is this expected?

michaelawyu: Ah, thanks! Let me drop this leftover line as well.

britaniar (dismissed):

michaelawyu:
Note: will add the step to create image manifest to bundle images from different archs under the same tag in a separate PR, after the pipeline has been verified.

(Kudos to @britaniar for raising the issue 🙏)

britaniar (approved):
Thanks for making the change.

PR #1202, opened 2025-09-19
fix: move managed resource VAP out of CRDinstaller

nwnt:

Move the managed resource VAP and its binding installation out of CRD installer.  They are now installed by the hub and member agents.  For hub, they are installed after the hub agent starts and never removed.  For member, they are installed during member join and removed during member leave.

Fixes # N/A

I have:

Unit tests and E2E tests.

### DIFF
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -196,11 +196,6 @@ func main() {
 		exitWithErrorFunc()
 	}
 
-	if err := managedresource.EnsureVAP(ctx, mgr.GetClient(), true); err != nil {
-		klog.Errorf("unable to create managed resource validating admission policy: %s", err)
-		exitWithErrorFunc()
-	}
-
 	// +kubebuilder:scaffold:builder
 
 	wg.Add(1)
@@ -215,6 +210,18 @@ func main() {
 
 		klog.InfoS("The controller manager has exited")
 	}()
+
+	// Wait for the cache to sync before creating managed resources
+	if !mgr.GetCache().WaitForCacheSync(ctx) {
+		klog.Error("failed to wait for cache sync")
+		exitWithErrorFunc()
+	}
+
+	if err := managedresource.EnsureVAP(ctx, mgr.GetClient(), true); err != nil {
+		klog.Errorf("unable to create managed resource validating admission policy: %s", err)
+		exitWithErrorFunc()
+	}
+	klog.Info("managed resource validating admission policy is successfully setup")
 
 	// Wait for the controller manager and the scheduler to exit.
 	wg.Wait()
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -196,8 +196,8 @@ func main() {
 		exitWithErrorFunc()
 	}
 
-	if managedresource.EnsureVAP(ctx, mgr.GetClient(), true); err != nil {
-		klog.Errorf("unable to create managedResource validating admission policy: %s", err)
+	if err := managedresource.EnsureVAP(ctx, mgr.GetClient(), true); err != nil {
+		klog.Errorf("unable to create managed resource validating admission policy: %s", err)
 		exitWithErrorFunc()
 	}
 
--- a/cmd/hubagent/main.go
+++ b/cmd/hubagent/main.go
@@ -52,6 +52,7 @@ import (
 	mcv1beta1 "go.goms.io/fleet/pkg/controllers/membercluster/v1beta1"
 	fleetmetrics "go.goms.io/fleet/pkg/metrics"
 	"go.goms.io/fleet/pkg/webhook"
+	"go.goms.io/fleet/pkg/webhook/managedresource"
 	// +kubebuilder:scaffold:imports
 )
 
@@ -192,6 +193,11 @@ func main() {
 	ctx := ctrl.SetupSignalHandler()
 	if err := workload.SetupControllers(ctx, &wg, mgr, config, opts); err != nil {
 		klog.ErrorS(err, "unable to set up ready check")
+		exitWithErrorFunc()
+	}
+
+	if managedresource.EnsureVAP(ctx, mgr.GetClient(), true); err != nil {
+		klog.Errorf("unable to create managedResource validating admission policy: %s", err)
 		exitWithErrorFunc()
 	}
 
--- a/pkg/webhook/managedresource/createordelete.go
+++ b/pkg/webhook/managedresource/createordelete.go
@@ -1,0 +1,46 @@
+/*
+Copyright (c) Microsoft Corporation.
+Licensed under the MIT license.
+*/
+
+package managedresource
+
+import (
+	"context"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func EnsureNoVAP(ctx context.Context, c client.Client, isHub bool) error {
+	objs := []client.Object{GetValidatingAdmissionPolicy(isHub), GetValidatingAdmissionPolicyBinding()}
+	for _, obj := range objs {
+		err := c.Delete(ctx, obj)
+		if err != nil && !apierrors.IsNotFound(err) {
+			if meta.IsNoMatchError(err) {
+				klog.Infof("ValidatingAdmissionPolicy is not supported in this cluster, skipping deletion")
+				return nil
+			}
+			return err
+		}
+	}
+	return nil
+}
+
+func EnsureVAP(ctx context.Context, c client.Client, isHub bool) error {
+	objs := []client.Object{GetValidatingAdmissionPolicy(isHub), GetValidatingAdmissionPolicyBinding()}
+	for _, obj := range objs {
+		opResult, err := controllerutil.CreateOrUpdate(ctx, c, obj, nil)
+		if err != nil {
+			if meta.IsNoMatchError(err) {
+				klog.Infof("ValidatingAdmissionPolicy is not supported in this cluster, skipping creation")
+			}
+			klog.Errorf("ValidatingAdmissionPolicy CreateOrUpdate (operation: %) failed: %s", opResult, err)
+			return err
+		}
+	}
+	return nil
+}
--- a/pkg/webhook/managedresource/createordelete.go
+++ b/pkg/webhook/managedresource/createordelete.go
@@ -33,12 +33,14 @@ func EnsureNoVAP(ctx context.Context, c client.Client, isHub bool) error {
 func EnsureVAP(ctx context.Context, c client.Client, isHub bool) error {
 	objs := []client.Object{GetValidatingAdmissionPolicy(isHub), GetValidatingAdmissionPolicyBinding()}
 	for _, obj := range objs {
-		opResult, err := controllerutil.CreateOrUpdate(ctx, c, obj, nil)
+		opResult, err := controllerutil.CreateOrUpdate(ctx, c, obj, func() error {
+			return nil
+		})
 		if err != nil {
 			if meta.IsNoMatchError(err) {
 				klog.Infof("ValidatingAdmissionPolicy is not supported in this cluster, skipping creation")
 			}
-			klog.Errorf("ValidatingAdmissionPolicy CreateOrUpdate (operation: %) failed: %s", opResult, err)
+			klog.Errorf("ValidatingAdmissionPolicy CreateOrUpdate (operation: %s) failed: %s", opResult, err)
 			return err
 		}
 	}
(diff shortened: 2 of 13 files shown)


cmd/hubagent/main.go
```diff
@@ -210,6 +211,18 @@ func main() {
 		klog.InfoS("The controller manager has exited")
 	}()
 
+	// Wait for the cache to sync before creating managed resources
+	if !mgr.GetCache().WaitForCacheSync(ctx) {
```
jim-minter, line 215: curious about why `WaitForCacheSync()` is needed -- what happens if it's omitted?

pkg/webhook/managedresource/createordelete.go
```diff
@@ -0,0 +1,48 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func EnsureNoVAP(ctx context.Context, c client.Client, isHub bool) error {
+	objs := []client.Object{GetValidatingAdmissionPolicy(isHub), GetValidatingAdmissionPolicyBinding()}
```
jim-minter: can `GetValidatingAdmissionPolicy()` and `GetValidatingAdmissionPolicyBinding()` be unexported at this point?

pkg/webhook/managedresource/createordelete.go
```diff
@@ -0,0 +1,48 @@
...
+
+func EnsureVAP(ctx context.Context, c client.Client, isHub bool) error {
+	objs := []client.Object{GetValidatingAdmissionPolicy(isHub), GetValidatingAdmissionPolicyBinding()}
+	for _, obj := range objs {
+		opResult, err := controllerutil.CreateOrUpdate(ctx, c, obj, func() error {
+			return nil
+		})
+		if err != nil {
+			if meta.IsNoMatchError(err) {
+				klog.Infof("ValidatingAdmissionPolicy is not supported in this cluster, skipping creation")
```
jim-minter: is it correct that a NoMatchError will cascade out of EnsureVAP?
(a) this is asymmetric with EnsureNoVAP(), which IMO is undesirable
(b) looking at the calling code, this behavior looks wrong?

I think the effect of this is that this PR would break agents on member clusters <1.30.0?

pkg/webhook/managedresource/createordelete.go
```diff
@@ -0,0 +1,48 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func EnsureVAP(ctx context.Context, c client.Client, isHub bool) error {
+	objs := []client.Object{GetValidatingAdmissionPolicy(isHub), GetValidatingAdmissionPolicyBinding()}
+	for _, obj := range objs {
+		opResult, err := controllerutil.CreateOrUpdate(ctx, c, obj, func() error {
```
jim-minter: I think that using `func() error { return nil }` is a sharp edge waiting for us that wasn't caught in a previous PR review.

I think that using this as a mutation function prevents the code from reconciling a pre-existing object to its desired state?  The effect of this is that we would fail to reconcile correctly if the end user somehow changed the VAP or Binding, or if later we changed its desired state here?

jim-minter: Can the following code be removed from `func install()` in a separate PR?

```go
	if mut == nil {
		mut = func() error { return nil }
	}
```

This code is avoiding an NPE at the cost of potentially promoting incorrect logic.

jim-minter: Please can you ensure there's unit testing that covers correctly reconciling an existing object?

pkg/webhook/managedresource/createordelete.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func EnsureNoVAP(ctx context.Context, c client.Client, isHub bool) error {
+	objs := []client.Object{GetValidatingAdmissionPolicy(isHub), GetValidatingAdmissionPolicyBinding()}
+	for _, obj := range objs {
+		err := c.Delete(ctx, obj)
+		if err != nil && !apierrors.IsNotFound(err) {
+			if meta.IsNoMatchError(err) {
+				klog.Infof("ValidatingAdmissionPolicy is not supported in this cluster, skipping deletion")
+				return nil
```
jim-minter: nit, this early return could confuse someone in the future if they added another object to `objs` and expected a deletion attempt always to be made on all the objects.  I'd consider refactoring like this:

```go
		switch {
		case err == nil, apierrors.IsNotFound(err):
			// continue
		case meta.IsNoMatchError(err):
			klog.Infof("object type %T is not supported in this cluster, continuing", obj)
			// continue
		default:
			return err
		}
```

jim-minter (dismissed):

cmd/hubagent/main.go
jim-minter: answer: it's needed for some reason internal to controller manager.

pkg/webhook/managedresource/createordelete.go
Arvindthiru: Can we not use client.Object and instead update the object's spec in the mutating function to what we expect ?

pkg/webhook/managedresource/validatingadmissionpolicy.go
```diff
@@ -14,63 +14,74 @@ const resourceName = "aks-fleet-managed-by-arm"
...
+	mutateValidatingAdmissionPolicy(vap, isHub)
+	return vap
+}
+
+func mutateValidatingAdmissionPolicy(vap *admv1.ValidatingAdmissionPolicy, isHub bool) {
+	vap.TypeMeta = metav1.TypeMeta{
+		APIVersion: "admissionregistration.k8s.io/v1",
+		Kind:       "ValidatingAdmissionPolicy",
+	}
+	vap.ObjectMeta = metav1.ObjectMeta{
```
jim-minter: You probably need to preserve the ResourceVersion

pkg/webhook/managedresource/validatingadmissionpolicy.go
```diff
@@ -14,63 +14,74 @@ const resourceName = "aks-fleet-managed-by-arm"
...
-func GetValidatingAdmissionPolicy(isHub bool) *admv1.ValidatingAdmissionPolicy {
+func getValidatingAdmissionPolicy(isHub bool) *admv1.ValidatingAdmissionPolicy {
 	vap := &admv1.ValidatingAdmissionPolicy{
-		TypeMeta: metav1.TypeMeta{
-			APIVersion: "admissionregistration.k8s.io/v1",
-			Kind:       "ValidatingAdmissionPolicy",
-		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name: resourceName,
 		},
```
jim-minter: duplicated in mutateValidatingAdmissionPolicy; can remove

pkg/webhook/managedresource/validatingadmissionpolicy.go
```diff
@@ -14,63 +14,74 @@ const resourceName = "aks-fleet-managed-by-arm"
...
+	}
+	mutateValidatingAdmissionPolicy(vap, isHub)
+	return vap
+}
+
+func mutateValidatingAdmissionPolicy(vap *admv1.ValidatingAdmissionPolicy, isHub bool) {
+	vap.TypeMeta = metav1.TypeMeta{
+		APIVersion: "admissionregistration.k8s.io/v1",
+		Kind:       "ValidatingAdmissionPolicy",
+	}
```
jim-minter: shouldn't be needed, but harmless

pkg/webhook/managedresource/validatingadmissionpolicy.go
nwnt: I removed the ones in their mutation instead.

pkg/webhook/managedresource/validatingadmissionpolicy.go
nwnt: Removed as well.

pkg/webhook/managedresource/validatingadmissionpolicy.go
```diff
@@ -14,63 +14,71 @@ const resourceName = "aks-fleet-managed-by-arm"
...
+	vap.ObjectMeta.Labels = map[string]string{
```
jim-minter: it probably makes sense to ensure the other fields in ObjectMeta though

target = ObjectMeta{ Labels: { "fleet.azure.com/managed-by": "arm" } }
target.ResourceVersion = actual.ResourceVersion
actual = target

pkg/webhook/managedresource/validatingadmissionpolicy.go
Arvindthiru: https://github.com/kubernetes-sigs/controller-runtime/blob/v0.22.1/pkg/controller/controllerutil/controllerutil.go#L320

Example scenaio of an update I can think of is member agent going down and coming back up,

If the user updated the object and when we rollback to what we expect the object to look like via an update we will only preserve the resourceVersion of the update the user made ?

@jim-minter is there another reason to preserve resource version I'm missing ?

jim-minter: The resource version comes from etcd, not anywhere else.  It increments every time the object is changed and is used for optimistic concurrency control.  It's necessary for it to be maintained through get->mutate->update otherwise the update call will be rejected.

Arvindthiru: Ah since we overwrite the enite object meta this is necessary here thanks

pkg/webhook/managedresource/createordelete.go
```diff
@@ -0,0 +1,85 @@
...
+		{
+			obj:    vapb,
+			mutate: mutateVAPB,
+		},
+	}
+
+	for _, objectMutator := range objsAndMutators {
+		opResult, err := controllerutil.CreateOrUpdate(ctx, c, objectMutator.obj, objectMutator.mutate)
+		switch {
+		case err == nil, apierrors.IsNotFound(err):
```
Arvindthiru: CreateOrUpdate will never return an IsNotFound error

nwnt: Removed that one.

pkg/webhook/managedresource/createordelete_test.go
```diff
@@ -0,0 +1,447 @@
...
+			// Verify initial state
+			if vap == nil {
+				t.Fatal("getVAPWithMutator() returned nil VAP")
+			}
+			if mutateFunc == nil {
+				t.Fatal("getVAPWithMutator() returned nil mutate function")
+			}
+
+			// Verify mutate function works
+			originalVAP := vap.DeepCopy()
```
Arvindthiru: nit: can we use gotVAP and wantVAP instead ?

pkg/webhook/managedresource/createordelete_test.go
```diff
@@ -0,0 +1,447 @@
...
+				err := fakeClient.Get(context.Background(), client.ObjectKeyFromObject(obj), obj)
+				if !apierrors.IsNotFound(err) {
+					t.Errorf("Expected object %s to be deleted, but Get() = %v", obj.GetName(), err)
+				}
+			}
+		})
+	}
+}
+
+func TestGetVAPWithMutator(t *testing.T) {
```
Arvindthiru, line 348: Do we need these tests since they are already covered by TestEnsureVAP ? same for TestGetVAPBindingWithMutator

test/e2e/managed_resource_vap_test.go
```diff
@@ -126,176 +131,248 @@ func expectDeniedByVAP(err error) {
...
-			if err != nil {
-				Expect(k8sErrors.IsNotFound(err)).To(BeTrue())
-			}
-			Expect(sysMastersClient.Create(ctx, managedNS)).To(Succeed())
-			var ns corev1.Namespace
-			err = sysMastersClient.Get(ctx, types.NamespacedName{Name: managedNS.Name}, &ns)
-			Expect(err).To(BeNil())
-			Expect(ns.Labels).To(HaveKeyWithValue(managedByLabel, managedByLabelValue))
-		})
+var _ = Describe("ValidatingAdmissionPolicy for Managed Resources", Label("managedresource"), Ordered, func() {
```
Arvindthiru: nit: remove new line

pkg/webhook/managedresource/createordelete_test.go
nwnt: I think these are clearer than the ones in TestEnsureVAP.  If we're to refactor it (which I'm doing soon enough), it's probably better to keep.

Arvindthiru (approved):

pkg/webhook/managedresource/createordelete.go
```diff
@@ -0,0 +1,86 @@
...
+			// continue
+		default:
+			klog.Errorf("Delete object type %T failed: %s", obj, err)
+			return err
+		}
+	}
+	return nil
+}
+
+func EnsureVAP(ctx context.Context, c client.Client, isHub bool) error {
```
ryanzhang-oss, line 37: I wonder where do we check the cluster version to not install VAP?

nwnt: Clusters with no VAP will result in `meta.IsNoMatchError(err)` returning true.  Clusters with VAP turned on, either it's v1.30+ or versions before v1.30 but have VAP enabled will not result in the same error.

PR #1215, opened 2025-10-28
fix: allow-list acsService username in validating admission webhook

nibooy:

fix: allow-list acsService username in validating admission webhook

Fixes #

I have:

### DIFF
--- a/pkg/webhook/managedresource/validatingadmissionpolicy.go
+++ b/pkg/webhook/managedresource/validatingadmissionpolicy.go
@@ -57,6 +57,7 @@ func mutateValidatingAdmissionPolicy(vap *admv1.ValidatingAdmissionPolicy) {
 				(
 					(
 						request.userInfo.username == "aksService" ||
+						request.userInfo.username == "acsService" ||
 						request.userInfo.username == "fleet-member-agent-sa"
 					)
 				    &&


pkg/webhook/managedresource/validatingadmissionpolicy.go
```diff
@@ -57,6 +57,7 @@ func mutateValidatingAdmissionPolicy(vap *admv1.ValidatingAdmissionPolicy) {
 				(
 					(
 						request.userInfo.username == "aksService" ||
+						request.userInfo.username == "acsService" ||
```
Arvindthiru, line 60: Please add a test covering this change ?

nibooy: The existing test coverage already covers the username "aksService", which is effectively the same configuration except for the username.
https://github.com/nwnt/fleet/blob/a828d882da318359263ce19538fbb609a9f836ca/test/e2e/framework/cluster.go#L109

If you think I should still add, I will.

ryanzhang-oss: I think adding a test in upstream is more like test the implementation than behavior as we won't know what's the actual username to white-list until this code is run in the real envioronment

PR #1210, opened 2025-10-16
feat: setup property checker for capacity

britaniar:
### Description of your changes

- Create a new struct called property checker with capacity related client and functions to verify cluster is valid based on property requirement.

- [x] Run `make reviewable` to ensure this PR is ready for review.

### Special notes for your reviewer

Another PR will be opened containing the deployment changes.
Another PR will be opened containing scheduler changes.

PR #1221, opened 2025-11-05
feat: update hub-agent to use custom cluster affinity plugin in scheduler when property checker is enabled

britaniar:

I have: 
- Hub agent reads in a Boolean to enable property checker functionalities and passes the compute service endpoint base url.
- Use newly added options to create scheduler with a custom cluster affinity plugin to use property checker.

- Tested in Standalone
Logs from hub agent used for testing:
```bash
I1106 19:15:29.423534       1 workload/setup.go:389] Setting up scheduler
I1106 19:15:29.423560       1 workload/setup.go:392] Azure property checker is enabled for cluster property validation
I1106 19:15:29.423626       1 workload/setup.go:408] Starting the scheduler
```

There are some copied files from [previous PR ](https://github.com/Azure/fleet/pull/1210) so that the setup could work.

### DIFF
--- a/cmd/hubagent/options/azureoptions.go
+++ b/cmd/hubagent/options/azureoptions.go
@@ -1,0 +1,31 @@
+//
+//Copyright (c) Microsoft Corporation.
+//Licensed under the MIT license.
+
+package options
+
+import (
+	"flag"
+)
+
+// AzurePropertyCheckerOptions holds the options for the Azure property checker.
+type AzurePropertyCheckerOptions struct {
+	// IsEnabled indicates whether the Azure property checker is enabled.
+	IsEnabled bool
+	// ComputeServiceAddressWithBasePath is the address of the Azure compute service with base path.
+	ComputeServiceAddressWithBasePath string
+}
+
+// NewAzurePropertyCheckerOptions creates a new AzurePropertyCheckerOptions with default values.
+func NewAzurePropertyCheckerOptions() AzurePropertyCheckerOptions {
+	return AzurePropertyCheckerOptions{
+		IsEnabled:                         false,
+		ComputeServiceAddressWithBasePath: "http://localhost:8421/compute",
+	}
+}
+
+// AddFlags adds flags for Azure Property Checker options to the given FlagSet.
+func (o AzurePropertyCheckerOptions) AddFlags(flags *flag.FlagSet) {
+	flags.BoolVar(&o.IsEnabled, "azure-property-checker-enabled", o.IsEnabled, "Enable Azure property checker for validating Azure-specific cluster properties.")
+	flags.StringVar(&o.ComputeServiceAddressWithBasePath, "azure-compute-server-address", o.ComputeServiceAddressWithBasePath, "The address of the Azure compute service with base path.")
+}
--- a/cmd/hubagent/options/options.go
+++ b/cmd/hubagent/options/options.go
@@ -112,6 +112,8 @@ type Options struct {
 	ResourceSnapshotCreationMinimumInterval time.Duration
 	// ResourceChangesCollectionDuration is the duration for collecting resource changes into one snapshot.
 	ResourceChangesCollectionDuration time.Duration
+	// AzurePropertyCheckerOpts contains options for Azure property checker
+	AzurePropertyCheckerOpts AzurePropertyCheckerOptions
 }
 
 // NewOptions builds an empty options.
@@ -134,6 +136,7 @@ func NewOptions() *Options {
 		PprofPort:                               6065,
 		ResourceSnapshotCreationMinimumInterval: 30 * time.Second,
 		ResourceChangesCollectionDuration:       15 * time.Second,
+		AzurePropertyCheckerOpts:                NewAzurePropertyCheckerOptions(),
 	}
 }
 
@@ -185,4 +188,5 @@ func (o *Options) AddFlags(flags *flag.FlagSet) {
 	flags.DurationVar(&o.ResourceChangesCollectionDuration, "resource-changes-collection-duration", 15*time.Second,
 		"The duration for collecting resource changes into one snapshot. The default is 15 seconds, which means that the controller will collect resource changes for 15 seconds before creating a resource snapshot.")
 	o.RateLimiterOpts.AddFlags(flags)
+	o.AzurePropertyCheckerOpts.AddFlags(flags)
 }
--- a/cmd/hubagent/workload/setup.go
+++ b/cmd/hubagent/workload/setup.go
@@ -33,6 +33,8 @@ import (
 	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
 	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
 	"go.goms.io/fleet/cmd/hubagent/options"
+	"go.goms.io/fleet/pkg/clients/azure/compute"
+	"go.goms.io/fleet/pkg/clients/httputil"
 	"go.goms.io/fleet/pkg/controllers/bindingwatcher"
 	"go.goms.io/fleet/pkg/controllers/clusterinventory/clusterprofile"
 	"go.goms.io/fleet/pkg/controllers/clusterresourceplacementeviction"
@@ -45,10 +47,12 @@ import (
 	"go.goms.io/fleet/pkg/controllers/schedulingpolicysnapshot"
 	"go.goms.io/fleet/pkg/controllers/updaterun"
 	"go.goms.io/fleet/pkg/controllers/workgenerator"
+	"go.goms.io/fleet/pkg/propertychecker/azure"
 	"go.goms.io/fleet/pkg/resourcewatcher"
 	"go.goms.io/fleet/pkg/scheduler"
 	"go.goms.io/fleet/pkg/scheduler/clustereligibilitychecker"
 	"go.goms.io/fleet/pkg/scheduler/framework"
+	"go.goms.io/fleet/pkg/scheduler/framework/plugins/clusteraffinity"
 	"go.goms.io/fleet/pkg/scheduler/profile"
 	"go.goms.io/fleet/pkg/scheduler/queue"
 	schedulerbindingwatcher "go.goms.io/fleet/pkg/scheduler/watchers/binding"
@@ -354,8 +358,22 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
 
 		// Set up the scheduler
 		klog.Info("Setting up scheduler")
-		defaultProfile := profile.NewDefaultProfile()
-		defaultFramework := framework.NewFramework(defaultProfile, mgr)
+		schedulerProfile := profile.NewDefaultProfile()
+		if opts.AzurePropertyCheckerOpts.IsEnabled {
+			klog.Info("Azure property checker is enabled for cluster property validation")
+			client, err := compute.NewAttributeBasedVMSizeRecommenderClient(opts.AzurePropertyCheckerOpts.ComputeServiceAddressWithBasePath, httputil.DefaultClientForAzure)
+			if err != nil {
+				klog.ErrorS(err, "Unable to create Azure vm size recommender client")
+				return err
+			}
+			klog.Info("Setting up cluster affinity plugin with Azure property checker")
+			clusterAffinityPlugin := clusteraffinity.New(clusteraffinity.WithPropertyChecker(azure.NewPropertyChecker(*client)))
+			profileOpts := profile.Options{
+				ClusterAffinityPlugin: &clusterAffinityPlugin,
+			}
+			schedulerProfile = profile.NewProfile(profileOpts)
+		}
+		defaultFramework := framework.NewFramework(schedulerProfile, mgr)
 		defaultSchedulingQueue := queue.NewSimplePlacementSchedulingQueue(
 			queue.WithName(schedulerQueueName),
 		)


pkg/propertychecker/azure/checker.go
```diff
@@ -0,0 +1,34 @@
+/* THIS FILE IS CREATED IN ANOTHER PROJECT AND COPIED HERE */
```
britaniar: As this states, this file is from my other PR and is just here so that the rest of the PR works.

cmd/hubagent/options/azureoptions.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"flag"
+)
+
+type AzurePropertyCheckerOptions struct {
+	// isEnabled indicates whether the Azure property checker is enabled.
+	isEnabled bool
+	// configFilePath is the path to the Azure property checker configuration file.
+	configFilePath string
```
zhiying-lin: you don't need to put the values into a config file, as we only have two right now.

input arguments should be fine.

cmd/hubagent/main.go
```diff
@@ -92,6 +92,9 @@ func main() {
 	opts := options.NewOptions()
 	opts.AddFlags(flag.CommandLine)
```
zhiying-lin, line 88: why not making the azure options as part of the type Options struct?

cmd/hubagent/options/azureoptions.go
britaniar: Oh was going off this suggestion https://github.com/Azure/fleet/pull/1210#discussion_r2480291616. Will change it to input arguments.

cmd/hubagent/workload/setup.go
```diff
@@ -384,6 +386,16 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
...
 		// Set up the scheduler
 		klog.Info("Setting up scheduler")
+		if opts.AzurePropertyCheckerOptions.IsEnabled() {
+			klog.Info("Azure property checker is enabled for cluster property validation")
+			_, err := compute.NewAttributeBasedVMSizeRecommenderClient(opts.AzurePropertyCheckerOptions.ComputeServiceAddressWithBasePath(), httputil.DefaultClientForAzure)
+			if err != nil {
+				klog.ErrorS(err, "unable to create azure vm size recommender client")
+				return err
+			}
+			// TODO: add back the property checker after changes have been merged
```
britaniar: Will update this in scheduler PR.

cmd/hubagent/options/azureoptions.go
zhiying-lin: oh, "config" in the original comment means putting "/compute" as part of an input argument, instead of hard coding in the code. :)

charts/hub-agent/templates/deployment.yaml
```diff
@@ -54,6 +54,10 @@ spec:
             - --cluster-unhealthy-threshold={{ .Values.clusterUnhealthyThreshold }}
             - --resource-snapshot-creation-minimum-interval={{ .Values.resourceSnapshotCreationMinimumInterval }}
             - --resource-changes-collection-duration={{ .Values.resourceChangesCollectionDuration }}
+            {{- if .Values.azurePropertyChecker.isEnabled }}
+            - --azure-property-checker-enabled={{ .Values.azurePropertyChecker.isEnabled }}
+            - --azure-property-checker-config-file={{ .Values.azurePropertyChecker.computeServiceAddressWithBasePath }}
+            {{- end }}
```
ryanzhang-oss:
```suggestion
            - --azure-property-checker-enabled={{ .Values.azurePropertyChecker.isEnabled }}
            {{- if .Values.azurePropertyChecker.isEnabled }}
            - --azure-property-checker-config-file={{ .Values.azurePropertyChecker.computeServiceAddressWithBasePath }}
            {{- end }}
```

cmd/hubagent/options/azureoptions.go
```diff
@@ -0,0 +1,40 @@
...
+
+package options
+
+import (
+	"flag"
+)
+
+type AzurePropertyCheckerOptions struct {
+	// isEnabled indicates whether the Azure property checker is enabled.
+	isEnabled bool
```
ryanzhang-oss: what does "azure property checker mean"?  We have an azure property provider already? Can this be false?

britaniar: This can be false. The property provider is the properties we are able to provide information about. While property checker is for just checking the properties.

charts/hub-agent/templates/deployment.yaml
```diff
@@ -53,6 +53,10 @@ spec:
             - --cluster-unhealthy-threshold={{ .Values.clusterUnhealthyThreshold }}
             - --resource-snapshot-creation-minimum-interval={{ .Values.resourceSnapshotCreationMinimumInterval }}
             - --resource-changes-collection-duration={{ .Values.resourceChangesCollectionDuration }}
+            - --azure-property-checker-enabled={{ .Values.azurePropertyChecker.isEnabled }}
+            {{- if .Values.azurePropertyChecker.isEnabled }}
+            - --azure-property-checker-config-file={{ .Values.azurePropertyChecker.computeServiceAddressWithBasePath }}
```
jwtty: Wrong argument name

cmd/hubagent/workload/setup.go
```diff
@@ -355,6 +359,21 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
...
+				klog.ErrorS(err, "unable to create azure vm size recommender client")
+				return err
+			}
+
+			klog.Info("Setting up cluster affinity plugin with Azure property checker")
+			clusterAffinityPlugin := clusteraffinity.New(clusteraffinity.WithPropertyChecker(azure.NewPropertyChecker(*client)))
+			profileOpts := profile.Options{
+				ClusterAffinityPlugin: &clusterAffinityPlugin,
+			}
+			defaultProfile = profile.NewProfile(profileOpts)
```
jwtty: rename it to "schedulerProfile"?

cmd/hubagent/workload/setup.go
```diff
@@ -354,8 +358,22 @@ func SetupControllers(ctx context.Context, wg *sync.WaitGroup, mgr ctrl.Manager,
...
 		// Set up the scheduler
 		klog.Info("Setting up scheduler")
-		defaultProfile := profile.NewDefaultProfile()
-		defaultFramework := framework.NewFramework(defaultProfile, mgr)
+		schedulerProfile := profile.NewDefaultProfile()
+		if opts.AzurePropertyCheckerOptions.IsEnabled() {
+			klog.Info("Azure property checker is enabled for cluster property validation")
+			client, err := compute.NewAttributeBasedVMSizeRecommenderClient(opts.AzurePropertyCheckerOptions.ComputeServiceAddressWithBasePath(), httputil.DefaultClientForAzure)
+			if err != nil {
+				klog.ErrorS(err, "unable to create azure vm size recommender client")
```
zhiying-lin:
```suggestion
				klog.ErrorS(err, "Unable to create Azure vm size recommender client")
```

cmd/hubagent/options/options.go
```diff
@@ -112,6 +112,8 @@ type Options struct {
 	ResourceSnapshotCreationMinimumInterval time.Duration
 	// ResourceChangesCollectionDuration is the duration for collecting resource changes into one snapshot.
 	ResourceChangesCollectionDuration time.Duration
+	// AzurePropertyCheckerOptions contains options for Azure property checker
+	AzurePropertyCheckerOptions *AzurePropertyCheckerOptions
```
zhiying-lin:
```suggestion
	AzurePropertyCheckerOpts AzurePropertyCheckerOptions
```
use structure instead of pointer, to keep it consistent with other options, eg, componentbaseconfig.LeaderElectionConfiguration & RateLimitOptions

cmd/hubagent/options/azureoptions.go
```diff
@@ -0,0 +1,41 @@
...
+package options
+
+import (
+	"flag"
+)
+
+// AzurePropertyCheckerOptions holds the options for the Azure property checker.
+type AzurePropertyCheckerOptions struct {
+	// isEnabled indicates whether the Azure property checker is enabled.
+	isEnabled bool
```
zhiying-lin:
```suggestion
	IsEnabled bool
```

cmd/hubagent/options/azureoptions.go
```diff
@@ -0,0 +1,41 @@
...
+	// computeServiceAddressWithBasePath is the address of the Azure compute service with base path.
+	computeServiceAddressWithBasePath string
```
zhiying-lin:
```suggestion
	ComputeServiceAddressWithBasePath string
```
usually the fields are public. You can refer to https://google.github.io/styleguide/go/best-practices#option-structure

cmd/hubagent/options/azureoptions.go
```diff
@@ -0,0 +1,41 @@
...
+
+// IsEnabled returns whether the Azure property checker is enabled.
+func (o *AzurePropertyCheckerOptions) IsEnabled() bool {
+	return o.isEnabled
+}
+
+// ComputeServiceAddressWithBasePath returns the address of the Azure compute service with base path.
+func (o *AzurePropertyCheckerOptions) ComputeServiceAddressWithBasePath() string {
+	return o.computeServiceAddressWithBasePath
+}
```
zhiying-lin:
```suggestion
```
these funcs are not needed any more once make them public

charts/hub-agent/templates/deployment.yaml
```diff
@@ -53,6 +53,10 @@ spec:
             - --cluster-unhealthy-threshold={{ .Values.clusterUnhealthyThreshold }}
             - --resource-snapshot-creation-minimum-interval={{ .Values.resourceSnapshotCreationMinimumInterval }}
             - --resource-changes-collection-duration={{ .Values.resourceChangesCollectionDuration }}
+            - --azure-property-checker-enabled={{ .Values.azurePropertyChecker.isEnabled }}
```
zhiying-lin: i'm thinking of not updating these charts at all to avoid the backport conflicts, as these default values of flags should be good. I don' think we will add e2e test in this repo.

charts/hub-agent/values.yaml
```diff
@@ -58,3 +58,7 @@ hubAPIBurst: 1000
 MaxConcurrentClusterPlacement: 100
 ConcurrentResourceChangeSyncs: 20
 MaxFleetSizeSupported: 100
+
+azurePropertyChecker:
```
zhiying-lin: you don't need these changes, either

cmd/hubagent/options/options.go
```diff
@@ -112,6 +112,8 @@ type Options struct {
 	ResourceSnapshotCreationMinimumInterval time.Duration
 	// ResourceChangesCollectionDuration is the duration for collecting resource changes into one snapshot.
 	ResourceChangesCollectionDuration time.Duration
+	// AzurePropertyCheckerOptions contains options for Azure property checker
+	AzurePropertyCheckerOptions AzurePropertyCheckerOptions
```
zhiying-lin:
```suggestion
	AzurePropertyCheckerOpts AzurePropertyCheckerOptions
```
to distinguish the field name and structure name,
similar to `RateLimiterOpts RateLimitOptions`

cmd/hubagent/options/azureoptions.go
```diff
@@ -0,0 +1,31 @@
...
+// NewAzurePropertyCheckerOptions creates a new AzurePropertyCheckerOptions with default values.
+func NewAzurePropertyCheckerOptions() AzurePropertyCheckerOptions {
+	return AzurePropertyCheckerOptions{
+		IsEnabled:                         false,
+		ComputeServiceAddressWithBasePath: "http://localhost:8421/compute",
+	}
+}
+
+// AddFlags adds flags for Azure Property Checker options to the given FlagSet.
+func (o AzurePropertyCheckerOptions) AddFlags(flags *flag.FlagSet) {
```
jwtty, line 28: I'm not sure if we want to make this a pointer. cc @zhiying-lin like RatelimiterOpts: https://github.com/kubefleet-dev/kubefleet/blob/main/cmd/hubagent/options/ratelimit.go#L43

cmd/hubagent/options/azureoptions.go
```diff
@@ -0,0 +1,31 @@
...
+// AzurePropertyCheckerOptions holds the options for the Azure property checker.
+type AzurePropertyCheckerOptions struct {
+	// IsEnabled indicates whether the Azure property checker is enabled.
+	IsEnabled bool
+	// ComputeServiceAddressWithBasePath is the address of the Azure compute service with base path.
+	ComputeServiceAddressWithBasePath string
+}
+
+// NewAzurePropertyCheckerOptions creates a new AzurePropertyCheckerOptions with default values.
+func NewAzurePropertyCheckerOptions() AzurePropertyCheckerOptions {
```
jwtty, line 20: ditto, do we want to return a pointer instead?

cmd/hubagent/options/azureoptions.go
zhiying-lin: my bad, we have to use the pointer in this case, otherwise, when we check `if opts.AzurePropertyCheckerOpts.IsEnabled {`, it's using original value, as the flag is using the copy of the option.

flags.BoolVar(&o.IsEnabled, "azure-property-checker-enabled", o.IsEnabled, "Enable Azure property checker for validating Azure-specific cluster properties.")

cmd/hubagent/options/azureoptions.go
zhiying-lin: i feel value is better here since the caller needs the value instead of the pointer.

cmd/hubagent/options/azureoptions.go
zhiying-lin: btw, here's the doc about receiver type, https://google.github.io/styleguide/go/decisions#receiver-type

PR #1326, opened 2026-06-23
fix: allow capacity http errors to requeue

britaniar:

When a querying the vm size recommender client, there may be times the client can run into transient client errors that will resolve on retry. At the moment the scheduler only runs once and sees all errors as unexpected. Now the possibility of transient http errors is possible with the addition of using azure property checker as we query an azure client. Therefore, these should not be considered unexpected, and the scheduler should be able to requeue on these errors.

Fixes #

I have: 
- Updated the scheduler to requeue on http errors. 
- Simplified http error logging to be one line for easier debugging.

Simplified logging:

```
## BEFORE
E0612 02:18:53.167602 1 compute/vmsizerecommenderclient.go:122] "Failed to generate VM size re...
request failed with status 503: POST http://127.0.0.1:5859/compute/subscriptions/a5191316-d253-43e3...

## AFTER
E0623 12:38:21.725116  158733 vmsizerecommenderclient.go:121] "Failed to generate VM size recommendations" err="request failed with status 503: POST http://127.0.0.1:35911/subscriptions/sub-123/providers/Microsoft.Compute/locations/eastus/vmSizeRecommendations/vmAttributeBased/generate" subscriptionID="sub-123" location="eastus" clientRequestID="c68fbda3-fc22-400f-abd0-e15b4f7af56f" latencyMs=2
```

With this change the scheduler would be pending until the errors return other errors/status codes that we don't deem transient while it retries.

```yaml
status:
  conditions:
  - lastTransitionTime: "2026-06-24T22:49:33Z"
    message: Scheduling has not completed
    observedGeneration: 4
    reason: SchedulePending
    status: Unknown
    type: ClusterResourcePlacementScheduled
```

**Open Question:** Do we want to keep the status as not complete/pending? 
- I think we should update the status for these specific scenarios (http errors from trying to query the capacity) so that the user has more of an idea of what is happening without having to sift through the logs.

### DIFF
--- a/pkg/scheduler/framework/framework.go
+++ b/pkg/scheduler/framework/framework.go
@@ -40,12 +40,12 @@ import (
 
 	clusterv1beta1 "go.goms.io/fleet/apis/cluster/v1beta1"
 	placementv1beta1 "go.goms.io/fleet/apis/placement/v1beta1"
-	"go.goms.io/fleet/pkg/clients/httputil"
 	"go.goms.io/fleet/pkg/scheduler/clustereligibilitychecker"
 	"go.goms.io/fleet/pkg/scheduler/queue"
 	"go.goms.io/fleet/pkg/utils/annotations"
 	"go.goms.io/fleet/pkg/utils/condition"
 	"go.goms.io/fleet/pkg/utils/controller"
+	fleetErrors "go.goms.io/fleet/pkg/utils/errors"
 	"go.goms.io/fleet/pkg/utils/parallelizer"
 )
 
@@ -540,9 +540,11 @@ func (f *framework) runAllPluginsForPickAllPlacementType(
 	passed, filtered, err := f.runFilterPlugins(ctx, state, policy, clusters)
 	if err != nil {
 		klog.ErrorS(err, "Failed to run filter plugins", "policySnapshot", policyRef)
-		// If the error is a transient HTTP error (e.g., 503 from external service),
-		// return it as-is so the scheduler can requeue. Otherwise, wrap it as unexpected behavior.
-		if httputil.IsTransientHTTPError(err) {
+		// Check if the error is retryable using structured error semantics.
+		// If the error (or any error in its chain) implements RetryableError and indicates
+		// it's retryable, return it as-is so the scheduler can requeue.
+		// Otherwise, wrap it as unexpected behavior.
+		if retryable, found := fleetErrors.IsRetryable(err); found && retryable {
 			return nil, nil, err
 		}
 		return nil, nil, controller.NewUnexpectedBehaviorError(err)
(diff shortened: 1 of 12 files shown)


logs.txt
michaelawyu, line 1: Hi Britania! Is this file needed?

pkg/scheduler/framework/framework.go
```diff
@@ -539,6 +540,11 @@ func (f *framework) runAllPluginsForPickAllPlacementType(
 	passed, filtered, err := f.runFilterPlugins(ctx, state, policy, clusters)
 	if err != nil {
 		klog.ErrorS(err, "Failed to run filter plugins", "policySnapshot", policyRef)
+		// If the error is a transient HTTP error (e.g., 503 from external service),
+		// return it as-is so the scheduler can requeue. Otherwise, wrap it as unexpected behavior.
+		if httputil.IsTransientHTTPError(err) {
```
michaelawyu: Hi Britania! I have a minor concern about part: I fear that there might be some logic leaking/coupling issue, where the scheduler is depending on a non-scheduler, function-specific implementation to determine if an error can be retried.

michaelawyu: Ideally we should use an error struct that has categories or an `isRetrieable` flag, which is wrapped all the way to the top, and let the control loop inspect and determine if the error can be retried, as opposed to unwrapping the errors to its original form of varied formats and relying the format-specific logic to check if the op can be retried.

But I understand that it requires a lot of change to achieve this at the moment given our existing architecture -> Britania, if OK, would you mind adding a comment here just to note that we are making an expedient fix?

michaelawyu:
For the open question, I did a brief check on the flow and it might require some changes to do so as the loop would end as it is right now given the error before it gets to update the (scheduling policy snapshot) status (i.e., if the plugins fail, the scheduling cycle ends right away with an error). We either have to update the status as part of the error handling logic, or refactor the flow so that status update would always happen despite errors at the end of each scheduling cycle. An alternative would be to emit an event, which is slightly easier to do.

I agree 100% that it is better UX to let user know about the errors of course -> Britania, if you don't mind me asking, which way do you think works best?

pkg/scheduler/framework/framework.go
britaniar: I updated to use be closer to what you suggested.

britaniar:
I think at the moment emitting an event is good. I will add that. If we decide to update the status, we can do that at a later time probably as we get more usage.

pkg/utils/errors/errors.go
```diff
@@ -54,7 +54,38 @@ const (
 	ErrCategoryUncategorized ErrCategory = "uncategorized"
 )
 
+// RetryableError is an interface that errors can implement to indicate whether the
+// operation that caused the error can be retried. This allows the control loop to
+// make retry decisions based on error semantics rather than inspecting error formats.
+type RetryableError interface {
```
michaelawyu: Thanks so much for this, Britania! Just a nit: do you think it might be better if we call the interface something like `ErrorWithRetryPolicy`?

michaelawyu: And the IsRetryable function can then return something like (isRetryable, hasRetryPolicyConfigured) I think

michaelawyu: Please feel free to tweak the names as you see fit; I am not very good at naming things 🙈

michaelawyu (dismissed):
Just a naming nit; otherwise LGTM

michaelawyu (approved):
Thanks, Britania 🙏

PR #1338, opened 2026-09-01
chore: backport kubefleet main (2026-09-01)

britaniar:
## Summary

- Backport `kubefleet-dev/kubefleet` main at `81194d7d` onto `Azure/fleet` main at `f21dbd0c`.
- Preserve Azure Fleet module imports, Microsoft Go builder consistency, existing CRD symlinks, and authenticated Slack link-check handling.

## Validation

- `GOTOOLCHAIN=go1.26.6 make reviewable` (WSL)

## Incoming commits

Initial complete output of `git log --oneline --decorate HEAD~1^2..HEAD^2`, captured with the initial merge commit checked out:

```text
48bde0d8 (cncf/main) chore: upgrade GOLANGCI Linter (v1.64.7 -> v2.13.2) (#862)
e5f16f29 chore: bump azure/setup-helm from 5.0.0 to 5.0.1 (#855)
10e88344 chore: bump distroless/base from `97b9d04` to `2d7d29b` in /docker (#852)
431ba73f interface: [FEP-0001] renamed Cluster Requests API (now Cluster Claims) (#803)
898e0bbb fix: skip access provider population in cluster profiles if no endpoint/CA data exists (#849)
9bcf5b23 fix: repin qemu-user-static to a content-addressable digest (#851)
caae3101 feat: build multi-arch (amd64/arm64) images and publish both tags in a single buildx (#745)
8a5e79d8 chore: bump github/codeql-action/autobuild from 4.37.6 to 4.37.7 (#842)
71fc1090 chore: bump step-security/harden-runner from 2.19.2 to 2.21.0 (#839)
```

Additional commit fetched and merged after the initial backport:

```text
81194d7d (cncf/main) fix: cross-compile images instead of emulating the builder stage (#858)
```

PR #745, opened 2024-04-01
feat: add resource override webhook configuration

britaniar:
### Description of your changes

Fixes #

I have:

- [x] Run `make reviewable` to ensure this PR is ready for review.

PR #1353, opened 2026-09-10
fix: restore FIPS-capable runtime base for agent images

ytimocin:

The hub-agent, member-agent and refresh-token binaries build with `CGO_ENABLED=1 GOEXPERIMENT=systemcrypto`. That backend `dlopen()`s libcrypto from the **runtime** image and, on a node with `/proc/sys/crypto/fips_enabled=1`, calls `SetFIPS(true)` — which needs a loadable FIPS provider in that image.

Azure Linux distroless ships one (`symcryptprovider.so` in `/usr/lib/ossl-modules`). Debian-based `gcr.io/distroless/base` ships `libcrypto.so.3` with no `ossl-modules` directory at all, so the call fails and the process panics before `main()`:

```
panic: opensslcrypto: FIPS mode requested (system FIPS mode) but not available: OpenSSL 3.5.6 7 Apr 2026
crypto/internal/backend.init.0()
	/usr/local/go/src/crypto/internal/backend/openssl_linux.go:39 +0x109
```

The `gcr.io` base arrived via the #1338 backport of upstream kubefleet#745 and shipped in v0.18.14, crash-looping (exit 2, `CrashLoopBackOff`) every hub cluster scheduled onto a FIPS-enabled node and taking `Scenario_Fleet_Hubful` with it. `crd-installer` was never switched and stayed healthy.

This restores `mcr.microsoft.com/azurelinux/distroless/base:3.0`. Only the final `FROM` changes — `:3.0` is a real multi-arch manifest list (linux/amd64 + linux/arm64), so the multi-platform buildx from #745 and the cross-compiling builder from #858 are unaffected.

I have:

- All three images build for `linux/amd64,linux/arm64` — amd64 through the cross toolchain, arm64 native, no QEMU.
- Built image's base layer digest matches the Azure Linux amd64 base exactly.
- The resulting `hubagent` starts normally on both architectures with FIPS forced on; the v0.18.14 image panics with exit 2 under the same conditions.

Reproducible in seconds, no FIPS hardware needed:

```bash
docker run --rm --platform linux/amd64 -e GOFIPS=1 mcr.microsoft.com/aks/fleet/hub-agent:v0.18.14  # panics, exit 2
docker run --rm --platform linux/amd64 -e GOFIPS=1 mcr.microsoft.com/aks/fleet/hub-agent:v0.18.13  # starts normally
```

Upstream kubefleet uses `gcr.io/distroless` here and always has, so this stays an Azure/fleet-only delta that future backports will try to overwrite — the comment on the `FROM` line records why. Upstream carries the same latent crash (systemcrypto on the providerless base since kubefleet#261); it goes unnoticed there only because FIPS is requested by the node, not the project.

### DIFF
--- a/docker/hub-agent.Dockerfile
+++ b/docker/hub-agent.Dockerfile
@@ -58,11 +58,11 @@ COPY pkg/ pkg/
 RUN echo "Building hubagent with GOOS=${TARGETOS} GOARCH=${TARGETARCH} CC=$(readlink -f /usr/local/bin/target-gcc)" && \
     CGO_ENABLED=1 CC=target-gcc GOOS=${TARGETOS} GOARCH=${TARGETARCH} GOEXPERIMENT=systemcrypto go build -o hubagent ./cmd/hubagent/
 
-# Use distroless as minimal base image to package the hubagent binary.
-# The pinned digest must reference a multi-arch image index so BuildKit can
-# resolve the matching base layer for each target architecture.
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/base:nonroot@sha256:2d7d29b504e7166f6d0c7655a18ebf5def5b37b029f8c4f8667e434ba774844f
+# GOEXPERIMENT=systemcrypto above dlopen()s libcrypto from THIS image and needs
+# a FIPS provider on FIPS-enabled nodes. Azure Linux has one; gcr.io/distroless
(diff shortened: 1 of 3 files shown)


nibooy (approved):

michaelawyu (approved):
LGTM ;) We probably need to do a renaming as Stephane suggested to guard these three files in rebases, but it does not have to be done in this PR.

